PR #60, opened 2016-11-18
Fix warnings and add additional node to jenkins

pfultz2:


pfultz2:
ok, so it looks like the hip part fails due to compiler bug. Do we want still run hip on jenkins?

kknox:
Lets talk about it face to face on monday

kknox:
@pfultz2 Can you document why you think it would be a good idea to call std::abort() in a library?  As a generic statement, I do not like it when a library aborts a process.

@guacamoleo I agree with all other changes.

pfultz2:
The program is not valid at that point. It would be better to throw exception, but you guys don't handle those in the library. The library should never reach these points, but if they do, then there is a serious problem that should be fixed.

kknox:
@pfultz2 
Can you squash these 13 commits to something around 5 commits that make the most sense?  I've also requested @guacamoleo to review.

pfultz2:
I don't know how to squash the commits like that.

kknox:
That's fine; would you mind if I squashed all the commits into 1 when I merge?  Github can do that automatically.

pfultz2:
Yea sure, go ahead.

PR #95, opened 2017-05-24
changing LocalWrite to GlobalRead for ease, de-linting from pyflakes

guacamoleo:


guacamoleo:
and enabling short vectors since the compiler has progressed

PR #103, opened 2017-08-01
Bumping the rocm version used for build

kknox:
ROCm v1.6 has been released, test & build on the latest
runtime as the kernels for all the build machines have
already been upgraded


kknox:
This commit fails unit testing, but it fixes build issues; acceptable to merge

PR #114, opened 2017-09-28
V_pk_fma

amcamd:
Changes to get hgemm to call 32 bit vectorized_packed_fma in place of 16 bit multiply and add. Note that this pull request has no changes for load and store commands. The plan is to address these in assembly code, not with half4 or half8 loads.

There is a change to initialize matrices with random values in [-max,max] rather than values in [0,max]. The reason for this change is to reduce the accumulator value in the reduction sums. This is achieved by having negative and positive values in the reduction sum. This is required because the 16 bit reduction sums for matrix size 4096 were giving overflow and underflow to inf or -inf.

### DIFF
--- a/Tensile/BenchmarkProblems.py
+++ b/Tensile/BenchmarkProblems.py
@@ -121,6 +121,7 @@ def benchmarkProblemType( problemTypeConfig, problemSizeGroupConfig, \
         "MathTemplates.cpp",
         "MathTemplates.h",
         "TensileTypes.h",
+        "KernelHeader.h",
         "ReferenceCPU.h",
         "SolutionHelper.cpp",
         "SolutionHelper.h",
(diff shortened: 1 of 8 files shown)


newling:
@amcamd @guacamoleo 
I've noticed that MIOpenGEMM kernels run slower when buffers are initialized with random values in [-R, R], as opposed to [0, R]. Maybe something to check out for in Tensile? I don't know what type of compression is being done in the background for this to happen, I'd be interested if you have any idea.

amcamd:
Addressed all comments with 4 commits:

1. Moved few lines of code from KernelWriter.py to KernelWriterSource.py
2. Renamed TYPE_MAC_HALF2 to TYPE_MAC, and conditionally defined TYPE_MAC
3. Rearrange loops to minimize code divergence
4. Move half declarations from TensileTypes.h to KernelHeader.h

amcamd:
Hi David,

I can confirm that tests passed for the following .yaml files

test_hgemm.yaml  :   hip, (OCL cannot be run because no half type in OCL)

test_sgemm.yaml  :  both hip and OCL

test_dgemm_defaults.yaml  :  both hip and OCL

Thanks,
Andrew

PR #116, opened 2017-10-02
fixed generic tensor contractions and added tests

guacamoleo:


guacamoleo:
only failures are ctu assembly instruction failures, which denote a regression in the assember, not in the kernels.

PR #121, opened 2017-10-10
better exposing memory combining for short-vectors

guacamoleo:


guacamoleo:
Passed all tests on rocm except for 1 solution (out of 64) for a 5D tensor contraction. That's at least as good as before.

PR #123, opened 2017-10-18
GlobalSplitU added

guacamoleo:
GlobalSplitU added
assembly divisions have better support for "Division by Invariant Integers using Multiplication"
DepthU can be any integer, no longer just power of 2's
Added tests for assembly GlobalSplitU


guacamoleo:
passed rocm_release

PR #147, opened 2018-02-11
Feature fix gcn9 asm

bensander:
Reviewer note: second commit changes many of the same lines, in a regular pattern.  
I used a macro approach to ease the transition between asm tool versions, we could even detect this automatically at some point.


kknox:
This looks good to me; the failures are no worse than before.  I'm going to squash the commit so that it looks like 1 commit in Tensile, but you will want to delete your feature-fix-gcn9-asm branch afterwards and start with a fresh branch

PR #150, opened 2018-02-12
add bytearray writer script

amcamd:
This script sill enable rebuilding from assembly in 1_BenchmarkProblems/Cijk_Ailk_Bjlk_HB_00/00_Final. It builds a bytearray from the .co file and uses it to build a new Kernels.cpp. The new Kernels.cpp can then be used to build a new client.


kknox:
@amcamd Does not want this merged yet

PR #151, opened 2018-02-13
Feature buffer load

bensander:


kknox:
Looks like this PR breaks double precision: 
```python
Tensile/KernelWriterAssembly.py", line 49, in toString
    instStr = "%s %s" % (name, (self.formatting % params) )
TypeError: not all arguments converted during string formatting
```

bensander:
What is best way to reproduce this error?  It looks like an 0x803 machine running DP.

I think BufferLoad should work for either arch but currently we only install for 0x900.  So I am thinking to just install the bufferload MemoryInstructions for both arch.

PR #153, opened 2018-02-15
multiple devices kernel selection logic

amcamd:
Select kernel based on device on which executable is running. 

There are library .yaml files for vega10, mi25, r9nano, and fallback. These contain tuning information for these devices. For vega10, mi25 and r9nano assembly code is used where possible. When Tensile.a is built (for standalone Tensile, or for inclusion in rocBLAS), it will contain kernels for these devices. The library .yaml file for the fallback device builds kernels from source code, and the fallback kernels are designed to run on any device. Source code is also used for hgemm on r9nano since gfx803 does not contain the instructions used in the hgemm assembly code.


amcamd:
replace with pull request #154

PR #159, opened 2018-02-20
Fix bug in BufferLoad mode related to writing too many registers

bensander:
In Buffer modes, only 32-bits is allocated to store the offset.
GLOBAL_OFFSET macros were writing 64 bit offsets.
Change so these only write 32-bits when the Buffer modes are enabled


kknox:
@amcamd Are you able to test with this PR against rocblas unit tests?  This passes all tensile unit tests, so I'm not opposed to checking it in if that makes it easier for you to run rocblas unit tests

PR #163, opened 2018-03-02
Determine if assembler require explicit co syntax

bensander:
Try to assemble the new syntax and check the error code

### DIFF
--- a/Tensile/KernelWriterAssembly.py
+++ b/Tensile/KernelWriterAssembly.py
@@ -20,10 +20,9 @@
 ################################################################################
 
 from SolutionStructs import DataType
-from Common import globalParameters, printExit, printWarning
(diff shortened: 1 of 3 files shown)


Tensile/KernelWriterAssembly.py
```diff
@@ -20,10 +20,9 @@
 ################################################################################
 
 from SolutionStructs import DataType
-from Common import globalParameters, printExit, printWarning
+from Common import globalParameters, printExit, printWarning, print1, print2
```
amcamd: Linter gives the following error:

M7IZMN2KPFQYNMDFED4FE2ZJHDBR3F7AMRA4TCOLVKDIIAHSVVMA/src/tensile/Tensile/KernelWriterAssembly.py:23: 'Common.print1' imported but unused

My understanding is that the print1 needs to be removed.

PR #179, opened 2018-03-14
Add serial-in-u DataInitTypeAB mode (5=serial-in-u)

bensander:
- Will initialize matrix so serial index is in the K dimension

- Use common initialization routine for A/B/C.
   - Now support NAN inits for A and B

- Refactor init logic so it resets the data values for each matrix, which
is a new requirement of this mode.
  - Separate copyData function to copy A/B
  - specializeAB to track if the matrix init is same for all sizes (old
    modes) or "specialized" for matrix dims (this new mode).


bensander:
Andrew, thought this would be good for you to review and merge before client refactor since it touches some of the same files.

PR #188, opened 2018-03-21
Add DirectToLDS option [Initial Checkin]

bensander:
If possible, returned data flows directly into LDS rather than
through VGPRs.  Save registers and latency.

- Determine if vector width and transpose op support DirectToLDS.
- Disable allocation of staging GLA and GLB registers (unless
  KeepDirectToLdsAlloc=1).  Can save ~16 VGPRS depending on
  tile size.
- Separate computation of lds write offset into calculateLdsWriteOffset
- Write m0 with LDS offset
- Test with new assembler,
- Uses dummy vgpr for DirectToLds

- Enable tail loop
- Increment m0 only when necessary
- Cleanup
- Only enable DirectToLds if coalesced loads fetch 256 consecutive bytes
  of memory.

### DIFF
--- a/Tensile/KernelWriterAssembly.py
+++ b/Tensile/KernelWriterAssembly.py
@@ -20,7 +20,7 @@
 ################################################################################
 
 from SolutionStructs import DataType
-from Common import globalParameters, printExit, printWarning, print2
+from Common import globalParameters, printExit, printWarning
 from KernelWriter import KernelWriter
 from math import log, ceil
 
@@ -41,11 +41,16 @@ class MemoryInstruction:
     self.endLine = "\n"
   ########################################
   # write in assembly format
-  def toString(self, params, comment, nonTemporal=0, highBits=0):
+  def toString(self, params, comment, nonTemporal=0, highBits=0, lds=0):
     name = self.name
     if highBits:
       name += "_d16_hi"
-    instStr = "%s %s" % (name, (self.formatting % params) )
+    # Remove destination from format string
+    # TODO-LDSDestBug - don't set lds flag to 0 here:
+    lds = 0
+    f = self.formatting.split(",",1)[-1] if lds else self.formatting
+
+    instStr = "%s %s" % (name, (f % params) )
     if nonTemporal%2==1:
       instStr += " glc"
     if nonTemporal/2==1:
@@ -251,8 +256,9 @@ class KernelWriterAssembly(KernelWriter):
     self.do["Sync"]        = True
     self.do["MAC"]         = True
     self.do["PostLoop"]    = True
+    self.do["KeepDirectToLdsAlloc"] = False  # If true, keep regs used for LDS alloc even if not used
 
-    # Various debug flags and modes 
+    # Various debug flags and modes
     self.db = {}
     self.db["EnableAsserts"]     = True  # Enable assertion codegen
     self.db["DebugKernelMaxItems"] = 16  # Capture first N(=16) print values, ignore subsequent.  If -1, debug writing is faster but writing more than 16 values is undefined.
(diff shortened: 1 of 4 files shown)


Tensile/KernelWriterAssembly.py
```diff
@@ -41,11 +41,16 @@ def __init__(self, name, numAddresses, numOffsets, \
...
-  def toString(self, params, comment, nonTemporal=0, highBits=0):
+  def toString(self, params, comment, nonTemporal=0, highBits=0, lds=0):
     name = self.name
     if highBits:
       name += "_d16_hi"
-    instStr = "%s %s" % (name, (self.formatting % params) )
+    # Remove destination from format string
+    # TODO-LDSDestBug - don't set lds flag to 0 here:
+    lds = 0
+    f = self.formatting.split(",",1)[-1] if lds else self.formatting
```
jgoldsAMD: Grr cannot select more than one line. If lds is set to 0, how does this line work?

Tensile/KernelWriterAssembly.py
```diff
@@ -400,6 +406,25 @@ def initKernel(self, kernel, tPA, tPB ):
 
     # True=slightly fewer [v_mov] instructions but extra registers
     self.globalReadIncsUseVgpr = False if kernel["BufferLoad"] else True
+    self.directToLds = 1
```
jgoldsAMD: Do we want to conditionally enable this as is done above?

Tensile/KernelWriterAssembly.py
```diff
@@ -2964,6 +2985,16 @@ def globalReadDo(self, kernel, guardK, tP):
     graIdx = 0
     g2lIdx = 0
     loadWidth = tP["globalReadInstruction"].totalWidth
+    ldsOffset = 0
+
+    if kernel["DirectToLds%s"%tP["tensorChar"]]:
+      # DirectToLds only enabled for TLU=1 cases, where ther registers are directly copied into LDS
```
jgoldsAMD: Typo "ther".

Tensile/KernelWriterAssembly.py
```diff
@@ -2964,6 +2985,16 @@ def globalReadDo(self, kernel, guardK, tP):
...
+      lwaSgpr = self.getTmpSgpr(1) 
+      kStr += inst("v_readfirstlane_b32", sgpr(lwaSgpr), \
+          vgpr("LocalWriteAddr%s"%tP["tensorChar"]), \
+          "Set lds write address to SGPR")
+      kStr += inst("s_mov_b32", "m0", sgpr(lwaSgpr), "Set LDS write address")
+      #kStr += inst("s_nop", "0",  "m0 wait state")
```
jgoldsAMD: When might this be needed?

Tensile/KernelWriterAssembly.py
```diff
@@ -3065,7 +3109,21 @@ def globalReadDo(self, kernel, guardK, tP):
                         "0 idxen offset:0",\
                         "load single f16")
                   elif kernel["ProblemType"]["DataType"].isSingle():
-                    kStr += inst("buffer_load_dword", \
+                    if kernel["DirectToLds%s"%tP["tensorChar"]]:
+                      tmpVgpr = self.vgprPool.checkOut(1)
+
+                      # TODO-LDSDestBug.  Remove dummy tmpVgpr when LDS dest asm bug fixed:
```
jgoldsAMD: Looks like this refers to the above code where I commented on lds being set to 0. It would be great if there were some way to tie the need for this workaround to the version of the assembler. Of course, we would need to know the working version number to set the appropriate cutoff! Or, can we try to assemble assuming the bug is fixed and use a failure to trigger the workaround?

Tensile/KernelWriterAssembly.py
bensander: thx, good catch.   LLVM matches SP3 assembler and requires a dummy destination VGPR even for direct-to-lds.  I will remove this lds=0 in the next checkin, i think we just live with the asm syntax we have.

PR #193, opened 2018-03-26
Feature precise bounds check

bensander:
Use buffer load num_records field to precisely check if load for A and B are in-bounds or not.  Eliminates need for pointer shifting.  

This PR includes work in refactor-sgpr and direct-to-lds fixes.

### DIFF
--- a/Tensile/KernelWriter.py
+++ b/Tensile/KernelWriter.py
@@ -274,7 +274,7 @@ class KernelWriter:
         kStr += self.globalReadIncrement(kernel, self.unrollIdx, \
             tensorParametersB)
       if self.enable["Wait"]:
-        kStr += self.wait(kernel, tensorParametersA, tensorParametersB, 0, -1, -1, "wait for global read")
+        kStr += self.wait(kernel, tensorParametersA, tensorParametersB, 0, -1, -1, "3wait for global read")
       if self.enable["LocalWrite"]:
         # local write
         kStr += self.comment("local write a")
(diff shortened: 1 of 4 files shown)


Tensile/KernelWriter.py
```diff
@@ -274,7 +274,7 @@ def kernelBody( self, kernel, tensorParametersA, tensorParametersB ):
         kStr += self.globalReadIncrement(kernel, self.unrollIdx, \
             tensorParametersB)
       if self.enable["Wait"]:
-        kStr += self.wait(kernel, tensorParametersA, tensorParametersB, 0, -1, -1, "wait for global read")
+        kStr += self.wait(kernel, tensorParametersA, tensorParametersB, 0, -1, -1, "3wait for global read")
```
jgoldsAMD: Aren't these comment changes in PR#192?

Tensile/KernelWriterAssembly.py
```diff
@@ -3035,25 +3031,27 @@ def globalReadDo(self, kernel, guardK, tP):
...
-          sgpr(maxAddrSgpr+0), \
-          sgpr("Sizes%s+%u"%("Sum" if sizeIdxIsSum else "Free", sizeIdx)),  \
-          sgpr("Strides%s+%u"%(tP["tensorChar"],strideIdx)), \
-          "mul d%u lower"%dim)
-
-      if kernel["BufferLoad"]:
-        offsetVgpr = self.vgprPool.checkOut(1)
-      else:
+      # TODO-64B:
+      # Assumes the product of the two sizes is <4GB here.
```
jgoldsAMD: Can you add an assert to verify the product is small enough? Will help catch problems later in case we forget.

bensander: we need one on the host side - don't want to add logic in the kernel if we can avoid it.

PR #197, opened 2018-03-28
fix seemingly random alloc of multiples of 2880*2880*2880 bytes

zaliu:
The seemingly random alloc failure to allocate ~ 286GB was caused by the defaultProblemSizes assuming the default Batched is False.  When Batched is True as specified by the user, and defaultProblemSizes is used, calculation of memory allocation mistook the third dimension as batch index, hence the wrong huge memory size.  One fix is to have a defaultBatchedProblemSizes when Batch is True.

pre-checkin.py passed; nightly.py has no new failures caused by this fix.


bensander:
Is there a way to override the defaultProblemSize with a user-specified problem size?   If so which part of the YAML file allows this?

PR #201, opened 2018-04-03
Feature buffer store2

bensander:
Remove ~85% of the logic for global write address computation.


bensander:
Will rebase in separate PR

PR #214, opened 2018-04-16
add mi25 Device 6860 to vega10

amcamd:
Add mi25  Device 6860  to list of vega10 devices so that all tuning for vega10 is also used for mi25.


bensander:
Lgtm

PR #244, opened 2018-05-11
Only use PreciseTiming on Assembly path

bensander:


zaliu:
LGTM.  I ran 2 HIP yaml files with this change in place, both runs produced valid timings in both Phase 1 and Phase 4.

PR #261, opened 2018-05-25
Multi-thread the kernel writing, provides 3X-4X speedup

bensander:


pfultz2:
The GIL limits the cpu utilization with threads. I assume this sets `CpuThreads` to 4 because performance doesn't get better with more threads. You might get better performance using [`multiprocessing.Pool.map`](https://docs.python.org/2/library/multiprocessing.html#multiprocessing.pool.multiprocessing.Pool.map) instead.

bensander:
Correct, speedup seems to hit max at 3-4 CPUs.   Not sure if it is GIL or the synchronization remaining in the code.  May give multiprocessing module a try in the future.

PR #276, opened 2018-06-26
__hfma no longer needed

zaliu:
Identical assembly code was generated using either __hfma or C += A*B.
There is no need to depend on __hfma anymore.


amcamd:
I see there are tests in nightly,  thanks

PR #283, opened 2018-07-17
intercept solution selection logic and call HIP kernel

amcamd:
Call HIP kernel in place of assembly kernel when strideA2K == 0 || strideB2k == 0


amcamd:
replace with pull request #285

PR #320, opened 2018-08-27
fix RuntimeLanguage with OCL issue

micytw:
test in KernelLanguage with Source case, the Assembly case not support.
test by rocblas_sgemm_asm_single_kernel.yaml with modification
1. RuntimeLanguage : OCL
2. KernelLanguage = ["Source"]


bensander:
Lgtm. We should merge.  
Also add a cl runtime year to nightly or pre checkin

PR #326, opened 2018-08-31
Disable printf, don't interleave unlesss dotInterleave>1

bensander:


Tensile/KernelWriterAssembly.py
```diff
@@ -2884,7 +2884,7 @@ def lwaFirstOffset(self, kernel, tP):
           hex(kernel["MacroTile%s"%tP["tensorChar"]] + kernel["LdsPad%s"%tc]), \
           vgpr(destVgpr), \
           "lw%s%s**(MT%s + PAD)"%(tP["tensorChar"], self.unrollChar, tP["tensorChar"]))
-    if dotInterleave:
+    if dotInterleave>1:
```
jgoldsAMD: Doesn't this break Vega10? I thought I tried this in my tree a while back and it caused issues.

PR #354, opened 2018-09-27
Refactor solution lookup3

bensander:


Tensile/Common.py
```diff
@@ -452,7 +455,11 @@
     {"UseSgprForGRO":             [ -1 ] },
     {"AssertSummationElementMultiple": [ 1 ] },
     {"AssertFree0ElementMultiple": [ 1 ] },
+<<<<<<< HEAD
```
amcamd: need to fix merge conflict

Tensile/SolutionStructs.py
```diff
@@ -1537,6 +1537,16 @@ def assignDerivedParameters(state):
           reject(state, "GlobalLoadVectorWidthB %u must be == VectorWidth %u or == 1" % \
                   (state["GlobalLoadVectorWidthB"], state["VectorWidth"]))
```
amcamd: I am not sure if I understand  the logic,  I was thinking it would be something like:
if state["KernelLanguage"] == "Assembly" and state["VectorWidth"] > 1:
  state["AssertMinApproxSize"] = 2
else:
  state["AssertMinApproxSize"] = 0

Tensile/Common.py
bensander: I think this was cleaned up in a later checkin.  Given the scope of changes may be best to review final set.  The intermediate checkins are to help isolate failures if that becomes necessary

Tensile/SolutionStructs.py
bensander: Can you review the final version:

    if state["KernelLanguage"] == "Assembly":
      # Asm kernels only work if all dims are > 32
      state["AssertMinApproxSize"] = 1
    elif state["VectorWidth"] > 1:
      # VW>1 kernels require dims>1
      state["AssertMinApproxSize"] = 2
    else:
      # these work everywhere, no special restrictions
      state["AssertMinApproxSize"] = 0

Tensile/Source/TensileTypes.h
```diff
@@ -77,106 +81,311 @@ TensileStatus tensileTeardown();
...
+};
+
+
+template <class ProblemKeyType>
+struct EuclideanDistance {
+  double operator() (const ProblemKeyType &p1, const ProblemKeyType &p2) const
+  {
+    double distance = 0;
+    for (int i=0; i<p1.numSizes(); i++) {
+      if (p1.sizes(i) > p2.sizes(i))
```
amcamd: can this be speeded up with:
 
double t = p1.sizes(i) - p2.sizes(i);
distance += t * t;

bensander: sizes are unsigned and in some case p2 may be bigger than p1 - subtracting would create large unsigned number.  This code is not often on critical path.

bensander:
Replace with #400

PR #358, opened 2018-10-02
Include atomic for usage with _hipFunctions

aaronenyeshi:
Fixes an error for implicit instantiation of undefined std::atomic<hipFunction_t*> _hipFunctions. We need to include this library if we want to use std::atomic .

We see this bug in compiling rocBLAS with latest HIP-Clang. I've tested this on Ubuntu 16.04, on Vega10 card with HIP-Clang, latest llvm/clang and ROCm-Device-Libraries.

### DIFF
--- a/Tensile/Source/SolutionHelper.h
+++ b/Tensile/Source/SolutionHelper.h
@@ -28,6 +28,7 @@
 #include <string>
 #include <tuple>
 #include <mutex>
+#include <atomic>
 
 /*******************************************************************************
  * Kernel Cache


aaronenyeshi:
Fixes this type of error:
```
In file included from /root/rocBLAS/build/release/Tensile/Solutions.cpp:28:
In file included from /root/rocBLAS/build/release/Tensile/Solutions.h:30:
/root/rocBLAS/build/release/Tensile/SolutionHelper.h:43:31: error: implicit instantiation of undefined
      template 'std::atomic<ihipModuleSymbol_t **>'
  std::atomic<hipFunction_t*> _hipFunctions;
                              ^
/usr/lib/gcc/x86_64-linux-gnu/5.4.0/../../../../include/c++/5.4.0/bits/atomic_base.h:129:12: note:
      template is declared here
    struct atomic<_Tp*>;
           ^
```

PR #360, opened 2018-10-06
disable one hanging TT configuration; add one large k config to hgemm_asm_tn

zaliu:
-  disable one hanging TT configuration in test_hgemm_tn_tt for now
-  add one large k configuration to hgemm_asm_tn


zaliu:
Given that quite a number of Tensile CI runs are timing out, I'm merging this PR. It should help.

PR #368, opened 2018-10-11
for HPA cast multiplicands before multiplication

amcamd:
Add static cast for multiplicands in HPA MAC. For HPA MAC below has 32 bit c and 16 bit a and b.

c += a*b

MAC is now defined as    c += static_cast<float>(a) * static_cast<float>(b)


amcamd:
Correct, changed to existing MAC macro rather than define new MAC_HPA

PR #377, opened 2018-10-19
Initial int8 code

amcamd:
- initial i8 gemm in source code
- use __builtin_amdgcn_sdot4
- add int8x4 and int32 to SolutionStructs.py DataType
- add TensileInt8x4 and TensileInt32 to TensileTypes.h
- add DestDataType separate from DataType
- passes a single trivial test


Tensile/Tests/nightly/assertions/test_hgemm_asem2_asm.yaml
```diff
@@ -23,6 +23,7 @@ BenchmarkProblems:
     - # ProblemType
       OperationType: GEMM
       DataType: h
+      DestDataType: h
```
bensander: Would a default assignment work here (if DestDataType not defined then defaults to DataType) ?  Would be less disruptive in long run for existing YAML files.

amcamd:
In mixed precition if DestDataType is higher precision than DataType it should be possible to get highest precision by applying alpha to AB after it is calculated in DestDataType. This is equivalent to calculating  AB  = A*B,  D = alpha * AB + beta C. Here AB is in DestDataType. An alternative would be to calculate alphaA = alpha * A (in DataType), then alphaAB = A*B ,  then  D = alphaAB + beta * C , this would give the same or lower precision.

PR #408, opened 2018-11-29
add -mno-code-object-v3 when needed

zaliu:


zaliu:
Yes, this fix works with both 1.9.1 and 2.0 (roc-master 9479).

PR #413, opened 2018-12-11
D matrix write

yoichiyoshida:
Preliminary support for matrix D write. Only supports D with the same ld and stride as C.


Tensile/KernelWriterSource.py
```diff
@@ -1217,6 +1221,7 @@ def graFinalOffsets(self, kernel, tP):
   ##############################################################################
   def graApplyUserOffsets(self, kernel):
     kStr = ""
+    kStr += "  D += offsetC;%s" % self.endLine
```
amcamd: do you need an offsetD?

PR #416, opened 2018-12-13
assembly i8 i32 gemm 

amcamd:
- The assembly code only runs on gfx906, for other architectures, source code will be used.
- The tests do not run as part of CI because CI machines are gfx900. The tests however pass on gfx906 machines.
- The following instructions are used:
  - v_dot4_i32_i8 : for mixed precision i32 += i8\*i8 + i8\*i8 + i8\*i8 + i8\*i8
  - v_mul_lo_u32 : for i32 multiply by alpha and beta
  - v_add_i32 : for i32 addition of alpha*A*B and beta*C
- Note that there is no saturation arithmetic in this version, so if there is any overflow of the i32 range the result will be corrupted

### DIFF
--- a/Tensile/KernelWriterAssembly.py
+++ b/Tensile/KernelWriterAssembly.py
@@ -1585,13 +1585,8 @@ class KernelWriterAssembly(KernelWriter):
                 cStr = "v[%s+%u+%u*%u]" % ("vgprValuC", a, b, kernel["ThreadTile0"])
                 aStr = "v[%s+%u]"       % ("vgprValuA_X%u_I%u"%(m,iui), a)
                 bStr = "v[%s+%u]"       % ("vgprValuB_X%u_I%u"%(m,iui), b)
-# this will be same as "single precision" except for the dot4 instruction replacing the v_mac_f32 instruction
-#               kStr += "v_mac_f32 %s, %s, %s%s" % (cStr, aStr, bStr, self.endLine) # single precision
                 kStr += "v_dot4_i32_i8  %s, %s, %s, %s op_sel:[0,0] op_sel_hi:[1,1] //valuC[%u]%s" % (cStr, aStr, bStr, cStr, cidx, self.endLine)
 
-                if macIdx == kernel["PerformanceWaitLocation"]:
-                    kStr += "s_waitcnt lgkmcnt(%u) // extra wait for performance%s" \
-                        % (kernel["PerformanceWaitCount"], self.endLine)
                 if macIdx == kernel["PerformanceSyncLocation"]:
                     kStr += "s_barrier // extra barrier for performance%s" \
                         % (self.endLine)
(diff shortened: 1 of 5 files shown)


Tensile/KernelWriterAssembly.py
```diff
@@ -1568,6 +1571,32 @@ def defineMACMacro(self, kernel, innerUnroll):
...
+              for iui in range(0, innerUnroll):
+                cidx = a + b*kernel["ThreadTile0"] + 0
+                cStr = "v[%s+%u+%u*%u]" % ("vgprValuC", a, b, kernel["ThreadTile0"])
+                aStr = "v[%s+%u]"       % ("vgprValuA_X%u_I%u"%(m,iui), a)
+                bStr = "v[%s+%u]"       % ("vgprValuB_X%u_I%u"%(m,iui), b)
+# this will be same as "single precision" except for the dot4 instruction replacing the v_mac_f32 instruction
+#               kStr += "v_mac_f32 %s, %s, %s%s" % (cStr, aStr, bStr, self.endLine) # single precision
+                kStr += "v_dot4_i32_i8  %s, %s, %s, %s op_sel:[0,0] op_sel_hi:[1,1] //valuC[%u]%s" % (cStr, aStr, bStr, cStr, cidx, self.endLine)
+
+                if macIdx == kernel["PerformanceWaitLocation"]:
```
bensander: I suspect we don't need either of these waits anymore

Tensile/KernelWriterAssembly.py
```diff
@@ -6559,6 +6631,10 @@ def globalWriteBatch(self, kernel, beta, edge, lsu, atomic, gwvw, atomicW, \
             else:
               kStr += self.chooseGlobalLoad(useBuffer, bps, sumIdx, \
                         addr0, addr1, soffset=0, offset=0, extraFields="", hi16=0)
+          elif kernel["ProblemType"]["DataType"].isInt8x4():
+            # assume this is the same as for isSingle
+            kStr += self.chooseGlobalLoad(useBuffer, bps, sumIdx, \
+                      addr0, addr1, soffset=0, offset=0, extraFields="")
```
bensander: Could make this an "OR" since the code is same on both Int8x4 and Single?

Tensile/KernelWriterAssembly.py
```diff
@@ -6528,6 +6596,10 @@ def globalWriteBatch(self, kernel, beta, edge, lsu, atomic, gwvw, atomicW, \
             else:
               kStr += self.chooseGlobalStore(useBuffer, bps, sumIdx, rpv, \
                         addr0, addr1, 0, ntStr, hi16=0)
+          elif kernel["ProblemType"]["DataType"].isInt8x4():
+            # assume this is the same as for isSingle
+            kStr += self.chooseGlobalStore(useBuffer, bps, sumIdx, rpv, \
+                      addr0, addr1, 0, ntStr)
```
bensander: Use OR?

Tensile/KernelWriterAssembly.py
amcamd: I removed the wait, and tests pass, this is limited testing. Can the s_barrier also be removed?

Tensile/KernelWriterAssembly.py
amcamd: Changed to use OR

Tensile/KernelWriterAssembly.py
amcamd: Changed to use OR

jgoldsAMD:
No good reason why putting extra waitcnts or barriers in the middle of your math would make things fast or correct. Make sure we have appropriate waitcnts and/or barriers elsewhere.

Tensile/KernelWriterAssembly.py
amcamd: Also removed s_barrier

PR #421, opened 2018-12-19
Improve runtime for Tensile by caching solution names

sdquiring:
I first tried this in October but couldn't get things exactly working before I got pulled off the project.  Found some time yesterday to give this another shot.

Basically a lot of time is spent right now in a few loops where we are doing something like

    for solution in solutionsForSchedule:
      if solution not in solutions:
        solutions.append(solution)

in a few places.  This is inefficient because Solution uses the `__str__` operator to do equality comparison which is inefficient, and this is done n^2 times in this loop.

By lazily caching the result of the `__str__` operator we can dramatically improve performance.

Without this change, rocBLAS compiles in 21 minutes, with this change, 8:42.

This is based on a bit of profiling I did using py-spy.   I think there is more room for improvement, but that would be a bigger task.  Our current method of building a string using repeated `+=` operations is inefficient because Python strings are immutable, meaning that it has to copy the entire string for every `+=` operation.  It would be much faster to use a `cStringIO` object, or Python template strings, or something else, but that's a much bigger can of worms.

### DIFF
--- a/Tensile/LibraryLogic.py
+++ b/Tensile/LibraryLogic.py
@@ -106,8 +106,8 @@ def analyzeProblemType( problemType, problemSizeGroups, inputParameters ):
   print1("# Solutions Used:")
   for i in range(0, len(logicAnalyzer.solutions)):
     s = logicAnalyzer.solutions[i]
-    s.state["SolutionIndex"] = i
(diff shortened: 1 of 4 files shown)


saadrahim:
LGTM, all the tests pass. This is really needed to reduce loading on Jenkins. The 12 minutes saved per PR will add up fast.

PR #430, opened 2019-01-04
FLOPS to DLOPS for int8 igemm

amcamd:
- report performance foe int8 igemm in DLOPS in place of FLOPS
- DLOPS is for Deep Learning OperationS
- 1 DLOP is the same one call to v_dot4_i32_i8


jgoldsAMD:
Isn't it better to report OPS instead of DLOPS? Reporting OPS would be consistent with how we report packed math and DOT2 perf.

amcamd:
Change to use OPS in place of DLOPS

PR #461, opened 2019-01-22
Opt store addr vgpr

bensander:


Tensile/KernelWriterAssembly.py
```diff
@@ -6386,12 +6508,19 @@ def globalWriteBatch(self, kernel, beta, edge, lsu, atomic, gwvw, atomicW, \
     elementMask = []
     elementSumIdx = []
     lastData = None
+    lastd1 = lastv1 = None
```
yoichiyoshida: these don't look to be used for anything

PR #466, opened 2019-01-29
fix merge tensile script to handle duplicate solution configurations

wbgilmartin:
updated the merge_rocblas_yaml_files.py in Utilities to fix a bug in this script.

It was not handling the cases when the exact file had solutions which were duplicates in the original logic file.


wbgilmartin:
Made updates to fix the null range logic issue.

PR #475, opened 2019-02-20
Adding a pre-commit hook which must be manually added

saadrahim:
pre-commit hook based on the style from rocBLAS. Jenkins will eventually check for style. A PR with the entire repo formatted to pass checks will be added soon.

Install the pre-commit hook by running ./githooks/install

- clang format based on rocBLAS with feedback from Scott. A sample C++ file attached
[sample.cpp.txt](https://github.com/ROCmSoftwarePlatform/Tensile/files/2886888/sample.cpp.txt)

- python format checks via yapf (https://github.com/google/yapf). Python style follow https://www.python.org/dev/peps/pep-0008/#tabs-or-spaces

PR #476, opened 2019-02-20
After formatting based on new pre-commit hook and future Jenkins tests

saadrahim:
This is what Tensile will look like after applying https://github.com/ROCmSoftwarePlatform/Tensile/pull/475


saadrahim:
Does not work now, closing request

saadrahim:
Trying just C++ now

PR #487, opened 2019-03-11
fix for bug swdev-180337. can not build with ROCm2.2 RC1.

wbgilmartin:
fixes swdev-180337

moves the hip device calls out of the static constructors for the kernel mapping logic


Tensile/TensileCreateLibrary.py
```diff
@@ -654,8 +684,8 @@ def writeSolutionAndExactTable(scheduleName, deviceNames, schedProbName, problem
   s += "// The entrypoint to find a solution for this problem is through the master solution master\n"
   s += "static SolutionMapper_%s solutionMapper_%s(\n" % (problemType, schedProbName)
   s += "  \"%s\", // schedule+problem name\n" % (schedProbName) 
-  s += "  {%s}, // Device names\n" % (', '.join('"{0}"'.format(w) for w in deviceNames))
-  s += "  &masterSolutionMapper_%s, // add to this master solution mapper\n" % (problemType)
+  #s += "  {%s}, // Device names\n" % (', '.join('"{0}"'.format(w) for w in deviceNames))
```
amcamd: Can you remove these two lines rather than comment them out

Jenkinsfile
```diff
@@ -1,426 +0,0 @@
-#!/usr/bin/env groovy
```
amcamd: Does this Jenkinsfile need to be removed? Was it removed for testing purposes and for this commit it should not be removed?

wbgilmartin: Yes. 

This solution requires synchronization of the changes to both rocBLAS and Tensile. It is expected that the tests are going to fail and so this pr is to provide a review only request.

I approved the intent is to submit a valid pr which can get submitted quickly and merged, hopefully without breaking anything.

PR #490, opened 2019-03-13
Upgrading rocm 2.2

saadrahim:
Upgrading to build on rocm 2.2

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -328,7 +328,7 @@ parallel rocm22_ubuntu_gfx900:
         docker_build_args:' --pull' )
 
     def hcc_compiler_args = new compiler_data(
-        compiler_name:'hcc-rocm20-ubuntu',
+        compiler_name:'hcc-rocm22-ubuntu',
         build_config:'Release',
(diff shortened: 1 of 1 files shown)


amcamd:
The changes look good. Can you confirm that it runs CI on both gfx900 and gfx906. The CI build or test may fail due to running on ROCm 2.2, and we will need to fix it.

saadrahim:
> 
> 
> The changes look good. Can you confirm that it runs CI on both gfx900 and gfx906. The CI build or test may fail due to running on ROCm 2.2, and we will need to fix it.

Looks like it is building on both gfx900/gfx906. I am waiting test to finish and then will merge.

saadrahim:
> 
> 
> > The changes look good. Can you confirm that it runs CI on both gfx900 and gfx906. The CI build or test may fail due to running on ROCm 2.2, and we will need to fix it.
> 
> Looks like it is building on both gfx900/gfx906. I am waiting test to finish and then will merge.

CI is hung on linking the client executable. I am merging but we need to look into this.

PR #495, opened 2019-03-14
check __hcc_workweek__ >= 19015 before calling amd_mixed_dot

amcamd:
- new instruction amd_mixed_dot was added after __hcc_workweek__ 19015
- add contitional compile based on __hcc_workweek__ >= 19015) || __HIP_CLANG_ONLY__ for amd_mixed_dot


b-sumner:
LGTM

PR #499, opened 2019-03-25
add metadata to assembly kernels

amcamd:
Add Code Object V2 and Code Object V3 metadata as described here

https://llvm.org/docs/AMDGPUUsage.html#code-object-metadata

### DIFF
--- a/Tensile/KernelWriterAssembly.py
+++ b/Tensile/KernelWriterAssembly.py
@@ -1434,6 +1434,25 @@ class KernelWriterAssembly(KernelWriter):
   def macroRegister(self, name, value):
     return ".set %s, %s%s" % (name, value, self.endLine)
 
+  def v2Argument(self, name, size, align, valueKind, valueType, AddrSpaceQual = None):
+    kStr = ""
+    kStr += "      - Name:            %s\n" % name
+    kStr += "        Size:            %s\n" % size
+    kStr += "        Align:           %s\n" % align
+    kStr += "        ValueKind:       %s\n" % valueKind
+    kStr += "        ValueType:       %s\n" % valueType
+    if AddrSpaceQual != None:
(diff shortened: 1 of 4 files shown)


Tensile/KernelWriterAssembly.py
```diff
@@ -1814,6 +1833,164 @@ def functionSignature(self, kernel ):
...
+        srcAlign = "2"
+        srcValueType = "F16"
+        dstSize = "2"
+        dstAlign = "2"
+        dstValueType = "F16"
+        cptSize = "2"
+        cptAlign = "2"
+        cptValueType = "F16"
+    
+    elif kernel["ProblemType"]["DataType"].isInt8x4():
```
yxsamliu: is this a vector of char or array of char in the function declaration?

amcamd: kernel is a Python dict,

Tensile/KernelWriterAssembly.py
```diff
@@ -1814,6 +1833,164 @@ def functionSignature(self, kernel ):
...
+      kStr += ".amd_amdgpu_hsa_metadata\n"
+      kStr += "Version: [ 1, 0 ]\n"
+      kStr += "Kernels:\n"
+      kStr += "  - Name: %s%s" % (self.kernelName, self.endLine)
+      kStr += "    SymbolName: '%s@kd'%s" % (self.kernelName, self.endLine)
+      kStr += "    Language: OpenCL C\n"
+      kStr += "    LanguageVersion: [ 2, 0 ]\n"
+      kStr += "    Args:\n"
+
+      kStr += self.v2Argument(                               'D', dstSize, dstAlign, "GlobalBuffer", dstValueType, "Generic")
```
yxsamliu: For buffers, size and align should be size and align of the pointer type itself, not the pointee type. Therefore size and align for buffer should always be 8.

PR #519, opened 2019-05-02
Updates to cmake configuration for tensile

pfultz2:
This updates the cmake for Tensile based on the `static-host-code` branch.

It removes the broken version check in TensileConfigVersion.cmake.

It adds a `tensile_add_library` function to create the library

* This changes the interface to use cmake argument parsing so `tensile_add_library(tensile ${LOGIC_DIR} LIBRARY_PRINT_DEBUG)` can be written instead of passing in a bunch of empty strings
* It takes a library name from the user to create the library so it can avoid possible name conflicts.
* It doesnt change the `CXX` compiler in the cmake.
* It doesn't search for hip or opencl. Instead the user is responsible for setting the usage requirements(ie `target_link_libraries(tensile PRIVATE hip::device --amdgpu-target=gfx900)`).

I also added tests to Tensile for the cmake integration which should help ensure things wont break in the future.


sdquiring:
Thanks for this.  Unfortunately it duplicates some of my own efforts.  

I've implemented my own CMake function to create a Tensile kernel library:
https://github.com/sdquiring/Tensile/blob/69888f02289a843d3ac4b658c6b5ad8bebf8f58c/Tensile/Source/TensileConfig.cmake#L142

The new static host library is not a part of what's created with this function.  There are two ways of using the result of this in a parent library:

 - Using `EMBED_LIBRARY` and the optional `EMBED_KEY` which creates a static library that embeds the code into your executable
 - Using the TENSILE_CODE_OBJECTS / TENSILE_LIBRARY_FILE variables set by this function that allow the calling CMake code to use these files in some other way.

I like the idea of testing CMake functionality in a pre-checkin test.

pfultz2:
Ok let me update it to work of off that. 

The `Tensile_ROOT` is set by cmake so it never needs to be set by the user. I also set it to the installation of Tensile instead of the `bin` directory in the installation(perhaps we want to access other files in the installation).

Maybe we can make `OUTPUT_PATH` an optional parameter and by default its a directory in the build directory. 

> Using EMBED_LIBRARY and the optional EMBED_KEY which creates a static library that embeds the code into your executable

What does `EMBED_KEY` do? I see `EMBED_LIBRARY` sets the cmake target name but it doesnt make it a static library though.

Also where does the `Tensile` target come from? The one that is used in `target_link_libraries(${Tensile_EMBED_LIBRARY} PUBLIC Tensile)`. I didnt see it defined anywhere.

pfultz2:
> Using the TENSILE_CODE_OBJECTS / TENSILE_LIBRARY_FILE variables set by this function that allow the calling CMake code to use these files in some other way.

Are these a build-only dependency? Or do we need to install these files with our library?

sdquiring:
Closing this pull request to relieve some pressure on our CI servers.  Feel free to re-open it if there's an update.

PR #524, opened 2019-05-06
bfloat16 arithmetic added to source kernels

amcamd:
- Testing gives correct answer for 8x8 gemm with A=B=C=D=1.
- This is an initial checkin for bfloat16 arithmetic. 
- Additional testing and development will follow. 
- The ReferenceCPU.h file will be refactored in a subsequent PR to use templated Type, DestType, and ComputeType in place of std::is_same<Type, tensile_bfloat16>().

### DIFF
--- a/Tensile/Configs/rocblas_hpa_bfloat16_hip_single_kernel.yaml
+++ b/Tensile/Configs/rocblas_hpa_bfloat16_hip_single_kernel.yaml
@@ -8,7 +8,7 @@ GlobalParameters:
   EnqueuesPerSync: 1
   SyncsPerBenchmark: 1
   LibraryPrintDebug: False
-  NumElementsToValidate: 0
+  NumElementsToValidate: 8
   ValidationMaxToPrint: 4
   ValidationPrintValids: False
   ShortNames: False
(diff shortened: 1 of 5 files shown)


Tensile/Configs/rocblas_hpa_bfloat16_hip_single_kernel.yaml
```diff
@@ -8,17 +8,26 @@ GlobalParameters:
...
+  NumElementsToValidate: 8
   ValidationMaxToPrint: 4
   ValidationPrintValids: False
   ShortNames: False
   MergeFiles: True
   Platform: 0
   Device: 0
   KernelTime: True
-  PinClocks: True
+  PinClocks: False
```
bensander: Did you mean to change this?
Perhaps could also change SleepPercent to 0 if this intended to be file for tuning?

amcamd: I was aiming to make it simpler. I am using this .yaml file to generate the simplest kernel and test I can think of for initial testing. 

The other _single_kernel.yaml files generate a kernel that gives good performance for large size. This file should change to a generating a kernel for large size as the bfloat16 implementation continues, and as other .yaml files are added for testing and tuning.

PR #526, opened 2019-05-07
Static host code refactor

sdquiring:
This branch includes:

 - A new statically built host runtime (under `lib`) which is under development but which can support SGEMM.
   - This code uses a YAML-based configuration (generally known as a "kernel library") that replaces the need to generate any host-side code.
   - There are some additions to the Python code to generate this new configuration file, and to generate code to embed this data into an executable.
   - Jenkins runs some unit tests on this host library
 - Python 3 support (thanks Akila)
 - Adding a Python UnitTests directory (also Akila)
 - Improved Jenkins integration which includes capturing the test results.

Currently, the new code is not used by default, the older code still works.  The plan is to replace the new code step by step.

Currently this is passing Jenkins:
http://hsautil.amd.com/job/ROCmSoftwarePlatform/job/Tensile/job/static-host-code/


leekillough:
Please "squash and merge" if you merge -- we don't want too many commits.

I'll review the changes and provide more feedback.

lib/include/Tensile/Comparison.hpp
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    template<typename A, typename... Args>
+    inline int LexicographicCompare(A const& lhs, A const& rhs, Args const&... rest)
+    {
+        if(lhs < rhs) return -1;
+        if(lhs > rhs) return  1;
+        return LexicographicCompare(rest...);
+    }
+
```
leekillough: I assume this is being called in pairs at a time, like (A1 <=> B1) || (A2 <=> B2) || (A3 <=> B3) ... returning the first pair which isn't identical.

I'd either add a `static_assert` or use SFINAE to make it only match an even number of arguments:

    template<typename A, typename... Args>
    inline int LexicographicCompare(A const& lhs, A const& rhs, Args const&... rest)
    {
        static_assert(sizeof...(Args) % 2 == 0, "LexicographicCompare requires an even number of arguments");
        if(lhs < rhs) return -1;
        if(lhs > rhs) return  1;
        return LexicographicCompare(rest...);
    } 
<p>

    #include<type_traits>

    template<typename A, typename... Args, typename = typename std::enable_if<sizeof...(Args) % 2 == 0>::type>
    inline int LexicographicCompare(A const& lhs, A const& rhs, Args const&... rest)
    {
        if(lhs < rhs) return -1;
        if(lhs > rhs) return  1;
        return LexicographicCompare(rest...);
    }

sdquiring: Good idea, will do.  Right now this is used to easily implement comparison operators for the Index classes (nested in ContractionProblem).

lib/include/Tensile/geom.hpp
```diff
@@ -0,0 +1,108 @@
...
+    template <typename T>
+    inline TENSILE_API std::ostream & operator<<(std::ostream & stream, vector4<T> const& v)
+    {
+        return stream << "(" << v.x << ", " << v.y << ", " << v.z << ", " << v.w << ")";
+    }
+
+    using dim3 = vector3<size_t>;
+    using int3 = vector3<int>;
+}
+
```
leekillough: I wonder if these vector2, vector3, vector4, etc. can be templatized or made generic with variadic templates and `std::array` or `std::tuple`.

Things like inheriting from `std::array<N>` and inheriting its constructors (`using array<N>::array`), then adding any extra methods, so that the class works like a proxy class to `std::array<N>`.

This is just a general comment for refactoring the code going forward; it's not a criticism of the current PR.

lib/source/Debug.cpp
```diff
@@ -0,0 +1,74 @@
...
+{
+    std::once_flag debug_init;
+
+    Debug & Debug::Get()
+    {
+        std::call_once(debug_init, [](){ instance = new Debug(); });
+
+        return *instance;
+    }
+
```
leekillough: It is generally faster to use local `static` variables than to call `call_once`, and it's more concise:

    auto& Debug::Get()
    {
          static auto* instance = new Debug();
          return *instance;
    }

(There's no need for the namespace `debug_init` or `instance` variables anymore.)

In C++11 and later, initialization of local `static` variables is guaranteed to be thread-safe (race-free).

lib/include/Tensile/Singleton.hpp
```diff
@@ -0,0 +1,61 @@
...
+    class LazySingleton
+    {
+    public:
+        static Class & Instance()
+        {
+            std::call_once(m_instanceConstructionFlag, Construct);
+
+            return *m_instance;
+        }
+
```
leekillough: Much simpler, without a need for `Construct()`, external `m_instance`, or `m_instanceConstructionFlag`:

    static auto& Instance()
    {
       static auto* m_instance = new Class();
       return *m_instance;
    }

lib/source/Debug.cpp
leekillough: Actually, you might not even need dynamic allocation:

    static auto& Debug::Get()
    { 
        static Debug instance;
        return instance;
    }

lib/include/Tensile/Singleton.hpp
leekillough: Like above, you might not need dynamic allocation:

    static auto& Instance()
    {
        static Class m_instance;
        return m_instance;
    }

sdquiring: I guess that's a question of if we want the objects to be instantiated at static initialization time, or at the time of first use.  Changing that behaviour would make this a StaticSingleton, not a LazySingleton.  We could have both.

lib/source/Debug.cpp
sdquiring: This was written before the LazySingleton class.  If we have both Static and Lazy implemented we can just select which behaviour we want by inheriting from the right parent class.

lib/include/Tensile/geom.hpp
sdquiring: Yes, I think that would make it cleaner.  I'm also now wondering how wise it is to create another class called `dim3`, if it might be a good idea to use a different name instead.

pfultz2:
> Please "squash and merge" if you merge -- we don't want too many commits.

You can squash when merging even from github.

leekillough:
> > Please "squash and merge" if you merge -- we don't want too many commits.
> 
> You can squash when merging even from github.

That is what I was referring to, hence the quotes.

lib/include/Tensile/Singleton.hpp
leekillough: The codes which used `call_once` were initializing their important data (not the `call_once_flag`) at time of first use; I'm merely suggesting that it be implemented using local `static` variables instead.

Initialization at startup is generally bad because initialization order depends on link or load order. With local `static` initialization occurs at time of first use and does not suffer from indeterminate initialization order.

`call_once` is slower than doing the equivalent with local `static` variables.

lib/source/Debug.cpp
leekillough: As I said in the other one, a local `static` is faster than using `call_once` and has the same semantics. I'm not talking about changing the semantics; just optimizing the original code to use local `static` to initialize at first use, rather than calling `call_once` or using `new`.

sdquiring: Done.

lib/include/Tensile/Comparison.hpp
sdquiring: Done.

lib/include/Tensile/Singleton.hpp
sdquiring: Done.

Tensile/BenchmarkStructs.py
```diff
@@ -18,26 +18,22 @@
...
-from SolutionStructs import Solution, ProblemType, ProblemSizes
+from .SolutionStructs import Solution, ProblemType, ProblemSizes
 
-################################################################################
-# Benchmark Process
-# steps in config need to be expanded and
-# missing elements need to be assigned a default
-################################################################################
+""" Benchmark Process
+    steps in config need to be expanded and
```
sdquiring: Docstrings need to go inside the class.

Tensile/Code.py
```diff
@@ -199,13 +222,14 @@ def formatWithComment(self, instStr, comment):
 
   def __str__(self):
     return self.text
+"""
+ Construct a waitcnt from specified lgkmcnt and vmcnt:
```
sdquiring: Move inside function.

Tensile/SolutionStructs.py
```diff
@@ -824,7 +678,10 @@ def setGlobalLoadVectorWidth(state, tc, totalVectors):
     validDepthU = True
     if totalVectors < state["NumThreads"]:
       # Try to reduce size of vector so every thread has a load to do
-      pv = state["NumThreads"] / totalVectors # partial vector
+      pv = state["NumThreads"]//totalVectors
+      #state["NumThreads"] / totalVectors # partial vector
+      # import pdb
```
sdquiring: Delete debug code.

Tensile/TensileCreateLibrary.py
```diff
@@ -19,62 +19,130 @@
...
+    path906 = soFilepath + '-000-gfx906.hsaco'
+
+    if os.path.exists(path900):
+	return [path900]
+    elif os.path.exists(path906):
+	return [path906]
+    raise RuntimeError("Could not create code object file.")
+
+def buildSourceCodeObjectFiles(kernelFiles, kernels, outputPath):
+    cpus = Common.CPUThreadCount()
```
sdquiring: Variable no longer used.

akilaMD:
Python 2 to 3 conversion summary
-	Lists were lazy in Python 2, but must be explicitly defined in Python 3. 
o	To define a list, use the list() function.
o	To return the length of a list, you must use len(list(<list_name>))

-	Division of any two numbers using the '/' character returns a float in Python 3. To return an integer, use the floor division character '//'. 
o	For Tensile, use the ceil_divide() and the roundUpToNearestMultiple() functions for ceiling division or for rounding up numbers, both of which are available in Utils.py. 

-	Print statement is now a function. Adding the Python 3 print() function causes the script to become incompatible with Python 2, so to keep Python 2 compatibility, add the following statement at the top of each script:

-	Unicode characters are now treated as strings in Python 3. Declare them by typing u'<unicode_val>'
o	Use 'from __future__ import unicode_literals' to keep Python 2 compatibility.

-	Similarly to unicode characters, bytes are now treated as strings in Python 3. Declare them by typing b'<byte_val>'
o	Use .decode() to return a string for a particular byte value. This will remove the b character.

-	Unlike Python 2, 'None' values cannot be compared to integers/floats/strings as they’re ‘NoneTypes’.
o	Use str(), int(), and float() when possible to explicitly define numeric types in Python 3

-	Basestrings have been removed and must be treated as regular strings without any special characters.  
-	When calling os.chmod or using any decimal characters, they must be declared using 0o instead of 0.
-	.iteritems() and .itervalues() have been replace with .items() and .values()
-	For metaclasses, import the six package & define @six.add_metaclass() as a property above the class.

-	When importing python scripts as packages, add "from <location>" in front of each import statement
o	Packages are no longer implicitly declared if an __init__.py file in the same directory.

PR #537, opened 2019-05-15
New Jenkinsfile using rocJenkins shared library

akilaMD:
- Only runs tox tests on py35
- Stage names are different (when compared to develop branch)
  1. "Build" stage is now "Docker" stage
  2. "Host Test" stage is now "Compile" stage
  3. Compiler name, OS names, and "Release" have all been removed
- Packaging stage should be skipped 
- Passed CI. See here: http://hsautil.amd.com/job/ROCmSoftwarePlatform/job/Tensile/view/change-requests/job/PR-537/
- "dockerfile-build-ubuntu-rock" is the same as "dockerfile-build-rocm-terminal". I did this because the shared library currently will only look for that specific file, otherwise it doesn't build anything.  
- Updated syntax changes to reflect Yoichi's suggestions


Jenkinsfile
```diff
@@ -1,4 +1,12 @@
 #!/usr/bin/env groovy
+// This shared library is available at https://github.com/ROCmSoftwarePlatform/rocJENKINS/
+@Library('rocJenkins') _
+
+// This is file for internal AMD use.
```
yoichiyoshida: English

docker/dockerfile-build-ubuntu-rock
```diff
@@ -0,0 +1,48 @@
...
+FROM ${base_image}
+LABEL maintainer="kent.knox@amd"
+
+USER root
+ARG user_uid
+
+# Install dependent packages
+# Dependencies:
+# * hcc-config.cmake: pkg-config
+# * tensile: python2.7, python-pip
```
yoichiyoshida: new dependencies are python3 and rock-dkms

docker/dockerfile-build-ubuntu-rock
```diff
@@ -0,0 +1,48 @@
+# This Dockerfile provides a starting point for a ROCm installation of hipCaffe.
```
yoichiyoshida: not building hipCaffe

docker/dockerfile-build-ubuntu-rock
```diff
@@ -0,0 +1,48 @@
+# This Dockerfile provides a starting point for a ROCm installation of hipCaffe.
+
+# Parameters related to building hip
+ARG base_image
+
+FROM ${base_image}
+LABEL maintainer="kent.knox@amd"
```
yoichiyoshida: kent is no longer the maintainer

docker/dockerfile-install-ubuntu
```diff
@@ -0,0 +1,43 @@
+# Parameters related to building hip
```
yoichiyoshida: I thought we weren't installing anything, why do we need this file?

akilaMD: Good catch. Yeah, I must've added this because it's defined in shared libraries as the install file, but Tensile doesn't use it. Let me delete the file.

Jenkinsfile
akilaMD: ???

docker/dockerfile-build-ubuntu-rock
akilaMD: I added rock-dkms to this dockerfile. python3 is already added, but python 2.7 should be removed as a dependency, for sure. Not sure this is worth another run in CI though...

docker/dockerfile-build-ubuntu-rock
akilaMD: he's maintaining rocBLAS too apparently! Should be changed (though I don't know who the real maintainer is...maybe Saad?), but I don't want this to trigger another CI build at the moment.

docker/dockerfile-build-ubuntu-rock
akilaMD: Yep. We really need to fetch dockerfiles directly from rocJenkins and not in every repo...

docker/dockerfile-build-ubuntu-rock
akilaMD: I should also note that dockerfile-build-ubuntu-rock is word-for-word the same file as dockerfile-build-rocm-terminal. To use the shared library, the build image must be named "dockerfile-build-ubuntu-rock"

Jenkinsfile
yoichiyoshida: check line 5

docker/dockerfile-build-ubuntu-rock
yoichiyoshida: I mean add any new dependencies you have in the comments, not just silently add it in the code. People reading through will get confused if you just add things without documentation

docker/dockerfile-build-ubuntu-rock
yoichiyoshida: this project is called Tensile... not hipCaffe

docker/dockerfile-build-ubuntu-rock
sdquiring: We should probably just delete this comment since it just duplicates the list of packages below.

akilaMD: Yeah, I've marked it as resolved for that reason. You've taken care of this file anyway in your PR

Jenkinsfile
akilaMD: ah, missing an a

PR #538, opened 2019-05-15
Drop Python 2 support

sdquiring:
As we discussed in the Tensile meeting yesterday, some of the failing Jenkins jobs are due to testing both python 2 and 3.  If we drop 2, we can stop it from running on Jenkins and clear up some of the backlog.

There's a parallel PR incoming for rocBLAS which switches the virtual environment created by CMake to a Python 3 based one.

Note that this will intentionally drop support for python 2, there will be syntax errors.  This is intentional in order to prevent any developers from writing code that is incompatible.


sdquiring:
Now passing in Jenkins.  Are there any objections to dropping Python 2? Also, what do we do to get this mentioned in release notes?

sdquiring:
If there are no objections, I'll merge this and the rocBLAS change today.

PR #539, opened 2019-05-16
Fix issue with multiple benchmark problem types.

sdquiring:
Also causing the client to error out if a kernel file doesn't exist.


sdquiring:
We've been trying to get the noCI label to prevent Jenkins from building a PR but it doesn't seem to be working.

PR #547, opened 2019-05-29
[hip] Add `__host__`/`__device__` attribute to get bfloat16 compiled on HIP/Clang

mhbliao:
Necessary change to get code using bfloat16 to be compiled on HIP/Clang


mhbliao:
rebased up to the latest bfloat16

PR #550, opened 2019-06-04
Add fix for strided GEMMs near the end of allocated memory

yoichiyoshida:
Fix the pre-calculated number of elements passed into GEMM kernels for the strided cases to not include the stride past the end.


yoichiyoshida:
All Tensile and rocBLAS CI tests pass on standalone gfx900 and gfx906 machines

PR #557, opened 2019-06-07
correction for metadata

amcamd:
- add two missing arguments for strideD

### DIFF
--- a/Tensile/KernelWriterAssembly.py
+++ b/Tensile/KernelWriterAssembly.py
@@ -2032,7 +2032,9 @@ class KernelWriterAssembly(KernelWriter):
       kStr += self.v2Argument(                               'B',     '8',      '8', "GlobalBuffer", srcValueType, "Generic")
       kStr += self.v2Argument(                           "alpha", cptSize, cptAlign,      "ByValue", cptValueType)
       kStr += self.v2Argument(                            "beta", cptSize, cptAlign,      "ByValue", cptValueType)
-      kStr += self.v2Argument(                       "strideC1J",     '4',      '4',      "ByValue",        "I32")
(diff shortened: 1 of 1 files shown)


yxsamliu:
I think there may be other issues with the metadata. The kernel args metadata has to follow the same order as the actual kernel args. In one of the execution of an assembly kernel I saw the actual kernel args are defined as the following:
  struct {
    // Size of Tensor's packed dims, in elements
    uint64_t tensor2dSizeC;
    uint64_t tensor2dSizeA;
    uint64_t tensor2dSizeB;
    float * dataD;
    const float * dataC;
    const float * dataA;
    const float * dataB;
    float alpha;
    float beta;
    unsigned int strideD1J;
    unsigned int strideD2K;
    unsigned int strideC1J;
    unsigned int strideC2K;
    unsigned int strideA1L;
    unsigned int strideA2K;
    unsigned int strideB1L;
    unsigned int strideB2K;
    unsigned int sizeI;
    unsigned int sizeJ;
    unsigned int sizeK;
    unsigned int sizeL;
    int staggerUIter;
    unsigned int problemNumGroupTiles0;
    unsigned int problemNumGroupTiles1;
    unsigned int magicNumberProblemNumGroupTiles0;
    unsigned int gridNumWorkGroups0;
    unsigned int numFullBlocks;
    unsigned int wgmRemainder1;
    unsigned int magicNumberWgmRemainder1;
    unsigned int pad;
  } hipFunctionArgs;

It seems the first 3 kernel args are missing in the metadata and also some other kernel args in the metadata do not match the actual kernel args. Therefore my question is:

Do all assembly kernels have the same kernel args with the same order?

Does the kernel arg metadata match the actual kernel args passed to the assembly kernels?

PR #564, opened 2019-06-12
corrections for metadata

amcamd:
- add missing arguments
- update V3 metadata strings


amcamd:
All Tensile pre_checkin and nightly tests pass on both gfx900 and gfx906.

PR #585, opened 2019-06-27
adding implementation of tile aware selection

wbgilmartin:
This implements the library part of the tile aware solution selection for tensile. 

adds a granularity selection library as an alternative to exact solution lookup along with the yaml serialization. 
implements the granularity computation in ContractionSolution::projectedPerformance.

implements a kernel test in RunGEMMKernelTileSelection_test.cpp and KernelsTileLite.yaml however this is disabled do to the fact that we don't have the python part which constructs the library (enabled with cmake)


lib/configs/SolutionLibraries/KernelsTileLite.yaml
```diff
@@ -0,0 +1,205 @@
+GlobalParameters:
```
sdquiring: The SolutionLibraries directory contains built solution libraries (that can be loaded by the host code).  This could go in 'configs' or in a new subdirectory of 'configs' along with sgemm_all_transposes.yaml.

lib/include/Tensile/GranularitySelection.hpp
```diff
@@ -0,0 +1,191 @@
+/**
```
sdquiring: This set of table / entry / library classes is kind of extraneous since we need to make use of the problem and solution classes directly.  It also prevents sending in the Hardware object in order to get the number of CUs.

lib/test/hip/RunGEMMKernelTileSelection_test.cpp
```diff
@@ -0,0 +1,253 @@
...
+    }
+};
+
+TEST_P(RunGEMMKernelSolutionSelectionTest , KernelsTileSelection)
+{
+
+    ContractionProblem problem = GetParam();
+    //std::cout << problem << std::endl;
+
+    auto library = LoadLibraryFile<ContractionProblem>(TestData::File("kernels_tile_selection/TensileLibrary.yaml").native());
```
sdquiring: Is this built manually?

lib/test/hip/RunGEMMKernelTileSelection_test.cpp
```diff
@@ -0,0 +1,253 @@
+/*******************************************************************************
```
sdquiring: I don't see a difference between this and the test cases in RunGEMMKernel_test.cpp.  Could this be added to that as another test case?

lib/test/hip/RunGEMMKernelTileSelection_test.cpp
wbgilmartin: Currently this is built manually, Until I get the python part that generates it for the logic and solution data.

lib/include/Tensile/GranularitySelection.hpp
wbgilmartin: I will have to rethink this a little bit

wbgilmartin: The last commit addresses this.

lib/test/hip/RunGEMMKernelTileSelection_test.cpp
wbgilmartin: I don't have this tests enabled so (it is not included in CMakeList.txt).

I don't have the python implementation which will generate the related solution. That will be part of my next pull request. When that is done I will enable the test.

lib/include/Tensile/GranularitySelectionLibrary.hpp
```diff
@@ -0,0 +1,131 @@
...
+            }
+
+            while(iter != solutions.end())
+            {
+                auto mySolution = iter->second;
+                double myPerformance = mySolution->projectedPerformance(problem);
+
+                if(mySolution)
+                {
+                    auto myPerformance = mySolution->projectedPerformance(problem);
```
sdquiring: Seems to be a duplicate of the above call.

lib/include/Tensile/GranularitySelectionLibrary.hpp
```diff
@@ -0,0 +1,131 @@
...
+            return rv;
+        }
+        
+        virtual std::shared_ptr<MySolution>
+            findBestSolution(MyProblem const& problem,
+                             Hardware  const& hardware) const override
+        {
+            const bool debug = Debug::Instance().printPropertyEvaluation();
+
+            this->description();
```
sdquiring: Did you mean to print this?

PR #590, opened 2019-07-11
add client and tuning updates for tile aware selection

wbgilmartin:
implements changes to the client code as well as the python code to tune for the tile aware selection. (determines the ideal performance for the given kernel and summation index)

This will get enabled by adding TileAwareSelection: true to the benchmark parameters for the group. (the default is false)

The will generated the measurement of the ideal performance for the kernels using (default) k = [32,64,96,128,256,512,1024,2048,4096,8192,16192]

This can be overridden with - SolutionSummationSizes: [32,64,96,128,256,512,1024,2048,4096,8192,16192] 

added to the BenchmarkFinalParameters

When enabled this will produce the TensileLibrary.yaml for the solution selection which gets added to the  3_LibraryLogic subdirectory as part of the tensile tuning.


wbgilmartin:
Close until 2.8

PR #592, opened 2019-07-15
Support float complex gemm by hip solutions

jichangjichang:
This PR make Tensile to support the cgemm and zgemm by hip solutions.

https://github.com/ROCmSoftwarePlatform/rocBLAS/pull/613
Tony also submitted this corresponding rocBlas PR.
Tensile PR must be merged first prior to rocBlas PR.

### DIFF
--- a/Tensile/Source/Client.h
+++ b/Tensile/Source/Client.h
@@ -291,7 +291,8 @@ void specializeData(
       if (db & 0x2) {
         std::cout << "[" << serialIdx << "] = " << val << "\n";
       }
-      initialData[serialIdx] = val++;  // actually initialize the element
+      initialData[serialIdx] = val; // actually initialize the element
+      val += static_cast<DataType>(1);
 
       // increment bound coord
       boundCoord[numIndicesSummation-1]++;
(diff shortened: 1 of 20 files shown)


Tensile/Source/Client.h
```diff
@@ -291,7 +291,7 @@ void specializeData(
       if (db & 0x2) {
         std::cout << "[" << serialIdx << "] = " << val << "\n";
       }
-      initialData[serialIdx] = val++;  // actually initialize the element
+      initialData[serialIdx] = val += static_cast<DataType>(1);  // actually initialize the element
```
zaliu: In the original code, since val++ is postfix, val is incremented *after* being assigned to initialData[serialIdx].
Thus your new code is not equivalent to the original code.

jichangjichang: I just updated the codes.
Please review it again, thanks.

sdquiring:
This change includes mixed line-endings for DataType.py and test_DataType.py .  Can we be more careful about this in the future?

jichangjichang:
> This change includes mixed line-endings for DataType.py and test_DataType.py .  Can we be more careful about this in the future?

I noticed that the DataType.py and test_DataType.py are dos format. After we use vim to modify some changes into these 2 files, there are ^M in line end.
But if we try "git am" the commit, the am process would fail.
Therefore, I remove the ^M from line end.

Should we always keep ^M line end for dos format file?  Thanks.

PR #593, opened 2019-07-15
power study switches and scripts

ramjana:


ramjana:
will PR for later release

PR #603, opened 2019-07-23
integration of LogicFile for tile aware selection

wbgilmartin:
This work integrates the logic file and tuning workflow into Tensile for creating the library for the tile aware selection.

This is enabled with TileAwareSelection: true defined in the ProblemType definition of the tuning yaml file. (defaults to false).

If enabled the tensile will perform the tuning to determine the ideal performance for the dedicated Summation sizes. Defaults to K = [32,64,96,128,256,512,1024,2048,4096,8192,16192]. This can be overridden with: 
BenchmarkFinalParameters:
    - SolutionSummationSizes: [ ... ]


Tensile/SolutionSelectionLibrary.py
```diff
@@ -0,0 +1,123 @@
...
+from . import __version__
+
+from copy import deepcopy
+from sys import stdout
+import array
+import csv
+import os
+import time
+
+try:
```
sdquiring: Is 'yaml' used directly in this file?

Tensile/SolutionSelectionLibrary.py
```diff
@@ -0,0 +1,123 @@
...
+
+    solutionMinNaming = Solution.getMinNaming(solutions)
+
+    
+    for solution in solutions:
+      solutionName = Solution.getNameMin(solution, solutionMinNaming)
+      solution["SolutionNameMin"] = solutionName
+      allSolutions.append(solution)
+      
+    dataFile = open(dataFileName, "r")
```
sdquiring: Open files should be in a `with` block.

Tensile/SolutionSelectionLibrary.py
```diff
@@ -0,0 +1,123 @@
...
+def analyzeSolutionSelection( problemType, problemSizeGroups):
+
+  dataFileNameList = []
+  performanceMap = {}
+  solutionData = {}
+  allSolutions = []
+
+  for problemSizeGroup in problemSizeGroups:
+    problemSizes = problemSizeGroup[0]
+    dataFileName = problemSizeGroup[3]
```
sdquiring: We should make new sections of the YAML maps instead of arrays to avoid these magic numbers.

PR #604, opened 2019-07-24
New client host code

sdquiring:
# Summary of changes
 - Client executable is functional and writes `.csv` files which are compatible with the rest of the Python scripts.
 - The client executable is build as 'step 0' (`0_Build  1_BenchmarkProblems  2_BenchmarkData  3_LibraryLogic`) when running the main Tensile script
 - Currently we run both the old client and the new client, and compare the .csv files (so it takes longer than before).
 - Implemented all data types and kernel features necessary to pass nightly tests.

## Reorganized files
 - `lib/client` is moved to `Tensile/Source/client`
 - `lib/tests` is moved to `HostLibraryTests`
 - The rest of `lib/` is moved to `Tensile/Source/lib`
 - CMake files for Tensile have been moved to `Tensile/cmake`

## User changes

 - Building the host unit tests requires
    - The fix for the CMake issue (see below)
    - Recursive checkout of Tensile
    - Packages: libboost-all-dev, llvm-6.0-dev

Your compiler should be set to `hcc` when invoking `cmake`:

    $ cmake -D CMAKE_CXX_COMPILER=hcc -D CMAKE_BUILD_TYPE=Debug ../HostLibraryTests

# Note the workaround (https://github.com/RadeonOpenCompute/hcc/pull/1211):
Until the fix is available in `hcc` available through `apt`, you will need to patch  `/opt/rocm/lib/cmake/hcc/hcc-config.cmake` on your machines in order to build Tensile (and rocBLAS once it's integrated). The dockerfile applies this automatically for CI, and there is a script to do this checked in:

    $ sudo ./docker/scripts/fix_hcc_config


Tensile/Code.py
```diff
@@ -20,7 +20,7 @@
 ################################################################################
 
 from __future__ import print_function
-from .Common import globalParameters, printExit, printWarning
+from .Common import globalParameters, printExit
```
sdquiring: The pyflakes step in Jenkins can't cause a build failure.  I ran pyflakes as a pre-checkin step; this file has lots of variables that are assigned but never used.

sdquiring:
Are there any other incoming PRs? This is getting ridiculous.  I'm trying to get my changes in as soon as possible.

PR #607, opened 2019-07-31
Zeropad

bensander:


bensander:
thx.  The tests have some debug enabled (PrintTensor) and also are not in nightly yet since the CPU golden model doesn't support zpad yet.  Will reduce verbosity when we move them to full automated checking.

PR #609, opened 2019-08-01
Check hip-clang as well.

mhbliao:
Need to check hip-clang as well to successfully compile with it.

### DIFF
--- a/Tensile/Source/tensile_bfloat16.h
+++ b/Tensile/Source/tensile_bfloat16.h
@@ -32,7 +32,7 @@
 
 // If this is a C compiler, C++ compiler below C++11, or a host-only compiler,
 // we only include a minimal definition of tensile_bfloat16
-#if __cplusplus < 201103L || !defined(__HCC__)
+#if __cplusplus < 201103L || (!defined(__HCC__) && !defined(__HIPCC__))
 
 #include <stdint.h>
 typedef struct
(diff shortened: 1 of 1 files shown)


Tensile/Source/tensile_bfloat16.h
```diff
@@ -32,15 +32,15 @@
 
 // If this is a C compiler, C++ compiler below C++11, or a host-only compiler,
 // we only include a minimal definition of tensile_bfloat16
-#if __cplusplus < 201103L || !defined(__HCC__)
+#if __cplusplus < 201103L || (!defined(__HCC__) && !defined(__HIPCC__))
```
yxsamliu: probably we should use `__HIP__` instead of `__HIPCC__`. `__HIP__` is defined by hip-clang. `__HIPCC__` is defined by hipcc.

mhbliao: Both macros are defined by clang for HIP lang, extracted from <clang>/lib/Frontend/InitPreprocessor.cpp

 457   if (LangOpts.HIP) {                                                                                                                                                                                                                                                      
 458     Builder.defineMacro("__HIP__");                                                                                                                                                                                                                                        
 459     Builder.defineMacro("__HIPCC__");                                                                                                                                                                                                                                      
 460     if (LangOpts.CUDAIsDevice)                                                                                                                                                                                                                                             
 461       Builder.defineMacro("__HIP_DEVICE_COMPILE__");                                                                                                                                                                                                                       
 462   }

mhbliao: More specifically, __HIPCC__ is defined by hip-clang natively and it also defined by `hipcc` for HCC only. If we follow HIP programming guide, we should only need to check __HIPCC__

yxsamliu: Sorry, I forgot `__HIPCC__` is also defined by hip-clang.

PR #621, opened 2019-08-13
Cmake fixes for tensile.

pfultz2:
This should make tensile work with `find_package(Tensile)`. It also fixes the issue with the rocm smi find module as it doesn't correctly set if its found or not found.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -29,7 +29,7 @@ import java.nio.file.Path;
 tensileCI:
 {
     def tensile = new rocProject('Tensile')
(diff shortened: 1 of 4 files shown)


pfultz2:
rocBLAS should just be update to use `find_package(Tensile)` instead of including TensileConfig.cmake(this was originally done because the TensileConfigVersion.cmake was broken). 

Also, rocBLAS doesn't need to call to get the path or set Tensile_ROOT.

However, building directly from the repo, you will need to set `Tensile_ROOT`.

sdquiring:
So can you fix the Jenkinsfile and create a PR for rocBLAS?  Creating a branch in the main repo is not the preferred way since it causes CI to run both for the branch and for the PR.

sdquiring:
rocBLAS is now pinned to a particular commit of Tensile so we can merge this in and then later fix rocBLAS.

PR #625, opened 2019-08-14
Setting cpu_count based on cpu affinity attached to the process

saadrahim:


saadrahim:
Emacs automatically removed line ending, do you want me to redo?

sdquiring:
No, that's fine.  We should wait and make sure this fixes the CPU count though.

saadrahim:
Reading logic files: Launching 8 threads...

As shown above, this fixes the issue for number of threads used to compile. The server has 64 CPU cores.

PR #635, opened 2019-08-18
Use the faster solution selection algorithm by default

ekuznetsov139:
At present, the variable ExpandRanges is set to True by default (and, to be honest, I don't see the point of allowing it to go to False).

When ExpandRanges is True, SolutionSelectionAlg 0 and 1 result in identical output, the only difference being that SolutionSelectionAlg 0 runs in cubic time in the number of solutions. I've seen the solution selection stage take 45+ minutes.

### DIFF
--- a/Tensile/Common.py
+++ b/Tensile/Common.py
@@ -68,7 +68,7 @@ globalParameters["ForceRedoBenchmarkProblems"] = True # if False and benchmarkin
 globalParameters["ForceRedoLibraryLogic"] = True      # if False and library logic already analyzed, then library logic will be skipped when tensile is re-run
 globalParameters["ForceRedoLibraryClient"] = True     # if False and library client already built, then building library client will be skipped when tensile is re-run
 globalParameters["ShowProgressBar"] = True     # if False and library client already built, then building library client will be skipped when tensile is re-run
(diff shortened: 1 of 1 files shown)


bensander:
@wbgilmartin  - we should see if the current selection algorithm (0) has merit?  It certainly takes significantly longer to run, especially for large numbers of solutions.  It might do a better job of removing solutions which add little value.  The "1" algorithm keeps all solutions which are fastest for any of the tuning problems.

ekuznetsov139:
Algorithm 0 is _intended_ to remove solutions that add little value.

In practice, when ExpandRanges is true, it does the exact same thing as algorithm 1 (remove all non-winners), because unifiedProblemSizes contains a bunch of empty lists (see https://github.com/ROCmSoftwarePlatform/Tensile/blob/develop/Tensile/LibraryLogic.py#L292-L314), therefore LogicAnalyzer.numProblemSizes is [0, 0, 0, 0, 0, 0, 0, 0] (https://github.com/ROCmSoftwarePlatform/Tensile/blob/develop/Tensile/LibraryLogic.py#L324-L333), therefore  LogicAnalyzer.problemIndicesForGlobalRange is empty, therefore the logic in https://github.com/ROCmSoftwarePlatform/Tensile/blob/develop/Tensile/LibraryLogic.py#L963-L995 that is supposed to remove "least important winners" never kicks in.

PR #636, opened 2019-08-19
Changing Createlibrary binary path

saadrahim:
The TensileCreateLibrary call in rocBLAS fails as the cmake file does not allow changing the Tensile_ROOT. Line 45 should be PARSE_Tensile_Root. This is cleaning this up for the rocBLAS fix. I have that ready and will create PR after this is merged.

### DIFF
--- a/Tensile/Source/TensileCreateLibrary.cmake
+++ b/Tensile/Source/TensileCreateLibrary.cmake
@@ -36,16 +36,7 @@ function(TensileCreateLibraryCmake
   message(STATUS "Tensile_CODE_OBJECT_VERSION from TensileCreateLibraryCmake : ${Tensile_CODE_OBJECT_VERSION}")
   message(STATUS "Tensile_COMPILER            from TensileCreateLibraryCmake : ${Tensile_COMPILER}")
 
-  # Tensile_ROOT can be specified instead of using the installed path.
-  set(oneValueArgs Tensile_BIN)
-  cmake_parse_arguments(PARSE "" "${oneValueArgs}" "" ${ARGN})
-
-  if(PARSE_Tensile_BIN)
-    # python not pre-installed, use scripts downloaded to extern/Tensile
-    set(Tensile_CREATE_COMMAND "${PARSE_Tensile_BIN}/TensileCreateLibrary")
-  else()
-    set(Tensile_CREATE_COMMAND tensileCreateLibrary)
-  endif()
+  set(Tensile_CREATE_COMMAND "${Tensile_ROOT}/bin/TensileCreateLibrary")
 
 
   set(Tensile_SOURCE_PATH "${PROJECT_BINARY_DIR}/Tensile")
--- a/Tensile/Source/TensileCreateLibrary.cmake
+++ b/Tensile/Source/TensileCreateLibrary.cmake
@@ -1,5 +1,5 @@
 ################################################################################
-# Copyright (C) 2016 Advanced Micro Devices, Inc. All rights reserved.
+# Copyright (C) 2016-2019 Advanced Micro Devices, Inc. All rights reserved.
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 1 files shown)


Tensile/Source/TensileCreateLibrary.cmake
```diff
@@ -37,12 +37,12 @@ function(TensileCreateLibraryCmake
   message(STATUS "Tensile_COMPILER            from TensileCreateLibraryCmake : ${Tensile_COMPILER}")
 
   # Tensile_ROOT can be specified instead of using the installed path.
```
amcamd: should the comment on line 39 have Tensile_BIN in place of Tensile_ROOT?

pfultz2:
This should not be necessary. Calling `find_package(Tensile)` will find the correct executable. This works and finds the correct executable:

```cmake
virtualenv_install("git+https://github.com/ROCmSoftwarePlatform/Tensile.git@${TENSILE_TAG}")
list(APPEND CMAKE_PREFIX_PATH ${VIRTUALENV_HOME_DIR})

find_package(Tensile REQUIRED)

TensileCreateLibraryFiles(
    "${CMAKE_CURRENT_SOURCE_DIR}/yaml"
    "${CMAKE_CURRENT_BINARY_DIR}/yaml"
    EMBED_LIBRARY tensile_kernels
    EMBED_KEY     tensile_kernels
    VAR_PREFIX TENSILE
    )
```

You should not be setting the `Tensile_ROOT` variable in rocBLAS, which only needs to be set when using tensile as a submodule.

Tensile/Source/TensileCreateLibrary.cmake
```diff
@@ -37,12 +37,12 @@ function(TensileCreateLibraryCmake
...
-  set(oneValueArgs Tensile_ROOT)
+  set(oneValueArgs Tensile_BIN)
   cmake_parse_arguments(PARSE "" "${oneValueArgs}" "" ${ARGN})
 
-  if(PARSE_Tensile_ROOT)
+  if(PARSE_Tensile_BIN)
     # python not pre-installed, use scripts downloaded to extern/Tensile
-    set(Tensile_CREATE_COMMAND "${Tensile_ROOT}/TensileCreateLibrary")
```
saadrahim: @pfultz2 This line is wrong, it should be PARSE_Tensile_ROOT? Otherwise it does not do what is intended.

saadrahim:
> This should not be necessary. Calling find_package(Tensile) will find the correct executable. This works and finds the correct executable:
> virtualenv_install("git+https://github.com/ROCmSoftwarePlatform/Tensile.git@${TENSILE_TAG}")
> list(APPEND CMAKE_PREFIX_PATH ${VIRTUALENV_HOME_DIR})
> 
> find_package(Tensile REQUIRED)
> 
> TensileCreateLibraryFiles(
>     "${CMAKE_CURRENT_SOURCE_DIR}/yaml"
>     "${CMAKE_CURRENT_BINARY_DIR}/yaml"
>     EMBED_LIBRARY tensile_kernels
>     EMBED_KEY     tensile_kernels
>     VAR_PREFIX TENSILE
>     )
> You should not be setting the Tensile_ROOT variable in rocBLAS, which only needs to be set when using tensile as a submodule.

I am not setting Tensile_ROOT. find_package and list(APPEND) is not adding the directory that tensileCreateLibrary is located in. As a result, it cannot call it.

Tensile/Source/TensileCreateLibrary.cmake
pfultz2: Yes, but it should be `set(Tensile_CREATE_COMMAND "${Tensile_ROOT}/bin/TensileCreateLibrary")`, because `Tensile_ROOT` gets set to the top-level installed directory automatically [here](https://github.com/ROCmSoftwarePlatform/Tensile/blob/develop/Tensile/cmake/TensileConfig.cmake#L29). So this flag can be removed completely.

saadrahim: I agree with this, fixing to get it working this way.

PR #640, opened 2019-08-21
Code Object V3 support for HIP-Clang

zaliu:
Supports Code Object V3 for HIP-Clang.  Passes full rocblas-test in HIP-Clang docker environment (build 174) with either CO V3 or CO V2.

### DIFF
--- a/Tensile/KernelWriterAssembly.py
+++ b/Tensile/KernelWriterAssembly.py
@@ -470,7 +470,7 @@ class KernelWriterAssembly(KernelWriter):
           '-target', 'amdgcn-amd-amdhsa']
 
     if archHasV3:
-      rv += ['-mno-code-object-v3']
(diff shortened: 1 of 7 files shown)


Tensile/KernelWriterAssembly.py
```diff
@@ -2148,7 +2195,8 @@ def functionSignature(self, kernel ):
       cptSize = "4"
       cptAlign = "4"
       cptByte  = 4
-      cptValueType = "F32"
+      if globalParameters["CodeObjectVersion"] == "V2": cptValueType = "F32"
+      if globalParameters["CodeObjectVersion"] == "V3": cptValueType = "f32"
```
sdquiring: Could be simplified with a toUpper() or toLower() at the end, although this section looks like it should be replaced with more fields in DataType.

zaliu: Agreed.  I modified V3 metadata following @amcamd's V2 metadata coding style to help ease the review.  We'll simplify it later.  Thanks.

PR #641, opened 2019-08-22
Check batch stride0

bensander:
Scott - note we will need these "AllStridesZero" style assertion checks in new client too.


sdquiring:
Can we merge this in? I'd like to get the rest of CI working again.

PR #642, opened 2019-08-22
Add TensileParseLog function to parse rocblas logs and generate yaml configs

pfultz2:
The command `TensileParseLog -i $LOG_FILE -p $BUILD_DIR -o benchmark.yaml -l $LIBRARY` can parse the rocblas log files and generate the tensile files. This is included in the normal install of tensile so an extra checkout of tensile is not necessary.


amdkila:
Temporarily closing PR until after ROCm3.0 release

PR #654, opened 2019-09-04
Tensile fixes that are needed in order for rocBLAS hip-clang develop branch to compile successfully

amdkila:
- Built rocBLAS with hip-clang compiler successfully with Tensile_COMPILER=hcc and COV=V2 and V3, and Tensile_COMPILER=hipcc and COV=V2 and V3
- 84 test failures when running rocblas-test, all rotm.blas1 functions
-> Lots of failed assertions, but none for the 84 failed tests
- The main error had to do with how default functions are declared
-> Will be removed later today
-> Tensile build error should be fixed when we change how default functions are declared


Tensile/Source/lib/source/Tensile.cpp
```diff
@@ -35,14 +35,14 @@
...
+
     TENSILE_API Problem::~Problem() = default;
     TENSILE_API ProblemInputs::~ProblemInputs() = default;
     TENSILE_API Hardware::Hardware() = default;
     TENSILE_API Hardware::~Hardware() = default;
     TENSILE_API Solution::~Solution() = default;
     TENSILE_API SolutionAdapter::~SolutionAdapter() = default;
-
-#ifdef TENSILE_DEFAULT_SERIALIZATION
+
```
sdquiring: Removed ifdef - shouldn't this leave a dangling #endif?

Tensile/Source/lib/source/ContractionProblem.cpp
```diff
@@ -867,7 +868,7 @@ namespace Tensile
     TypedContractionInputs<A, B, C, D, Alpha, Beta>::TypedContractionInputs() = default;
 
     template <typename A, typename B, typename C, typename D, typename Alpha, typename Beta>
-    TypedContractionInputs<A, B, C, D, Alpha, Beta>::~TypedContractionInputs() = default;
+    TypedContractionInputs<A, B, C, D, Alpha, Beta>::~TypedContractionInputs() = default;
```
sdquiring: whitespace

PR #657, opened 2019-09-06
Add missing mutex header

pfultz2:
This fixes a compile error when building the host library.


saadrahim:
@pfultz2 Please create your own fork of Tensile. You doubling the load on CI by running a job for the branch and the PR.

PR #667, opened 2019-09-17
Support double complex gemm assembly kernels

rosenrodt:


bragadeesh:
@imcarsonliao can you review and provide approval first, then @zaliu can review & approve

Tensile/KernelWriterAssembly.py
```diff
@@ -2210,6 +2254,9 @@ def functionSignature(self, kernel ):
       elif kernel["ProblemType"]["DataType"].isDouble() or \
            kernel["ProblemType"]["DataType"].isSingleComplex():
         kStr += self.v2Argument(                         "alpha", cptSize, cptAlign,      "ByValue", cptValueType); ka_size += cptByte
+      elif kernel["ProblemType"]["DataType"].isDoubleComplex():
+        kStr += self.v2Argument(                         "alpha_real", cptSize, cptAlign,      "ByValue", cptValueType); ka_size += 4
```
zaliu: Since alpha and beta are of type double complex, both real and imaginary parts are of type double, so ka_size should increment by 8, not 4, right?

jichangjichang: @zaliu 
For Single Complex, the kernel launch call treate alpha field just one argument and its datatype's size is 8 bytes.
It should be more reasonable to set it as one argument but not two.
Thanks.

imcarsonliao: modify to the following code, single complex and double complex can be consistent in metadata - one alpha and one beta. 

kStr += self.v2Argument(  "alpha", '16', '16',      "ByValue", "Struct"); ka_size += 16
kStr += self.v2Argument(  "beta", '16', '16',      "ByValue", "Struct"); ka_size += 16

rosenrodt: @zaliu Both issues have now been addressed in latest commits.

PR #670, opened 2019-09-18
Add stridea, strideb predicates

bensander:
- PBD=1 requires batchb==0


Tensile/Contractions.py
```diff
@@ -212,6 +212,9 @@ class ProblemPredicate(Properties.Predicate):
     @classmethod
     def FromOriginalKeyPair(cls, pair):
         (key, value) = pair
+        # TODO - change to use SetConstStrideB
+        if key == 'PackBatchDims':
```
sdquiring: I think you want to check `value` here as well - this adds the predicate if this property even exists.

PR #674, opened 2019-09-23
Link against hip::device since TensileHost includes hip_runtime.h

pfultz2:


bragadeesh:
@sdquiring @zaliu , this seems a simple PR, can you review and resolve

PR #675, opened 2019-09-23
Build Tensile PR in rocBLAS 

amdkila:
- This PR builds Tensile PRs in rocBLAS. This can be viewed here: http://10.216.151.18:8080/job/Dependency%20Projects/job/Tensile/view/change-requests/ 
- Passed CI


.jenkins/Jenkinsfile2
```diff
@@ -0,0 +1,92 @@
+#!/usr/bin/env groovy
```
sdquiring: This file should have a more descriptive name.  Do we want to move the other jenkinsfile into here as well?

amdkila: Yeah, definitely agree about the name. Not sure what to call it, definitely open to suggestions. As for moving the other jenkinsfile here, sure. I've added .jenkins/jenkins to RocmSoftwarePlatform as an acceptable Jenkins file location, so I can move the current Jenkins file.

sdquiring: The main thing would be to put 'rocblas' somewhere in the name so that it's obvious that this is building that project.

amdkila:
Compile stage failures appear to be legitimate errors. This should be safe to merge.

docker/dockerfile-build-ubuntu-rock
```diff
@@ -4,28 +4,34 @@
 ARG base_image
 
 FROM ${base_image}
-LABEL maintainer="saad.rahim@amd"
+LABEL maintainer="tensile.maintainer@amd.com"
```
saadrahim: check other projects for convention

PR #677, opened 2019-09-25
Client update

bensander:
Collection of small client updates.  Scott, your feedback appreciated.

### DIFF
--- a/Tensile/Source/lib/include/Tensile/TensorDescriptor.hpp
+++ b/Tensile/Source/lib/include/Tensile/TensorDescriptor.hpp
@@ -100,6 +100,7 @@ namespace Tensile
     class TENSILE_API TensorDescriptor
     {
     public:
+        static const size_t UseDefaultStride;
 
         TensorDescriptor();
 
@@ -128,7 +129,7 @@ namespace Tensile
                          std::initializer_list<size_t> sizes)
             : m_sizes(sizes),
               m_dataType(t)
-        
+
         {
             this->calculate();
         }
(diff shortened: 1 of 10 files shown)


Tensile/Source/lib/include/Tensile/TensorDescriptor.hpp
```diff
@@ -258,26 +258,35 @@ namespace Tensile
...
-
-        stream << "Tensor("
-            << desc.sizes()[0] << ", "
-            << desc.sizes()[1] << ", "
-            << desc.sizes()[2] << ")";
-
-       stream << std::endl;
-
-        for(index3[2] = 0; index3[2] < desc.sizes()[2]; index3[2]++)
+        const size_t maxDims=8;
```
sdquiring: Shouldn't be too difficult to extend this to arbitrary dimensions - see Reference.cpp in the client code, and CoordCount / CoordNumbered.

Tensile/Source/lib/source/ContractionProblem.cpp
```diff
@@ -494,10 +494,6 @@ namespace Tensile
 
     void ContractionProblem::normalize()
     {
-        std::sort(m_freeIndices.begin(),  m_freeIndices.end());
```
sdquiring: We aren't sorting any of the dimensions now?

Tensile/Source/lib/include/Tensile/TensorDescriptor.hpp
bensander: I added a comment to indicate we can make this better.  The PR is an improvement over previous support which was hard-corded to only 3 dims.

PR #680, opened 2019-10-01
Operation conv

bensander:
Should be Mostly Harmless to existing functionality, this is WIP to define an easier interface for mapping convolutions to tensor contraction syntax.

### DIFF
--- a/Tensile/Common.py
+++ b/Tensile/Common.py
@@ -67,6 +67,14 @@ globalParameters["ValidationPrintValids"] = False # print matches too
 globalParameters["ForceRedoBenchmarkProblems"] = True # if False and benchmarking already complete, then benchmarking will be skipped when tensile is re-run
 globalParameters["ForceRedoLibraryLogic"] = True      # if False and library logic already analyzed, then library logic will be skipped when tensile is re-run
 globalParameters["ForceRedoLibraryClient"] = True     # if False and library client already built, then building library client will be skipped when tensile is re-run
+
+# Compare CPU reference convolution model vs golden tensor contracton model
+# Useful to test if conversion from tensor contraction is working as expected
+# In this mode, the filter,stride,dilation are specified in the problem type.
+# If the problem type uses constant Filter,Stride,Dilation,Pad* (ie these are not 'N'), then the
+# specified constant MUST match the dimension in the problem or the tensile runtime will assert.
+# The batch size, spatial dims, Cin, and Cout are always read from the problem description.
+globalParameters["ConvolutionVsContraction"] = False
 globalParameters["ShowProgressBar"] = True     # if False and library client already built, then building library client will be skipped when tensile is re-run
 globalParameters["SolutionSelectionAlg"] = 1          # algorithm to detetermine which solutions to keep. 0=removeLeastImportantSolutions, 1=keepWinnerSolutions (faster)
 globalParameters["ExpandRanges"] = True          # expand ranges into exact configs before writing logic file.  False ignores ranges.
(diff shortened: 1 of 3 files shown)


Tensile/SolutionStructs.py
```diff
@@ -44,12 +46,277 @@ def pvar(state, field):
...
+    print ("- '(TBD)' indicates the parm is flexible and must be specified at runtime.")
+    print ("- '(i)' where i is an integer constant, indicates the parm is hard-coded at compile time.")
+    print ("  The runtime value must match the compile-time value.")
+    print ("- Unspecified strides use default stride value:")
+    print ("    stride[i] = (stride[i-1]*size[i]) for i>0 ; 1 for i==0")
+
+    print ()
+    print ("ProblemType Definition:")
+    for k in (
+        'OperationType','DestDataType','DataType','HighPrecisionAccumulate',\
```
sdquiring: This list could be separated out into a class property.

Tensile/SolutionStructs.py
```diff
@@ -44,12 +46,277 @@ def pvar(state, field):
...
+      elif self.tensorAFormat in ("CNHW", "CNDHW"):
+        self.registerA( [(cidx,Fbs.Sum,cdim), (nidx,Fbs.Batch,ndim)] + spatialDims + filterDims )
+      else:
+        raise RuntimeError ("unknown activation format '%s'"%self.tensorAFormat)
+
+      problemTypeOut["NumIndicesC"] = 2+len(spatialDims)
+
+      transposeCK =  convolutionType=="ConvolutionBackwardData"
+      ndim.strideB = 0
+      if self.tensorBFormat == "KCYX" and not transposeCK or\
```
sdquiring: Could this translation be separated out into a function and unit tested?

Tensile/SolutionStructs.py
```diff
@@ -44,12 +46,277 @@ def pvar(state, field):
...
+     print ("warning: need UseInitialStrides")
+     #problemTypeOut["UseInitialStrides"] = 1
+   #else:
+     #problemTypeOut["UseInitialStrides"] = 0
+
+
+    #self.printUsage(problemTypeOut)
+
+  # registerA and registerB:
+  # Provide a list of indices in convolution order - these will be reversed when assigned to IndexAssignmentsAB
```
sdquiring: docstrings are more useful than preceding comments - they can be queried and used to generate docs.

Also, what do these functions do? It sounds like they provide a list of indices, not that the caller is providing it.

Tensile/Common.py
```diff
@@ -799,6 +811,43 @@
     "NumIndicesC":              2,
     "UseInitialStrides":        False,
 
+    # For OperationType == Convolution*
+    # *HW and *YX   create solution with 2 spatial dimensions.
+    # *DHW and *ZYX create solution with 3 spatial dimensions.
+    "TensorAFormat":           "NCHW",  # NCHW, NHWC, CNHW
```
sdquiring: What about putting these into another data structure?  I can see it being confusing if we have both the conv- and gemm-specific options here.

Tensile/SolutionStructs.py
bensander: can you say more - the list of problem type properties could be a @property of Convolution class?

Tensile/SolutionStructs.py
bensander: good idea

Tensile/SolutionStructs.py
sdquiring: The goal is to make it more obvious why these specific properties are relevant here and to make it possible to use this list elsewhere.

I don't specifically mean using `@property` - I guess class member or class variable would be more accurate: just assign these to a variable at the class level, say `PropertyNames` or `PropertyKeys`.  Then you would access it here with` Convolution.PropertyNames`.

Tensile/SolutionStructs.py
bensander: added comments, try to explain what they do.

Tensile/Common.py
bensander: yes, see latest code.  Looks nice in the file and has cleaner separation.

Tensile/Tests/nightly/convolution_config/test_simple.py
```diff
@@ -0,0 +1,82 @@
+import pytest
```
sdquiring: Great!

Other unit tests are so far in the Tensile/UnitTests directory.  They run both with nightly and with pre-checkin since they run very quickly (Reported time is under 1s).

bensander: i'll move em over there soon

PR #684, opened 2019-10-03
Index mappings and packFreeDims

AlexBrownAMD:
Index mapping variables to determine original order of free/bound/batch indices. Add packFreeDims option.

### DIFF
--- a/Tensile/Source/lib/include/Tensile/ContractionProblem.hpp
+++ b/Tensile/Source/lib/include/Tensile/ContractionProblem.hpp
@@ -43,6 +43,8 @@ namespace Tensile
 
         ContractionProblem() = default;
 
+        using IndexList = std::vector<int>;
+
         /**
          * Represents a pair of free indices in a tensor contraction.
          */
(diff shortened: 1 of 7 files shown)


Tensile/Source/lib/include/Tensile/ContractionProblem.hpp
```diff
@@ -159,12 +161,19 @@ namespace Tensile
...
-        FreeIndices  const&  freeIndicesB() const { return m_freeIndicesB; }
-        FreeIndices  const&  freeIndices() const { return m_freeIndices; }
+        IndexList    const& freeIndicesA() const { return m_freeIndicesA; }
+        IndexList    const& freeIndicesB() const { return m_freeIndicesB; }
+        FreeIndices  const& freeIndices() const { return m_freeIndices; }
+        IndexList    const& batchIndicesA() const { return m_batchIndicesA; }
+        IndexList    const& batchIndicesB() const { return m_batchIndicesB; }
         BatchIndices const& batchIndices() const { return m_batchIndices; }
+        IndexList    const& boundIndicesA() const { return m_boundIndicesA; }
+        IndexList    const& boundIndicesB() const { return m_boundIndicesB; }
```
bensander: nice, full set of functionality

Tensile/Source/lib/source/ContractionProblem.cpp
```diff
@@ -485,12 +485,19 @@ namespace Tensile
         m_batchSizes.resize(m_batchIndices.size());
         m_boundSizes.resize(m_boundIndices.size());
 
+        m_freeIndicesA.resize(m_a.dimensions(), -1);
```
bensander: a.dimensions() includes all dims of a - free, batched,and bound, so this will 'oversize' a vector meant to contain just free indices for A.   Make sure we document this may contain -1/invalid values.

AlexBrownAMD: Yeah, the oversized vector (and -1's) lets us recreate the order IndexAssignmentsA in the c++ environment. Just added some comments to the header to help explain that freeIndicesA contains extra values of -1.

PR #685, opened 2019-10-03
Adding gfx908

saadrahim:


saadrahim:
Enabling gfx908 tests even though some are still failing

PR #689, opened 2019-10-07
Conv checker

bensander:
Add CPU conv golden model and comparison mode, and more tests.


Tensile/Source/lib/include/Tensile/ConvolutionProblem.hpp
```diff
@@ -0,0 +1,155 @@
...
+        public:
+            ActivationFormat();
+            void FromIdentifier(std::string identifier, size_t formatNumSpatialDims,
+                    size_t numSpatialDims, std::vector<size_t> *filters);
+            std::string description() const;
+
+            size_t batchPosition() const { return m_batchPosition; };
+            size_t channelPosition() const   { return m_channelPosition; };
+
+            // fi is filter index 0...MaxNumSpatialDims.
```
sdquiring: - When we have these explanatory comments, they should be Doxygen style comments so we can eventually do something with them.
 - What is 'fi' in this context? I don't see that anywhere around here.

Tensile/Source/lib/include/Tensile/ConvolutionProblem.hpp
```diff
@@ -0,0 +1,155 @@
...
+            TensorFormat format() const { return m_format;};
+        protected:
+            std::string  m_formatIdentifier;
+            TensorFormat m_format;
+        };
+
+        class TENSILE_API ActivationFormat : public BaseFormat {
+        public:
+            ActivationFormat();
+            void FromIdentifier(std::string identifier, size_t formatNumSpatialDims,
```
sdquiring: Should be a static function.

Tensile/Source/lib/include/Tensile/ConvolutionProblem.hpp
```diff
@@ -0,0 +1,155 @@
...
+
+            // 0,1,2 order is X,Y,Z
+            // size is number of spatial dims, no extra InvalidPos values
+            std::vector<size_t> m_spatialPositions;
+        };
+
+        class TENSILE_API WeightFormat : public BaseFormat {
+        public:
+            WeightFormat();
+            void FromIdentifier(std::string identifier, bool transposeCK, size_t numSpatialDims,
```
sdquiring: Should be a static function.

Tensile/Contractions.py
```diff
@@ -63,6 +63,11 @@ def FromOriginalState(cls, d):
         batchIndices = []
         boundIndices = []
 
+        try:
+            cls.convolution = d.convolution
```
sdquiring: `cls` here is the *class*, not any instance.  This is like

ProblemType.convolution = d.convolution.

The instance is `rv`.

Tensile/Source/client/source/ReferenceValidator.cpp
```diff
@@ -122,104 +145,119 @@ namespace Tensile
...
-                {
-                    auto const& typedReference = dynamic_cast<BFloat16ContractionInputs const&>(*m_referenceInputs);
-                    auto const& typedResult = dynamic_cast<BFloat16ContractionInputs const&>(*inputs);
-                    validateTyped(typedReference, typedResult);
-                }
-                else
-                {
-                    throw std::runtime_error("Data type not implemented.");
-                }
+              validateSolution(inputs);
```
sdquiring: Not using the return value here. This is why I don't really like return codes to report errors.

Tensile/Source/client/source/Reference.cpp
```diff
@@ -237,6 +237,259 @@ namespace Tensile
...
+                    if (convFilterSize != -1)
+                      assert(convFilterSize == problemFilterSize);
+                    fcount[fi] = problemFilterSize;
+                }
+            }
+
+            std::vector<size_t> activationDims;
+            switch (convProblem.tensorA().format()) {
+                case ConvolutionProblem::TensorFormat::NCHW:
+                    for (int fi=ConvolutionProblem::MaxNumSpatialDims-1; fi>=0; fi--)
```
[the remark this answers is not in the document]
bensander: agree, will review in follow-up checkin

Tensile/Source/lib/include/Tensile/ConvolutionProblem.hpp
```diff
@@ -0,0 +1,155 @@
...
+        ConvolutionProblem() {}
+
+        void FromIdentifier(std::string identifier);
+        void validate(const ContractionProblem &problem) const;
+
+        const std::vector<size_t> filter() const { return m_filters;};
+        const std::vector<size_t> stride() const { return m_strides;};
+        const std::vector<size_t> dilation() const { return m_dilations;};
+
+        const ActivationFormat &tensorA() const { return m_tensorA; };
```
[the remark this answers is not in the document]
bensander: rename to 'formatA' ?  Can update in follow-up checkin.

Tensile/Source/client/source/ReferenceValidator.cpp
bensander: tricky case here,  with new code we are calling validateSolution from two places and want to record the comparison status in separate member fields.  Return value seems reasonable as opposed to making the validateSolution aware of the caller so it can set appropriate member var.

Tensile/Source/client/source/DataInitialization.cpp
```diff
@@ -181,6 +181,7 @@ namespace Tensile
               m_dMaxElements(0),
               m_cEqualsD(args["c-equal-d"].as<bool>()),
               m_convolutionVsContraction(args["convolution-vs-contraction"].as<bool>())
```
sdquiring: Need a comma here.

PR #695, opened 2019-10-09
glob code objects under data/kernels_lite

zaliu:


zaliu:
gfx908 pre_checkin run was a mess.  Let's fix this one and go from there.  Thanks.

PR #711, opened 2019-10-25
Conv new

bensander:
- Bug fixes for convolution config type and convolution reference model
- Refactor convolution_config tests
  - Level selects between simple assertions to running tensile client.
- Add UseInitialStride for source
- Fix zero-pad for source + add tests
- Plus some refactoring from last checkin:
  - rename Tensor* to Format* in ConvolutionProblem
  - Move code from Reference.cpp to ConvolutionProblem


Tensile/SolutionStructs.py
```diff
@@ -216,37 +244,181 @@ def __init__(self, problemTypeOut, convolutionType, config):
...
 
+  def dimIdx(self, convolutionChar):
+    return self.convolutionDims[convolutionChar][0]
+
+  def convolutionChar(self, dimIdx):
+    return self.indexAssignments[dimIdx][1].shortChar
+
+  def makeProblem(self, keepTbd, n, c, k, spatialIn=None):
+    """
+    Generate valid problem dims
```
sdquiring: Seems like we should be able to further split this into testable functions.

Also, have you looked at how TensorFlow deals with placeholder values?  Could be useful.

Tensile/SolutionStructs.py
```diff
@@ -216,37 +244,181 @@ def __init__(self, problemTypeOut, convolutionType, config):
...
+          raise RuntimeError ("problemSize must specify spatial parms or set ConvolutionConfig.spatial")
+
+    if len(spatialIn) != self.formatNumSpatialDims:
+      raise RuntimeError ("len(spatialIn=", spatialIn, ") must match formatNumSpatialDims(%d)"%self.formatNumSpatialDims)
+
+    spatialTbd = not all(i!=-1 for i in spatialIn)
+    filterTbd = not all(i!=-1 for i in self.filter)
+    strideTbd = not all(i!=-1 for i in self.stride)
+    dilationTbd = not all(i!=-1 for i in self.dilation)
+    padTbd = not all(i!=-1 for i in self.padStart) or not all(i!=-1 for i in self.padEnd)
```
sdquiring: These could be implemented as properties (with `@property`) and tested.

bensander: can you say more or provide an example? For example, these are function locals rather than class members - not sure how @ property applies?

sdquiring: The 4 that I highlighted are really properties of the object - `filterTbd` is a statement on the value of `self.filter`.  I'm suggesting that these be hoisted outside of this function and made into properties of the class.

    @property
    def filterTbd(self):
        return -1 in self.filter

    # ...
        if keepTbd and self.filterTbd:

Also, when you force-push, it invalidates the links from these comments to the code and makes it difficult to see what's changed.

PR #712, opened 2019-10-28
Test configs for ZGEMM HIP generator 

rosenrodt:
This PR adds test configs for ZGEMM HIP generator because the issues we're seeing so far are either resolved or circumvented. 

| | Time (Original) | Time (This PR) |
|---|---|---|
| `pre_checkin/double_complex` | 6m09s | 11m33s
| `nightly/double_complex` | 10m28s | ~54m27s~ **32m13s**

run with Xeon Silver 4110 & Radeon Vega20


rosenrodt:
I removed the `AssertSummationElementMultiple` and  `AssertFree0ElementMultiple` parts from test config. The run time is now around 30 min

PR #713, opened 2019-10-28
Move ConvolutionProblem to Tensile client code

amdkila:


amdkila:
Looks like this breaks Tensile, but works when rocBLAS calls Tensile. I'll try to fix this without breaking rocBLAS.

amdkila:
OK, this finally works. It builds the client successfully and rocBLAS works as well.

PR #715, opened 2019-11-01
Small portability changes.

twhite-cray:
These changes are needed for Tensile to work on my cluster. Tensile requires CMake version 3, but the command `cmake` points to version 2 on many computers. Changing to `cmake3` targets the newer version. Tensile requests a specific version of LLVM, 7.0, which eliminates other working versions, such as the one on my cluster that comes with AOCC, `AOCC.LLVM.2.0.0.B191.2019_07_19`. Thus I eliminated the requirement for specific version 7.0.

### DIFF
--- a/Tensile/ClientExecutable.py
+++ b/Tensile/ClientExecutable.py
@@ -33,7 +33,7 @@ class CMakeEnvironment:
         self.options = options
 
     def generate(self):
-        args = ['cmake3']
+        args = ['cmake']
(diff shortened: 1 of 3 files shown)


twhite-cray:
The change to `cmake3` isn't portable after all. It breaks on Ubuntu. I'll find another workaround on my cluster.

I would still like a resolution on the requirement for LLVM 7.0 specifically, which appears to be artificially strict.

PR #723, opened 2019-11-11
Replace std::call_once() with local static variables

leekillough:
Initializing local `static` variables, which is guaranteed thread-safe in C++11 and later, is [faster](https://www.modernescpp.com/index.php/thread-safe-initialization-of-a-singleton) than explicitly calling `std::call_once()`.
```
    void DataTypeInfo::RegisterAllTypeInfoOnce()
    {
        static int call_once = (registerAllTypeInfo(), 0);
    }
```
This is simpler and faster than explicitly calling `std::call_once()`, since the compiler has special ways of optimizing local `static` initialization.

The first time `RegisterAllTypeInfoOnce()` is called, it will call `registerAllTypeInfo()` once in a thread-safe way. After that, there is almost no overhead in later calls to `RegisterAllTypeInfoOnce()`, while in the original code, it will call `std::call_once()` every time.


leekillough:
Fixed the build error which was caused by `registerAllTypeInfoOnce()` being defined outside of the original class. Just made it another `static` member function. Fixed a couple of other typos.

Added the HPA case `<Half, Half, Half, Half, float, float>` to the `TypedContractionInputs` explicit instantiations. It was missing, and caused a link error when I ported GEMM_EX to the new client.

Tensile/Source/lib/source/ContractionProblem.cpp
```diff
@@ -913,9 +913,9 @@ namespace Tensile
 
 #ifdef TENSILE_USE_HALF
     template struct TypedContractionInputs<Half>;
+    template struct TypedContractionInputs<Half, Half, Half, Half, float, float>;
```
sdquiring: Using this is going to fail.

HPA takes a half for alpha/beta for now.  See here:
https://github.com/ROCmSoftwarePlatform/Tensile/blob/653bda595bcfc043b59f97005a797a1bd638a748/Tensile/KernelWriterAssembly.py#L8041

amdkila: This could break hip-clang as well, given the mismatch with input arguments for ContractionProblem.cpp and ContractionProblem.hpp. I guess hip-clang uses an older Tensile commit for the time-being, but after branching, this may become an issue.

leekillough: The above change does not break anything. It simple adds another `ContractionInput`.

I'm asking that it be reinstated, since the tests pass.

leekillough: Regardless of kernels, we need structures to set up data structures with certain type combinations. This adds one which gets rid of a link error. It does not change the behavior of the library on other functions (other than the `call_once` changes, which are performance)

leekillough: The failures were in rocBLAS, not in Tensile

PR #727, opened 2019-11-13
add AMAS-free replacement kernels

zaliu:
Tensile kernel names recently changed so that AMAS is no longer part of a kernel name.  All 7 active replacement kernels must be updated to conform to the new kernel naming convention.


amdkila:
It's safe to merge- 908 test failure is due to a seg fault when running tests.

PR #728, opened 2019-11-13
Replace call once with local static

leekillough:
The addition of a new template data type instantiation should not affect the behavior of libraries which do not call it.

It will not cause a compilation error.

It simply instantiates a necessary data structure, without regard to kernels.

rocBLAS uses data structures independent of the kernels which get called.

The replacement of `call_once` with `static` initialization does not change the behavior. It's just faster.

This PR needs to be recommitted. It is unrelated to any failures.

### DIFF
--- a/Tensile/Source/lib/include/Tensile/DataTypes.hpp
+++ b/Tensile/Source/lib/include/Tensile/DataTypes.hpp
@@ -78,6 +78,7 @@ namespace Tensile
 
     private:
         static void registerAllTypeInfo();
+        static void registerAllTypeInfoOnce();
 
         template <typename T>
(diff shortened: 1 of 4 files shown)


leekillough:
I have verified that Tensile and rocBLAS build and pass with this change.

leekillough:
Also, HPA should use 32-bit alpha and beta. That's a kernel implementation issue right now, but rocBLAS and Tensile APIs should use 32-bit alpha and beta, which rocBLAS does. They can be converted down to Half if necessary at the final call site, but only at the points where the conversion is needed -- we should not change the entire API call stack to use half-precision alpha, beta for HPA.

PR #738, opened 2019-11-26
Fix Pytest in CI

sdquiring:
Still not sure why it was broken or why this fixes it but it seems to work.


sdquiring:
Superseded by #740.

PR #739, opened 2019-11-26
tox call unittests separte from nightly/pre_checkin

bensander:
See if avoids tax confusion with directories


bensander:
Redundant

PR #742, opened 2019-12-02
Allow for architecture-specific builds 

amdkila:
- This allows users to pass in the architecture they want to use when building rocBLAS
---> e.g. If on vega10, use gfx900 flag (-a gfx900) to only generate Vega 10 kernels
---> See this rocBLAS PR: https://github.com/ROCmSoftwarePlatform/rocBLAS/pull/847
- Also adds a couple of missing cmake variables to new Tensile client
- This works with both old and new Tensile clients
- This works in combination with the Tensile_LOGIC (-l flag)


b-sumner:
Isn't it confusing for developers if tensile uses different target names than the compiler uses?

amdkila:
> Isn't it confusing for developers if tensile uses different target names than the compiler uses?

I think that's a fair point. I can change those targets without much issue to gfx***, or maybe rename the Tensile_ARCHITECTURE cmake variable to something else to show that this just filters the Tensile_LOGIC directory (i.e. asm_full). The reason why I chose vega10/vega20/Arcturus/fiji/etc. is to (more or less) match the naming scheme in rocBLAS/src/library/blas3/Tensile/Logic/asm_full

b-sumner:
I remember now that tensile cares about more than just the ISA target, and in that case it might make sense to use the official board name or something like that.

amdkila:
@b-sumner I decided to match the compiler and use the ISA target.

Tensile/Source/CMakeLists.txt
```diff
@@ -178,9 +178,9 @@ else()
       include(${CMAKE_SOURCE_DIR}/TensileCreateLibrary.cmake)
       TensileCreateLibraryCmake(
         ${Tensile_LOGIC_PATH}           # path
-        ${Tensile_RUNTIME_LANGUAGE}     # OCL or HIP
         ${Tensile_COMPILER}             # hcc or hipcc
         ${Tensile_CODE_OBJECT_VERSION}  # V2 or V3
+        ${Tensile_ARCHITECTURE}         # all, gfx000, gfx803, gfx900, gfx906
```
yoichiyoshida: no gfx908?

Tensile/TensileCreateLibrary.py
```diff
@@ -956,6 +956,7 @@ def TensileCreateLibrary():
   argParser.add_argument("RuntimeLanguage", help="Which runtime language?", choices=["OCL", "HIP", "HSA"])
   argParser.add_argument("--cxx-compiler",           dest="CxxCompiler",       choices=["hcc", "hipcc"],       action="store", default="hcc")
   argParser.add_argument("--code-object-version",    dest="CodeObjectVersion", choices=["V2", "V3"], action="store", default="V2")
+  argParser.add_argument("--architecture",	     dest="Architecture",      choices=["all", "gfx000", "gfx803", "gfx900", "gfx906", "gfx908"], action="store", default="all")
```
yoichiyoshida: keep formatting consistent

Tensile/TensileCreateLibrary.py
```diff
@@ -987,13 +989,24 @@ def TensileCreateLibrary():
...
   print1("# CodeObjectVersion from TensileCreateLibrary: %s" % arguments["CodeObjectVersion"])
   print1("# CxxCompiler       from TensileCreateLibrary: %s" % CxxCompiler)
+  print1("# Architecture      from TensileCreateLibrary: %s" % arguments["Architecture"])  
 
   if not os.path.exists(logicPath):
     printExit("LogicPath %s doesn't exist" % logicPath)
 
+  # Translate GPU targets to filter filenames in Tensile_LOGIC directory
+  mapArchitecture = {'all':'_','gfx000':'none', 'gfx803':'r9nano',
+        'gfx900':'vega10', 'gfx906':'vega20', 'gfx908':'arcturus'}
```
yoichiyoshida: Don't define these in multiple places. If you need them in more than one spot, put it in common and import it into the files you need like all the other definitions.

Tensile/TensileCreateLibrary.py
```diff
@@ -956,7 +956,7 @@ def TensileCreateLibrary():
   argParser.add_argument("RuntimeLanguage", help="Which runtime language?", choices=["OCL", "HIP", "HSA"])
   argParser.add_argument("--cxx-compiler",           dest="CxxCompiler",       choices=["hcc", "hipcc"],       action="store", default="hcc")
   argParser.add_argument("--code-object-version",    dest="CodeObjectVersion", choices=["V2", "V3"], action="store", default="V2")
-  argParser.add_argument("--architecture",	     dest="Architecture",      choices=["all", "gfx000", "gfx803", "gfx900", "gfx906", "gfx908"], action="store", default="all")
+  argParser.add_argument("--architecture",	         dest="Architecture",      choices=["all", "gfx000", "gfx803", "gfx900", "gfx906", "gfx908"], action="store", default="all")
```
yoichiyoshida: Still doesn't fit formatting, maybe remove that tab

Tensile/TensileCreateLibrary.py
amdkila: It seems like I should just remove the map as the part of code that collects the config files is only done in TensileCreateLibrary.py- this isn't called in ClientWriter.py. I'll re-add the line that prints the Tensile_ARCHITECTURE value, but leave out the map.

amdkila:
Alright, this has finally passed CI. Should be safe to merge.

amdkila:
We need this merged in to fix the develop branch of rocBLAS

PR #746, opened 2019-12-06
Split Jenkinsfile into PreCheckin, Extended, and Debug jobs

eidenyoshida:


saadrahim:
I think this is ready to review.

.jenkins/Common.groovy
```diff
@@ -0,0 +1,76 @@
...
+        platform.runCommand(this, command)
+
+        publishHTML([allowMissing: false,
+                    alwaysLinkToLastBuild: false,
+                    keepAll: false,
+                    reportDir: "${project.paths.project_build_prefix}/docs/html",
+                    reportFiles: 'index.html',
+                    reportName: 'Documentation',
+                    reportTitles: 'Documentation'])
+
```
saadrahim: check whitespace

.jenkins/Common.groovy
```diff
@@ -0,0 +1,76 @@
...
+        
+    }
+    finally
+    {
```
saadrahim: remote as not needed

saadrahim:
> 
> 
> Now that Jenkinsfile is gone, if we want to disable CI, should we rename .jenkins/ instead?

Yes, that is one option or you can rename the individual files to disable the corresponding CI job. Also, jenkins is tracking jobs based on branches in a more manual manner. Only master, staging, develop, and current master-rocm-X.Y branches are built.

PR #749, opened 2019-12-16
add solution lookup caching to Tensile

wbgilmartin:
I verified that the performance matches the old client performance using this implementation.

Verified using rocblas.


Tensile/Source/lib/CMakeLists.txt
```diff
@@ -21,6 +21,8 @@
 
 cmake_minimum_required(VERSION 3.5)
 
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14" )
```
leekillough: Shouldn't this be `set( CMAKE_CXX_STANDARD 14 )` rather than passing explicit flags?

Tensile/Source/lib/include/Tensile/MatchingLibrary.hpp
```diff
@@ -71,9 +78,22 @@ namespace Tensile
 
             auto closestEntry = table->findBestMatch(problem, transform);
```
leekillough: Thread-safety?

Here's rocBLAS's implementation of [double-checked locking](https://en.wikipedia.org/wiki/Double-checked_locking):

https://github.com/ROCmSoftwarePlatform/rocBLAS/blob/98ac69fcc8f406ff35b96d63af9072862c564e17/library/src/include/logging.h#L276

PR #759, opened 2019-12-24
Modify tmp vgpr usage to fix fp16 test

AlexBrownAMD:


AlexBrownAMD:
> we may not need all those tmp registers for store offset generation. need to clean-up offset ptr calculation

Yea I agree, I have to revisit the changes clean up a few parts soon.

PR #744, opened 2019-12-04
gfx1010

sdquiring:
Adding support for gfx1010, for assembly kernels only for now.

PR #769, opened 2020-01-09
Mfma bf16 kgen pr1

jichangjichang:
- bfloat16 mfma kernel generator: now support 32x32x2x2bf16 mfma
- even size of DLRM sizes all PASS


Tensile/KernelWriterAssembly.py
```diff
@@ -6555,36 +6652,88 @@ def localReadDo(self, kernel, bufferIdx, iui, epsi, tP):
...
+          comment = "L -> Reg lro=%d swapByteOffset=%u ti=%u vIdx=%u rIdx=%u oIdx=%u buffer=%u iui=%u"\
+              %(tP["localReadOffset"],tP["localReadSwapByteOffset"],kernel["SubGroup%u"%tP["tensorIdx"]], vIdx, rIdx, oIdx, bufferIdx, iui)
+          localReadCode.addCode(Code.LocalReadInst(instruction.toCodeInst(paramTuple), comment))
+          valuIdx += blockWidth
+
+          # TODO - handle vector-load
+          if self.db["CheckValue1%s"%tc]:
+              localReadCode.addInst("s_waitcnt lgkmcnt(0)", "CheckValue1 wait for LDS read")
+              if kernel["ProblemType"]["DataType"].isHalf() or kernel["ProblemType"]["DataType"].isBFloat16():
+                localReadCode.append(self.assert_eq(destVgpr, hex(0x3c003c00))) # packed 1s
```
sdquiring: This doesn't look right for bf16.  Should be 0x3f803f80.

Tensile/KernelWriterAssembly.py
```diff
@@ -1900,11 +1912,55 @@ def defineMACs(self, kernel, m, innerUnroll):
 
     # bfloat16
     elif kernel["ProblemType"]["DataType"].isBFloat16():
-      if self.version == (9,0,8):
-        if kernel["ProblemType"]["HighPrecisionAccumulate"]:
-          kStr += ""
-        else:
-          kStr += ""
+      if self.version == (9,0,8) and kernel["ProblemType"]["HighPrecisionAccumulate"]:
```
sdquiring: Should be an architecture capability flag such as HasEccHalf.  See #744.

TonyYHsieh: Hi @sdquiring ,

I follow previous method to check should we use bfloat16 generator.
May you help to suggest how to modify??
What should we do for other code segment with self.version == (x, x, x) ???

jichangjichang:
rocblas-test pass all on gfx906 and gfx908:
 rocblas-test --gtest_filter=-*known_bug*

TonyYHsieh:
> LGTM, but you still need to resolve the comments Scott offered.

Hi Alex,
we only support BFfloat16 generate on GFX908.
We didn't test it on previous devices and we didn't have plan to support BFloat16 on previous devices yet.  For distinguish devices, we follow other code in Tensile to use version to check if we can run BFloat16. We think if we want to change to arch, we should change other part together. Can we follow up this change in other topic??

PR #771, opened 2020-01-13
More initialization patterns

AlexBrownAMD:
Added SerialIdx, SerialDim0, and SerialDim1 init patterns. These are problem size dependent and force reinitialization when running multiplie problems. They cannot be used with the prisitne on GPU option. Also added anewline in tensor printouts back in.


Tensile/Source/client/include/DataInitialization.hpp
```diff
@@ -124,6 +157,52 @@ namespace Tensile
...
+            {
+                auto const& sizes = tensor.sizes();
+                auto count = CoordCount(sizes.begin(), sizes.end());
+                std::vector<size_t> coord(tensor.dimensions(), 0);
+                for(size_t idx = 0; idx < count; idx++)
+                {
+                    CoordNumbered(idx, coord.begin(), coord.end(), sizes.begin(), sizes.end());
+                    array[tensor.index(coord)] = static_cast<T>(coord[1]);
+                }
+            }
```
sdquiring: These could be combined and take which coordinate as an argument.

PR #786, opened 2020-01-25
Use HIP atomicAdd as a workaround if compiler is hip-clang

amdkila:
Tentative workaround

### DIFF
--- a/Tensile/KernelWriterSource.py
+++ b/Tensile/KernelWriterSource.py
@@ -458,45 +458,46 @@ class KernelWriterSource(KernelWriter):
         kStr += "  } while (assumed != old);%s" % (self.endLine)
         kStr += "}%s" % (self.endLine)
         """
-        kStr += self.endLine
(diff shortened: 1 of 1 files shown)


iotamudelta:
Sorry to comment here (it's just atomicAdd...). Is there a reason not to use the newer branch also for the hcc path (i.e., only have one path)? In particular, I believe it'd be advantageous to use the library provided atomicAdd over rolling your own in general. Just a recommendation.

amdkila:
> Sorry to comment here (it's just atomicAdd...). Is there a reason not to use the newer branch also for the hcc path (i.e., only have one path)? In particular, I believe it'd be advantageous to use the library provided atomicAdd over rolling your own in general. Just a recommendation.

I agree with you, there's a performance gain to be had by avoiding the do-while loop. The option to use STL atomics should also be there, as users may try and use them for their own custom atomic functions.

saadrahim:
This is starting to delay hipclang testing. I shouldn't approve it as I am not aware of the consequences but someone should look at it. Also, looks like a lot of CI failures to review.

PR #792, opened 2020-01-31
 Use LLVM shared libraries if LLVMObjectYAML static library is missing

eidenyoshida:
SLES LLVM package does not contain static libraries.

See below for successful build of rocBLAS on SLES. 
http://10.216.151.18:8080/blue/organizations/jenkins/PreCheckin%2FrocBLAS/detail/PR-970/2/pipeline


sdquiring:
Can we detect that the LLVMObjectYAML library doesn't exist instead of being specific to detecting SLES?

eidenyoshida:
I've changed it to try finding the static library first and use the shared library if it doesn't detect it. 
rocBLAS is also able to build with SLES as well, please see below. 
http://10.216.151.18:8080/blue/organizations/jenkins/PreCheckin%2FrocBLAS/detail/PR-970/7/pipeline/187

PR #796, opened 2020-02-04
Add call count and other parameters to tuning scripts

amdkila:
- Adds call_count parameter and displays in yaml file
- Enables new client by default and changes a few parameters
- Displays kernel time before and after tuning
- Need to pass in log file with call_count (or w/o, it assumes call_count=1 then) into the analysis script
- Calculates weighted average and overall gain for each problem type (NN/NT/TN and strided)
- Corrects efficiency calculations (with and w/o mfma, map with CUs)


tuning/scripts/provision_repo.sh
```diff
@@ -9,8 +9,8 @@
 WORKING_PATH='release'
 HELP=false
 ROCBLAS=false
-PROVISION_BRANCH='develop'
-TENSILE_HOST='https://github.com/ROCmSoftwarePlatform/Tensile.git'
+PROVISION_BRANCH='call_count'
+TENSILE_HOST='https://github.com/amdkila/Tensile.git'
```
wbgilmartin: I think this should default to ROCmSoftwarePlatform and take in a script parameter to reference a different repo.

amdkila: Whoops, that should've been replaced. Also I agree- these should have command line parameters so that you can try a different fork and branch. I will add those in my next commit

PR #806, opened 2020-02-05
Mfma scaling fix

ramjana:
-store code register index fix for multiple row/col blocks 
- fix for localReadDo() supporting mulitple Column


AlexBrownAMD:
Tested this against the k2 branch and looks like that + bf16 work with the change

Tensile/Utilities/tensile_power.sh
```diff
@@ -0,0 +1,184 @@
...
+##genrate report calling
+
+echo "##########################################################"
+echo "Generating consolidated output pmlog.csv ....."
+echo "##########################################################"
+if [ "${generate_report}" = 1 ]; then
+   cmd=`ruby ./Tensile/Utilities/gen_csv.rb  ${output_dir} ${best_allsizes}`
+fi
+
+exit 1
```
bensander: what is this file for?

ramjana: script for generating power reports.. not sure how did that got into this PR.. I will remove the file..

PR #812, opened 2020-02-07
Adding hip-clang support in CI

amdkila:
- Also refactoring CI to use Stage 1 process


saadrahim:
We should send an email to inform the team of the CI changes.

PR #817, opened 2020-02-11
Zero pad

bensander:


sdquiring:
I really don't have a good handle on the code generator.  Can someone who has more experience there have a look at that part?

PR #822, opened 2020-02-14
Fix the flag checking for embeded library in cmake

pfultz2:
This makes it possible to avoid creating an embed library.


amdkila:
Thanks Paul for fixing this- I had made this mistake in the first place when I added some new CMake variables that already existed with the old Tensile client.. However, I think this should go into the develop branch rather than going in the master branch directly. Is it ok if I cherry-pick your commit and merge it into the develop branch?

pfultz2:
> Is it ok if I cherry-pick your commit and merge it into the develop branch?

That is fine. I just branched off maste(where we were pulling) to make the change as I didn't want to bring in too many changes with the fix in MIOpenTensile.

PR #835, opened 2020-02-26
Shorten file names

sdquiring:
Adding a mechanism which will allow kernel file names to be shortened by hashing the rest of the kernel name beyond a certain point.

Note that the kernel name does not change.

Also creating a new replacement kernel mechanism which looks inside the files for the kernel name so that replacement kernels will still work.

I was able to build rocBLAS' kernels with `--no-merge-files` with these changes.

### DIFF
--- a/Tensile/Common.py
+++ b/Tensile/Common.py
@@ -156,6 +156,7 @@ globalParameters["MaxLDS"] = 65536                # max LDS a kernel should atte
 globalParameters["MaxDepthU"] = 256               # max DepthU value to allow
 globalParameters["ShortNames"] = False            # on windows kernel names can get too long; =True will convert solution/kernel names to serial ids
 globalParameters["MergeFiles"] = True             # F=store every solution and kernel in separate file; T=store all solutions in single file
+globalParameters["MaxFileName"] = 128 # If a file name would be longer than this, shorten it with a hash.
 globalParameters["SupportedISA"] = [(8,0,3), (9,0,0), (9,0,6), (9,0,8), (10,1,0)]             # assembly kernels writer supports these architectures
 globalParameters["ClientBuildPath"] = "0_Build"                   # subdirectory for host code build directory.
 globalParameters["NewClient"] = 2                                 # 1=Run old+new client, 2=run new client only (All In)
--- a/Tensile/KernelWriter.py
+++ b/Tensile/KernelWriter.py
@@ -22,6 +22,7 @@
 from . import Code
 from . import Common
 from .Common import globalParameters, print2, CHeader, roundUp
+from .ReplacementKernels import ReplacementKernels
 from .SolutionStructs import Solution
 
 import abc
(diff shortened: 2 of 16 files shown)


pfultz2:
I ran this in miopentensile and I still get filename too long:

```
objcopy:/tmp/tmp.PLcRI3TlUY/Cijk_Ailk_Bjlk_DB_MT64x128x4_SN_APM1_AF0EM1_AF1EM1_AMAS3_ASEM1_BL1_DTL0_EPS1_FL1_GRVW2_GSU1_ISA908_IU1_K1_KLA_LPA0_LPB0_LDL1_NLCA1_NLCB1_ONLL1_PBD0_PK0_PGR1_PLR1_RK0_SU0_SNLL1_TT4_8_USFGRO0_VAW1_VW2_WG16_16_1_WGM1.eaab1092034a5b97b14ecc7b26046912.kernel.bc: File name too long
```

It does show the hash but the name is still too long.

Tensile/Common.py
```diff
@@ -156,6 +156,7 @@
 globalParameters["MaxDepthU"] = 256               # max DepthU value to allow
 globalParameters["ShortNames"] = False            # on windows kernel names can get too long; =True will convert solution/kernel names to serial ids
 globalParameters["MergeFiles"] = True             # F=store every solution and kernel in separate file; T=store all solutions in single file
+globalParameters["MaxFileName"] = 128 # If a file name would be longer than this, shorten it with a hash.
```
pfultz2: How about setting this to `64`? It seems like `128` will still make the name too long.

zaliu: rocBLAS build with this PR has a maximum file name length of 143.  In a docker container the limit seems to be around 224.  If any utility adds filename overhead as large as 80, that utility needs work.  Back to your issue.  I'm sure you can set your own MaxFileName to 64.  The default value of 128 appears to be quite sensible.

zaliu: rocBLAS build with this PR has a maximum file name length of 143.  In a docker container the limit seems to be around 224.  If any utility adds filename overhead as large as 80, that utility needs work.  Back to your issue.  I'm sure you can set your own MaxFileName to 64.  The default value of 128 appears to be quite sensible.

zaliu: A rocBLAS build with this PR has a maximum file name length of 143.  In a docker container the limit seems to be around 224.  If any utility adds filename overhead as large as 80, that utility needs work.  Back to your issue.  I'm sure you can set your own MaxFileName to 64.  The default value of 128 appears to be quite sensible.

zaliu: A rocBLAS build with this PR has a maximum file name length of 143.  In a docker container the limit seems to be around 224.  If any utility adds filename overhead as large as 80, that utility needs work.  Back to your issue.  I'm sure you can set your own MaxFileName to 64.  The default value of 128 appears to be quite sensible.

pfultz2: I am seeing a max file length of 277, which is too large. How do I set my own `MaxFileName` without forking Tensile?

pfultz2:
Ok, so after getting the forked repo, I am no longer seeing too long of filenames.

PR #837, opened 2020-02-27
Set default iteration value to 10 for rocblas-bench calls

amdkila:
Due to the unreliability of using 1 iteration in rocblas-bench calls, this sets the default value to 10 and makes a change to not affect the weighted average calculation.

### DIFF
--- a/tuning/automation/GenerateTuningConfigurations.py
+++ b/tuning/automation/GenerateTuningConfigurations.py
@@ -376,10 +376,7 @@ def OutputScript(problemMapper, scriptPath, namePart):
             lines.append(rocblas_call)
         with open(outputFileName, 'w') as f:
             for line in lines:
-                if "rocblas-bench" in line:
-                    f.write("%s0\n" % line)
-                else:
-                    f.write("%s\n" % line)
+                f.write("%s\n" % line)
 
     generateRunScript(scriptFileNames, scriptPath)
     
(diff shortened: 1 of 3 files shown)


tuning/automation/GenerateTuningConfigurations.py
```diff
@@ -397,7 +400,7 @@ def OutputScript2(problemMapper, scriptPath, namePart):
         with open(outputFileName, 'w') as f:
             for line in lines:
                 if "rocblas-bench" in line:
-                    f.write("ROCBLAS_TENSILE_LIBPATH=${TENSILE_LIBRARY} %s\n" % line)
+                    f.write("ROCBLAS_TENSILE_LIBPATH=${TENSILE_LIBRARY} %s0" % line)
```
wbgilmartin: If you look at ExtractSizes.py in the function: 
def GetRocBLASParser(): 

at the line :
    lineParser.add_argument("-i",dest="i", type=int,default=1)

Try setting the default to 10

amdkila: Thanks for the suggestion Bill. I've made that change in my latest commit.

tuning/automation/PerformanceAnalysis.py
```diff
@@ -102,6 +102,8 @@ def fillCallCounts(problemMapper, callCounts, callCountNN, callCountNNstrided, c
             midList = list()
```
wbgilmartin: problemMapper is a list of dictionaries so i don't think you need to iterate over it here:

see if this works

   for problem in problemMapper:
        midList = list()
        midList.append(problem["f"])
        midList.append(problem["transposeA"])
        midList.append(problem["transposeB"])
        theCount = problem["i"]
        if theCount == 10:
            theCount = 1
        callCounts.append(midList)

wbgilmartin: My offsets did not translate in the last comment

amdkila: Hmmm, I get the following error:

  File "../Tensile/tuning/automation/PerformanceAnalysis.py", line 102, in fillCallCounts
    midList.append(problem["f"])
TypeError: list indices must be integers or slices, not str

wbgilmartin: OK. This means I was wrong. The problem mapper is then a list.

PR #838, opened 2020-02-27
Optimizations to solve() function and more performance tests.

sdquiring:
Runtime optimizations:

 - `solution->solve()` optimized by moving debug options to default off and reserving space to avoid reallocations.
 - PropertyMatching can now use a binary search and early loop exit based on the first dimension of the key.
 - PropertyMatching no longer uses the batch size as a dimension of its distance metric.
 - CachingLibrary can now keep track of statistics regarding hits & misses.
 - Raw performance of the PropertyMatching lookup has been increased by moving virtual calls to templated calls & reducing datatype conversions.

Developer focused optimizations:

 - RunGEMMKernelTest now caches CPU reference results which significantly speeds up running the HostLibraryTests.
 - Moved the ParallelMap Python function to a separate Parallel.py file.  It's still accessible via Common.ParallelMap.


HostLibraryTests/llvm/LibraryPerformance_test.cpp
```diff
@@ -35,58 +34,156 @@
...
+            return iter->second;
+
+        return libraryCache[path] = LoadLibraryFile<ContractionProblem>(libraryPath());
+    }
 };
 
+std::map<std::string, std::shared_ptr<SolutionLibrary<ContractionProblem>>> LibraryPerformanceTest::libraryCache;
+
+TEST_P(LibraryPerformanceTest, PopulateCache)
+{
```
AlexBrownAMD: Is this test supposed to be empty?

Tensile/Source/lib/include/Tensile/PropertyMatching.hpp
```diff
@@ -150,10 +140,207 @@ namespace Tensile
...
+            }
+
             ReturnValue findBestKeyMatch(Key const& key, Transform transform) const
             {
                 const bool debug = Debug::Instance().printPropertyEvaluation();
+                const bool naive = Debug::Instance().naivePropertySearch();
+
+                if(naive)
+                {
+                    if(debug)
```
AlexBrownAMD: Is debug set at runtime?

wbgilmartin:
The rocblas problem that fails is:

- { rocblas_function: "rocblas_sgemm", transA: 'N', transB: 'N', M: 16328, N: 384, K: 384, lda: 16328, ldb: 125000, ldc: 125000 }

sdquiring:
Do you mean an actual segfault or a GPU Memory access fault?

On Mar 4, 2020, at 12:15 PM, wbgilmartin <notifications@github.com<mailto:notifications@github.com>> wrote:

[CAUTION: External Email]

The rocblas problem that fails is:

  *   { rocblas_function: "rocblas_sgemm", transA: 'N', transB: 'N', M: 16328, N: 384, K: 384, lda: 16328, ldb: 125000, ldc: 125000 }

wbgilmartin:
actually memory access fault.

Memory access fault by GPU node-4 (Agent handle: 0xe84100) on address 0x7f0ea6296000. Reason: Page not present or supervisor privilege.
Aborted (core dumped)

HostLibraryTests/llvm/LibraryPerformance_test.cpp
sdquiring: Yes, see the comment above.  I'll add a comment in the body of the test to make it clearer.

Tensile/Source/lib/include/Tensile/PropertyMatching.hpp
sdquiring: Yes, it's set by one of the TENSILE_DB bits.

sdquiring:
I believe I've fixed the memory access fault.  The problem was with a discrepancy between the MagicDivAlg value between the assembly of the kernel and the value in the TensileLibrary.yaml file used by the host runtime.

sdquiring:
The issues found in this branch are also present in develop.  Planning on merging this PR once latest pre-checkin CI passes.

PR #839, opened 2020-02-28
fixed lint issues

ramjana:


saadrahim:
Tensile/Code.py:314: local variable 'numOfColsperMfma' is assigned to but never used

There has to be better way to run on your system

PR #842, opened 2020-02-29
Enable Tox on newer systems & run lint first

sdquiring:
- `lint` environment just uses `python3` instead of `python3.5`
 - Bug in pytest has been fixed so move to newer version
 - Python 3.6 environment to run tests
 - Run lint first for quicker CI failures


eidenyoshida:
Hold off on merging for a second, it didn't pass on CI

PR #809, opened 2020-02-06
MFMA bfloat16 support for odd problem size (m- & k-dimension)

rosenrodt:
This PR completes the support for odd sized input matrices.

Running test against fairly recent ROCmSoftwarePlatform/rocBLAS@6f27dfe9 has passed

PR #855, opened 2020-03-10
Pr2 bf16 optimize

aazz44ss:
bf16 optimize

1. delay packing
in bf16, waitcnt 0 in packing will cause bubbles. there is no need to wait ds_read and do packing immediately.
Moving packing from localReadDo to makeSubIterSchedule can hide lds latency.

2. workgroup interleave
since bf16 NT has too many instructions to have good interleaving with mfma(4 ds_read + 2 v_or for packing / mfma), using 2 WG to interleave fetch instruction and mfma instruction will have better performance in some problem size.


Tensile/KernelWriterAssembly.py
```diff
@@ -7282,19 +7286,28 @@ def localReadDo(self, kernel, bufferIdx, iui, epsi, tP):
     loopChar = self.indexChars[loopDim]
     # mfma: for AB tile in NT layout
     if kernel["MatrixInstruction"] and (kernel["ProblemType"]["DataType"].isHalf() or kernel["ProblemType"]["DataType"].isBFloat16()) and numReadsAlongK > 1:
```
ramjana: where do we check the packing code is inserted only in NT case

Tensile/KernelWriterAssembly.py
```diff
@@ -7314,27 +7327,24 @@ def localReadDo(self, kernel, bufferIdx, iui, epsi, tP):
...
               # pack 2 half-dword into one
-              destVgpr = vgpr("Valu%s_X%u_I%u+%u" % (tc, bufferIdx, iui, int(valuIdx)))
-              imod.addInst("s_waitcnt lgkmcnt(0)", "")
               if kernel["MatrixInstruction"] and \
                  kernel["AssertSummationElementMultiple"] % kernel["MatrixInstK"] != 0 and \
                  self.inTailLoop:
-                imod.addInst("v_cmp_eq_u32", sgpr(isTailRemainderLoop, 2), sgpr("LoopCounter%s"%loopChar), 1, "check if last iter in tail loop")
-                imod.addInst("s_cmp_eq_u32", sgpr("NumRemainderSumElements%s"%loopChar), "0", "check if sum dim is divisible by k")
-                imod.addInst("s_cselect_b64", sgpr(isTailRemainderLoop, 2), "0x0", sgpr(isTailRemainderLoop, 2), "zero-fill only if last iter AND sum dim not divisible by k")
+                packCode.addInst("v_cmp_lt_u32", sgpr(isTailRemainderLoop, 2), sgpr("LoopCounter%s"%loopChar), kernel["MatrixInstK"], "check if last iter in tail loop")
```
ramjana: 8 instructions for packing half-words..  do we have enough instruction slots for these packing instruction that we do not create bubbles in miSIMD pipeline

rosenrodt: In main loop body there's 1 extra instruction for packing every 2 half-dwords, plus the instruction interleaving @aazz44ss proposed in this patch so they are well hidden in the shadow of MIs

What you pointed out here are for the tail loop only (originally came from odd-k problem size support back in PR #809). Admittedly there's slight perf penalty. But again these extra instruction wouldn't be there if some size assertions are satisfied, i.e., `kernel["AssertSummationElementMultiple"] % kernel["MatrixInstK] == 0`

Tensile/KernelWriterAssembly.py
aazz44ss: Currently LDS is NT layout for both NT and TN problem case, so will need packing for fp16 and bf16.

Tensile/KernelWriter.py
```diff
@@ -320,13 +321,66 @@ def makeSubIterSchedule(self, kernel, localReadCode, globalReadCode, \
...
+            packCount = packCount + 1
+          else:
+            break
+        if packCount > 0:
+          iterCode.addInst("s_nop ","1","VALU packing writes to be consumed by matrix instruction")  
+
+      iterCode.addInst("s_setprio ","1","Raise priority while processing macs")
+      iterCode.addCode(localWriteCode)
+      iterCode.addCode(pointerCode)
+      iterCode.addInst("s_setprio ","2","Raise priority while processing macs")
```
ramjana: not clear why we raising and lowering wave's priority here.   

s_setprio 3
s_setprio 1
s_setprio 2 

Are we raising priority for PACK mac code and lowering for non SIMD and raising again. ?

what other 'waves in same SIMD' are doing while PACK code is running?

PR #857, opened 2020-03-11
Modify store code for mfma 16x16x1x4

AlexBrownAMD:


Tensile/KernelWriterAssembly.py
```diff
@@ -8278,7 +8278,8 @@ def computeStoreVgprs(self, kernel, divisor, tid0Scale, tid1Scale):
                       "vectorStaticDiv vgprTmp = tid0 / matrixInstM")
           # determine row-id of each block(MFMA 'B') 2 rows for mfma32x32 4 rows for mfma16x16
           # for MFMA_4x4 16 blks are mapped along rows and cols determied by miwg0  (miwg0/4 = numRows in 16 blocks and remaining in col dimension)
-          rowIdPerColblock = globalParameters["WavefrontWidth"]//kernel["MatrixInstM"]
+          # rowIdPerColblock = globalParameters["WavefrontWidth"]//kernel["MatrixInstM"]
+          rowIdPerColBlock = 2 # Multiply by 4 for both 32x32 and 16x16, 4 simds
```
ramjana: rowIdPerColBlock = 2  doesnt work; it breaks 32x32.. 

  In 16x16 , irrespective of output blocks mapping (row or col) design always writes out 4 rows in each register ;  in 32x32 irrespective of output blocks mapping (row or col) design always writes out 2 rows in each register (8 rows registrer)

PR #864, opened 2020-03-16
avoid race condition and GPU resource contention in CI run

zaliu:
Both pre_checkin and extended ran fine in a ROCm 3.0 docker container on both gfx906 and gfx908.
Can we please quickly review this simple PR so that we can stand a better chance of getting clean Tensile CI runs - after the 8xgfx908 rack is stabilized that is.

### DIFF
--- a/Tensile/Tests/extended/pre_checkin
+++ b/Tensile/Tests/extended/pre_checkin
@@ -1,1 +1,0 @@
-../pre_checkin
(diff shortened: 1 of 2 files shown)


sdquiring:
Why were concurrent tests disabled? There is a lock to prevent multiple tests from accessing the GPU at the same time.

zaliu:
Spurious failures were observed even with -n 2 on local test desktops with 32GB VRAM and plenty of physical memory.  To facilitate debugging, concurrency was disabled for now.  With concurrency disabled, Extended runs passed consistently until a Tensile PR broke 2 Extended tests.

PR #868, opened 2020-03-19
Exact tile aware integration

wbgilmartin:
Adding the exact logic to tile aware selection tuning.

This will generate tuning results for exact sizes when tuning for the tile aware selection case which will be included as part of the final logic yaml.

I will add an exact library for the client as part of another pull request.


Tensile/ClientWriter.py
```diff
@@ -135,71 +142,91 @@ def runClient(libraryLogicPath, forBenchmark, enableTileSelection):
...
+      printWarning("ClientWriter Benchmark Process exited with code %u" % process.returncode)
+    popWorkingPath() # build
+    return process.returncode
 
 def getBuildOldClientScript(libraryLogicPath, forBenchmark):
   import io
   runScriptFile = io.StringIO()
   q = "" if os.name == "nt" else "\""
   echoLine = "@echo." if os.name == "nt" else "echo"
-  if globalParameters["NewClient"] < 2:
```
sdquiring: Again, the new client should not require this step.

Tensile/LibraryLogic.py
```diff
@@ -102,49 +112,38 @@ def analyzeProblemType( problemType, problemSizeGroups, inputParameters ):
...
-    print1("(%2u) %s : %s" % (i, \
-        Solution.getNameMin(s, solutionMinNaming), \
-        Solution.getNameFull(s)))  # this is the right name
-
-  selectionSolutionsIdsList = None 
-  selectionSolutions = None
   if enableTileSelection:
-    validSelectionSolutions = SolutionSelectionLibrary.analyzeSolutionSelection(problemType, problemSizeGroups)
+    if globalParameters["NewClient"] == 2:
+      validSelectionSolutions = SolutionSelectionLibrary.analyzeSolutionSelection(problemType, selectionFileNameList, \
```
sdquiring: Why do the old and new clients required different analysis?

Tensile/SolutionSelectionLibrary.py
```diff
@@ -74,36 +123,108 @@ def analyzeSolutionSelection( problemType, problemSizeGroups):
...
+    selectionFileName = selectionFileNameList[fileIdx]
+    numSolutions = numSolutionsPerGroup[fileIdx]
+    rowLength = solutionStartIdx + numSolutions
+    solutionBaseKeys = []
+
+    for solution in solutions:
+      baseKey = getSolutionBaseKey(solution)
+      solutionBaseKeys.append(baseKey)
+
+    selectionfFile = open(selectionFileName, "r")
```
sdquiring: The data analysis code should be encapsulated from the csv format.

Tensile/ClientWriter.py
```diff
@@ -55,14 +55,21 @@ def main( config ):
...
       "TensileCreateLibrary.cmake",
       ]
 
   for f in filesToCopy:
     shutil_copy(
         os.path.join(globalParameters["SourcePath"], f),
         globalParameters["WorkingPath"] )
+  if globalParameters["NewClient"] == 2:
+    shutil_copy(
+        os.path.join(globalParameters["SourcePath"], "CMakeListsCreateLibrary.txt"),
```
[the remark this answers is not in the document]
wbgilmartin: This and the comment below are for step 4 in the tuning. It generates the new client library in the final step.

Tensile/SolutionSelectionLibrary.py
```diff
@@ -39,12 +39,69 @@ def makeKey(row):
...
+  globalSplitU = solution["GlobalSplitU"]
+  localSplitU = solution["WorkGroup"][2]
+
+  key = "%s_%s_%s_%s" % (macroTile0, macroTile1, localSplitU, globalSplitU)
+
+  return key
+
+def updateIfGT(theDicrionary, theKey, theValue):
+  if not theKey in theDicrionary:
+    theDicrionary[theKey] = theValue
```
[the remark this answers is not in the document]
wbgilmartin: fixed

Tensile/LibraryLogic.py
wbgilmartin: The tile aware is a different model and hence different analysis/logic/algorithm.

Tensile/SolutionSelectionLibrary.py
wbgilmartin: This is actually a csv format. I am not using a csv reader.

Tensile/ClientWriter.py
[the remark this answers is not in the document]
sdquiring: CMake should still not be needed for the new client library though. Just call TensileCreateLibrary.

Tensile/LibraryLogic.py
sdquiring: But this is not tile aware vs. not, it's old client vs. new.

Tensile/SolutionSelectionLibrary.py
sdquiring: What I mean is that the file I/O code should not be mixed into the data analysis code.

Tensile/LibraryLogic.py
wbgilmartin: But it is tile aware vrs. not tile aware. 

The old version of the tile aware, which was created a year ago, generated a different output format.  This implementation used the format generated from your client app. 

Eventually when the old client gets phased out this will reduce to one case.

Tensile/SolutionSelectionLibrary.py
sdquiring: Also, file objects should use [`with` statements](https://docs.python.org/3.7/reference/compound_stmts.html#the-with-statement) wherever possible.

Tensile/BenchmarkProblems.py
```diff
@@ -502,6 +502,22 @@ def writeBenchmarkFiles(stepBaseDir, solutions, problemSizes, stepName, filesToC
...
+  maxMacroTile0 = 0
+  maxMacroTile1 = 0
+  for solution in solutions:
+    macroTile0 = solution["MacroTile0"]
+    macroTile1 = solution["MacroTile1"]
+    if macroTile0 > maxMacroTile0:
+      maxMacroTile0 = macroTile0
+    if macroTile1 > maxMacroTile1:
+      maxMacroTile1 = macroTile1
+  idealM = 36 * maxMacroTile0
```
sdquiring, line 514: Where does the number 36 come from?

PR #873, opened 2020-03-23
workaround file permissions

leekillough:
Do not depend on the permissions of the Python script.


leekillough:
https://github.com/pypa/setuptools/issues/2041

PR #874, opened 2020-03-23
Use execute_process to force TensileCreateLibrary and Tensile to have execute privileges if invoked by rocBLAS

amdkila:
- This unblocks rocBLAS
- Hopefully a better solution emerges
- This does not change any default Tensile behavior, as both executables have 755 permissions by default

### DIFF
--- a/Tensile/Source/TensileCreateLibrary.cmake
+++ b/Tensile/Source/TensileCreateLibrary.cmake
@@ -50,7 +50,10 @@ function(TensileCreateLibraryCmake
   message(STATUS "Tensile_COMPILER            from TensileCreateLibraryCmake : ${Tensile_COMPILER}")
   message(STATUS "Tensile_ARCHITECTURE        from TensileCreateLibraryCmake : ${Tensile_ARCHITECTURE}")
 
-  set(Tensile_CREATE_COMMAND "python3" "${Tensile_ROOT}/bin/TensileCreateLibrary")
+  execute_process(COMMAND chmod 755 ${Tensile_ROOT}/bin/TensileCreateLibrary)
+  execute_process(COMMAND chmod 755 ${Tensile_ROOT}/bin/Tensile)
+  
+  set(Tensile_CREATE_COMMAND "${Tensile_ROOT}/bin/TensileCreateLibrary")
 
   set(Tensile_SOURCE_PATH "${PROJECT_BINARY_DIR}/Tensile")
   message(STATUS "Tensile_SOURCE_PATH=${Tensile_SOURCE_PATH}")


leekillough:
We don't need execute permissions with #873 -- I don't understand.

Simply call the interpreter without depending on execute permissions.

amdkila:
> We don't need execute permissions with #873 -- I don't understand.
> 
> Simply call the interpreter without depending on execute permissions.

I didn't mean to remove your changes- I didn't have them on my fork at the time until I tried to commit. At that point I had to fix the merge conflict. Based on the thread, it sounds like you were encountering other issues related to your fix?

leekillough:
> > We don't need execute permissions with #873 -- I don't understand.
> > Simply call the interpreter without depending on execute permissions.
> 
> I didn't mean to remove your changes- I didn't have them on my fork at the time until I tried to commit. At that point I had to fix the merge conflict. Based on the thread, it sounds like you were encountering other issues related to your fix?

No, my fix only got rid of the dependency on execute permissions, and it's already fixed on develop.

The other issues are unrelated:

The missing pip is due to an installation error (see my recent color-coded emails). It looks like several problems were introduced all at once:
 
1.	Execute permissions are dropped (https://github.com/pypa/setuptools/issues/2041 ). **We have already worked around this issue with #873 , so this PR is unnecessary**
2.	PIP does not get installed correctly in virtualenv, and so later on, a PIP command fails, and the build fails. This may need deep-dive investigation.
3.	Paths in error traceback messages show paths with *.whl files used as directories, e.g., *.whl/path. Not sure if this is just a traceback printout bug, or whether it indicates an error caused by the bad paths.

TorreZuk:
Execute permissions just allow it to run more similarly to before.  This inside an active virtualenv so the python of that virtual environment may not even be python3, we don't do a rigorous setup of python in the root cmakelists so leaving it to run as before may avoid the pip issue you are seeing.      Can you try to run this tensile branch on your machine with the pip errors/

leekillough:
> 
> 
> Execute permissions just allow it to run more similarly to before. This inside an active virtualenv so the python of that virtual environment may not even be python3, we don't do a rigorous setup of python in the root cmakelists so leaving it to run as before may avoid the pip issue you are seeing. Can you try to run this tensile branch on your machine with the pip errors/

I'll try it, but Python2 is dead, so if we have subcomponents still depending on Python2, it's likely to break.

TorreZuk:
Sure if you try it then Akila should wait for your tests.   When you put python3 in the cmake then the launch of it may look in your path for a python3 which could be installed one, but the virtualenv also has a python3, the python script tells it to use the virtualenv one in the top line but if you specify the executable it is searched for it may or may not be the python expected.  Could be two different pythons, so I say mixed pythons

leekillough:
> 
> 
> Sure if you try it then Akila should wait for your tests. When you put python3 in the cmake then the launch of it may look in your path for a python3 which could be installed one, but the virtualenv also has a python3, the python script tells it to use the virtualenv one in the top line but if you specify the executable it is searched for it may or may not be the python expected. Could be two different pythons, so I say mixed pythons

Then the virtualenv should have its own `PATH`. `python3` should behave the same as plain `python`.

Actually if the script is executable and you run it directly, then the OS looks at the shebang for a `python` or `python3` interpreter path, which is even more hard-coded than simply invoking `python3` from CMake. (Not withstanding `#!/usr/bin/env python3` or similar tricks.)

TorreZuk:
Exactly that is why Akila's executable change lets it again use the full path to the virutalenv python3 which as it was before as they have #! directives.

TorreZuk:
Yes, you may have missed my comment at the same time as yours.   The python in the virtualenv for installation may not live outside that cmake scope so using the full venv path that gets embedded in the script is required.

PR #871, opened 2020-03-22
Full fp32/fp16/bf16 support for Transpose LDS

rosenrodt:
This should resolve a number of issues found in #853. `half`/`bfloat16` data types are also working. 

~~Though there are known issues regarding odd-k support of `half`/`bfloat16` and is working on it. Once it's complete I will remove the draft status.~~ **Update: odd-k deferred; remove draft status**

Marked MI 32x32x2x1 tests with `xfail` due to #872

PR #880, opened 2020-03-25
TransposeLDS update for f32

ramjana:
bug fixes for glvw & DepthU, innerUnroll randomization

### DIFF
--- a/Tensile/KernelWriterAssembly.py
+++ b/Tensile/KernelWriterAssembly.py
@@ -4101,26 +4101,12 @@ class KernelWriterAssembly(KernelWriter):
     waveStartSgpr = self.getTmpSgpr(1)
     if not tP["tlu"]:
       if kernel["TransposeLDS"]:
-        # add wave starting offset to Offset0I(A), Offset1J(B) registers
-        if tP["grcg"]:
-          if tP["grcv"]:
-            PerpName = tP["lvp"]
-          else:
-            # Fractional load use the more accurate lsc, multiply by VW later
-            PerpName = tP["lsp"]
-        else:
-          if tP["grcv"]:
-            PerpName = tP["lvc"]
-          else:
-            PerpName = tP["lsc"]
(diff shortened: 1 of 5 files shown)


Tensile/KernelWriterAssembly.py
```diff
@@ -7741,7 +7720,7 @@ def localReadDo(self, kernel, bufferIdx, iui, epsi, uIdx, tP):
                   InstructionTileOuput = (globalParameters["WavefrontWidth"] // kernel["MIWG0"] ) * (kernel["MatrixInstN"] // kernel["InstSplit"])
                 blockOffset = (InstructionTileOuput) * kernel["DepthU"] * tP["bpe"]
                 ldsPadOffset = (blockOffset//kernel["LdsBlockSizePerPad"])*kernel["LdsPad%s"%tc]*tP["bpe"]
-                offset = (blockOffset + ldsPadOffset)*rIdx + uIdx*kernel["MatrixInstK"]*tP["bpe"] + tP["localReadSwapByteOffset"]
+                offset = (blockOffset + ldsPadOffset)*rIdx + (uIdx)*kernel["MatrixInstK"]*tP["bpe"] + tP["localReadSwapByteOffset"]
```
rosenrodt: The unroll address increment here (`uIdx`) shouldn't be needed. There's this `tP["localReadOffset"]` gone missing here doing exactly just that, which increases the offset for each loop

This is a valid refactoring when all instances of `tP["localReadOffset"]` are removed. But maintaining 2 ways of calculating things adds confusion IMHO. You can see that's why I removed `uIdx` from `localReadDo` in PR #871 
https://github.com/ROCmSoftwarePlatform/Tensile/pull/871/files#diff-c4acdc43110b6b9fd5fee5e02641c9ebR7562

ramjana: you are right.. having K into localReadDo() gives flexibility to read 'any k value'

PR #883, opened 2020-03-25
Add pre_checkin tests for working mfma fp32 configurations

AlexBrownAMD:


amdkila:
@AlexBrownAMD Can this be merged in? Looks like it's been approved.

zaliu:
I think we're trying to compress the configurations to minimize test time.  So that part is WIP.  Thanks.

PR #884, opened 2020-03-25
Run unit tests in initial compile phase since they run very quickly.

sdquiring:


eidenyoshida:
Looks like there is one pre-checkin failure. If that is expected, I can approve this

PR #894, opened 2020-03-29
allow even longer timeout for extended CI run

zaliu:
At least one extended CI run hit 15-hour timeout last night.  Increase it to 24 hours for now.

### DIFF
--- a/.jenkins/extended.groovy
+++ b/.jenkins/extended.groovy
@@ -20,7 +20,7 @@ def runCI =
 
     boolean formatCheck = false
 
-    prj.timeout.test = 900
+    prj.timeout.test = 1440
 
     def commonGroovy
 


saadrahim:
This is not sustainable. While I see you have no choice, please think of ways to reduce the time.

zaliu:
I have the complete profile of the pytest portion of the extended runs.  We will figure out how to reduce run time.  We still need to understand why 15 hours CI limit was not sufficient while offline runs took less than 9 hours to complete.  In other words, we need to understand what "15 hours" entails.

saadrahim:
CI has reduced cores. If the tests are CPU intensive and multithreaded, it will be impacted as access to 1/8 cores of the full system is provided.

PR #903, opened 2020-04-01
fix for store code in optNLL=1  for non-mfma case

ramjana:
- missing store address calculation CinRowStart + Coord0 * bpe..


zaliu:
It would help reduce some warning messages if no_load_loop were added to pytest.ini.

PR #917, opened 2020-04-08
restore use of deprecated headers

TorreZuk:
* Ready for this now?


TorreZuk:
Could be too stale as showing strange CI errors

PR #919, opened 2020-04-09
Fix link error in hipclang builds

AlexBrownAMD:


sdquiring:
@bensander @amdkila I didn't realize this info was transferred by a global variable.  Is there a plan for passing this around instead?

amdkila:
> @bensander @amdkila I didn't realize this info was transferred by a global variable. Is there a plan for passing this around instead?

I'm not aware of any plan. The hip-clang builds are failing at an earlier point, so I believe the hip-clang team is unaware of this failure at the moment. I've seen it in our CI though.

PR #930, opened 2020-04-15
reduce parallel jobs if using hip-clang

amdkila:
- This would need to be merged in before the rocBLAS PR, as rocBLAS hits this limit when it calls TensileCreateLibrary


amdkila:
Yeah, I see both of your points about not forcing this on people, especially on those with really powerful machines. Maybe hiding this as an optional flag for TensileCreateLibrary makes the most sense

PR #931, opened 2020-04-15
reduce parallel jobs if using hip-clang in CI

amdkila:


.jenkins/common.groovy
```diff
@@ -8,6 +8,7 @@ def runCompileCommand(platform, project, jobName, boolean debug=false)
     String compiler = jobName.contains('hipclang') ? 'hipcc' : 'hcc'
     String cov = jobName.contains('hipclang') ? "V3" : "V2"
     String buildType = debug ? 'Debug' : 'RelWithDebInfo'
+    String parallelJobs = jobName.contains('hipclang') ? "export HIPCC_COMPILE_FLAGS_APPEND=-parallel-jobs=2" : ":"
```
saadrahim, line 11: what about the linker parallel jobs?

PR #932, opened 2020-04-15
Prevent exception when loading embedded code objects with mixed architecture

AlexBrownAMD:


Tensile/Source/lib/source/hip/HipSolutionAdapter.cpp
```diff
@@ -148,7 +148,12 @@ namespace Tensile
             std::vector<hipModule_t> newModules(embeddedData.size());
 
             for(size_t i = 0; i < embeddedData.size(); i++)
-                HIP_CHECK_EXC(hipModuleLoadData(&newModules[i], embeddedData[i].data()));
+            {
+                auto error = hipModuleLoadData(&newModules[i], embeddedData[i].data());
+                if(error == hipErrorUnknown || error == hipErrorSharedObjectInitFailed)
+                    continue;
```
sdquiring, line 154: Do the gaps cause issues? Should be possible to eliminate the gaps with reserve() and push_back().

AlexBrownAMD: Just pushed an update

PR #933, opened 2020-04-16
Add weighted performance script

amdkila:
Note: --merge-files doesn't allow for kernels to be loaded with new client
To-Do:
- Load all code objects in directory to ensure --no-merge-files option works as expected
- Write output to csv


Tensile/Source/client/main.cpp
```diff
@@ -150,6 +150,7 @@ namespace Tensile
                 ("perf-l2-read-bw-mul",      po::value<double>()->default_value(2.0), "L2 read bandwidth multiplier")
                 ("perf-read-efficiency",     po::value<double>()->default_value(0.85), "Read efficiency")
                 ("perf-ops-per-cycle",       po::value<int>()->default_value(64), "Ops per cycle")
+                ("best-solution",            po::value<bool>()->default_value(false), "Use best known solution")
```
amdkila, line 153: Should be removed per best solution PR

Tensile/NetworkBenchmark.py
```diff
@@ -0,0 +1,103 @@
...
+    exit(1)
+
+from . import Code
+from . import Common
+from .Common import assignGlobalParameters, globalParameters
+
+import argparse
+import os
+import subprocess
+import sys
```
amdkila, line 33: should remove sys- will use subprocess instead to invoke tensile-client

Tensile/Configs/NetworkBenchmark/test_multiple_types.yaml
```diff
@@ -0,0 +1,56 @@
...
+ProblemType3:
+  OperationType: GEMM
+  DataType: s
+  TransposeA: True
+  TransposeB: False
+  UseBeta: True
+  Batched: True
+  ProblemSizes:
+    Train:
+    - Exact: [count: 11, size: !!python/tuple [5508, 5508, 1, 3108]]
```
amdkila, line 40: I made size a tuple so that it can be used as a key in the dictionary. Lists cannot be used as keys. If there's a better way to do this, I'd like to change this

amdkila:
Closing this so it doesn't take CI cycles

PR #938, opened 2020-04-16
Linker flags to make hipcc linker include object files containing only global variables

AlexBrownAMD:


sdquiring:
This would mean that any project using this embedded mechanism would have to include these flags.  Is it possible to add these flags from the TensileCreateLibraryFiles function to have them included automatically?

AlexBrownAMD:
Integrated with PR 1030

PR #940, opened 2020-04-21
Reuse C-tile VGPR in unrolled loop

rosenrodt:
This patch aims to reuse C-tile VGPR in the unrolled loop (e.g., half-type data packing) whenever MI is enabled. It opportunistically improves CU occupancy & also reduces VGPR-constrained kernels getting rejected


ramjana:
Can we extend this PR#  into allocating only 32 register(s) for C-tile irrespective of TT0,TT1 values?.. this should help to allocate more VGPR(s) for global / local or other purpose

rosenrodt:
@ramjana we are testing a patch which overlaps C-tile registers with AB-tile & RW pointers. Should save a lot of registers. By then we are already bound not by register but LDS.

PR #943, opened 2020-04-22
updated files for DLRM tuning

ramjana:


Tensile/Configs/rocblas_sgemm_nn_inc1_asm_full.yaml
```diff
@@ -210,143 +215,160 @@ BenchmarkProblems:
           - [32, 32, 1, 2]
         - PrefetchGlobalRead: [True]
         - ThreadTile:
+          - [ 1, 32 ]
```
amdkila, line 218: These files have too many thread tile combinations. This will only work if NewClient: 0 and MergeFiles: False

zaliu:
@amdkila, please review the sizes for accuracy.  After confirming the configurations are indeed usable, please approve.  Thanks!

amdkila:
@zaliu will do. It may take a day or so for me to update this PR as there are quite a few kernels to run. I started it about 1hr 30 mins ago and I'm still on the first file (inc1_nn).

Tensile/Configs/rocblas_sgemm_nt_inc1_asm_full.yaml
```diff
@@ -322,12 +482,315 @@ BenchmarkProblems:
...
+    - # BenchmarkProblemSizeGroup - Standard
+      InitialSolutionParameters:
+      BenchmarkCommonParameters:
+        - KernelLanguage: ["Assembly"]
+        - EdgeType: ["ShiftPtr"]
+        #- LoopTail: [True]
+        - PrefetchLocalRead: [True]
+      ForkParameters:
+        - MatrixInstruction:
+          - [16, 16, 4, 1]
```
amdkila, line 500: From this point onward, I had to delete all the sizes due to an assertion error. I got numElementsPerBatch=0 for 31 kernels for the 16x16x4x1 matrix instruction. It's difficult to tell which combinations I have to remove, so I removed all of them. I didn't have this issue for the nn file, and the tn one is still running for me

PR #937, opened 2020-04-16
implement python tile aware computation

wbgilmartin:
Implement python tile aware solution.

Add jupyter notebooks for experimenting with tile aware selection

PR #947, opened 2020-04-24
Set up no-hcc hip-clang build environment (master)

amdkila:


amdkila:
Tensile has never worked on CentOS, but the missing OpenMP issue shows up for Ubuntu as well in CI. Looks like we need to fix the Jenkinsfile- I'm seeing this on other PRs as well like #937.

PR #948, opened 2020-04-24
Support alpha=1 beta=0 store part for fp16

jichangjichang:
Alpha=1 beta=0 store section is not implemented for fp16 before.
Even the problem alpha=1 beta=0, it would go into alpha !=0 store section.
Support alpha=1 beta=0 by this patch.


jichangjichang:
rocblas-test all pass
rocblas-test --gtest_filter=-*known_bug*

PR #953, opened 2020-04-29
Temporarily revert pre_checkin tests to old client pending hipclang updates

AlexBrownAMD:
Run pre_checkin tests using old client for now so we can run CI on new PRs in hipclang until a new build is available.


sdquiring:
- Is this due to the issue with linking multiple code objects? Using MergeFiles=False could be a better temporary solution in that case.
 - Using a global option in either Tensile/Tests/yaml_only/test_config.py or in .jenkins/common.groovy (`--global-parameters=MergeFiles=False`) seems like a less intrusive change.

AlexBrownAMD:
No merge files allows some tests to pass, but doesn't fix the issue with extract kernel so CI would still fail. I have a new build that should have a patch for extract kernels - testing it out now.

PR #955, opened 2020-04-30
Enable extractkernel in hipclang builds

AlexBrownAMD:


Tensile/TensileCreateLibrary.py
```diff
@@ -167,19 +167,24 @@ def isSupported(arch):
       subprocess.check_call(extractArgs, cwd=buildPath)
 
       coFilenames = ["{0}-000-{1}.hsaco".format(soFilename, arch) for arch in archs]
-    elif (CxxCompiler == "hipcc"):
 
-      hipFlags = ["--genco", "-D__HIP_HCC_COMPAT_MODE__=1"] #needs to be fixed when Maneesh's change is made available
+    elif (CxxCompiler == "hipcc"):
 
-      hipFlags += ['-I', outputPath]
```
zaliu, line 174: We shouldn't rely on env variable HIP_CLANG_HCC_COMPAT_MODE=1 in HIP-Clang docker containers.  -D__HIP_HCC_COMPAT_MODE__=1 should still be part of hipFlags for compilation.

Tensile/TensileCreateLibrary.py
```diff
@@ -167,19 +167,24 @@ def isSupported(arch):
...
+      hipFlags = ['-I', outputPath]
```
zaliu, line 174: Since we are not relying on --genco, we should explicitly specify -fPIC for compilation.

PR #956, opened 2020-05-01
Performance improvements in ContractionProblem constructor and ContractionSolution::Solve

sdquiring:
- ContractionProblem constructor
 - ContractionSolution::Solve()


Tensile/Source/lib/include/Tensile/CachingLibrary.hpp
```diff
@@ -56,34 +81,67 @@ namespace Tensile
...
+
+        template <typename SubMap, typename K, typename... Ks>
+        Value find_impl(SubMap const& map, K const& key, Ks const&... ks)
+        {
+            auto iter = map.find(key);
+
+            if(iter == map.end())
+                return m_nullValue;
+
+            return find_impl(iter->second, ks...);
```
leekillough, line 129: Why the recursive search for multivalue maps? That seems to be a bottleneck. In rocBLAS we use std::tuple for keys and define our own hash and equal operators compatible with std::map and std::unordered_map.

https://github.com/ROCmSoftwarePlatform/rocBLAS-internal/blob/2de4b9d4e0fd4dd501201878a43763f291655c46/library/src/include/tuple_helper.hpp#L63

Even though computing the hash of the entire tuple is more expensive than computing it separately for each element of the key, it allows us to search the hash table in one step (using our short-circuited equal operator), rather than traversing a tree.

sdquiring: The previous implementation used a tuple as you describe.  The problem is in order to call `find()` you need to construct a tuple, which means copying each of the members in.

Copy-constructing a `ContractionProblem` is relatively expensive as it contains 4 `TensorDescriptor`s (2 vectors each) plus several index vectors.  This plus the destructor is a significant amount of the time under CachingLibrary

Using a ContractionProblem directly as the key for an unordered map allows it to be sent in by reference, avoiding the copy altogether.

C++20's unordered_map has a [templated `find()`](https://en.cppreference.com/w/cpp/container/unordered_map/find) implementation which would allow a tuple of references, but until we can use C++20 this is the best workaround I can come up with.

sdquiring: Also the recursion depth here is limited and template defined so inlining should avoid any bottlenecks.

leekillough: > The previous implementation used a tuple as you describe. The problem is in order to call `find()` you need to construct a tuple, which means copying each of the members in.
> 
> Copy-constructing a `ContractionProblem` is relatively expensive as it contains 4 `TensorDescriptor`s (2 vectors each) plus several index vectors. This plus the destructor is a significant amount of the time under CachingLibrary
> 
> Using a ContractionProblem directly as the key for an unordered map allows it to be sent in by reference, avoiding the copy altogether.
> 
> C++20's unordered_map has a [templated `find()`](https://en.cppreference.com/w/cpp/container/unordered_map/find) implementation which would allow a tuple of references, but until we can use C++20 this is the best workaround I can come up with.

Rvalue references can reduce copies. I try to use them and perfect forwarding as much as possible.

Tuples of references are copyable, right?

I don't object to this PR, but I'd like to take a look at the code after it's merged.

leekillough: I guess what I'm saying, is that a `tuple` of references should work in C++11, if you define custom `hash` and `equal` operators, and that tuples of references are copyable (since you're constructing references from references, and aren't assigning to a mutable object). 

However, you must make sure that such references are never dangling, such as if they refer to objects which are destroyed while the tuple remains alive.

Does C++11 not allow tuples of references in containers (as keys) for some reason? I know that tuples of references are allowed, because of `std::tie` and `std::forward_as_tuple`.

I'll look at the code once it settles, and send my suggestions.

sdquiring: It's not that a tuple of references is not allowed to exist, it's that the `std::unordered_map` needs to store copies of the values used as keys (in order to avoid dangling references), so references are not appropriate there.

Under C++20, you could have a

    std::unordered_map<std::tuple<foo, bar>, baz>

which you could then use a

    std::tuple<foo const&, bar const&>

to call `find()`.  This would allow us to avoid the copy when calling `find()`.

leekillough: When I have a pass at the code, I think I'm going to try `std::shared_ptr`, so that copies are made of smart pointers instead of copies of heavy data structures. I may also try `std::map` instead of `std::unordered_map`, to see if ordering items by a short-circuited `a < b` is more efficient than by a `hash(a) == hash(b)`.

sdquiring: I'm not sure that `shared_ptr` is valid here – the `ContractionProblem` object comes from the user, who may modify the original object.

sdquiring: That also requires a breaking change to the API which would require code changes in at least 3 other projects.

leekillough: > 
> 
> That also requires a breaking change to the API which would require code changes in at least 3 other projects.

It would only be in the implementation. You said that the solve cache suffered from expensive copies of `ContractionProblem` with `std::vector` inside, etc. Tuples made that worse, because copying a tuple has to copy all of its elements. 

All that I am saying, is to use:
```
std::unordered_map<std::shared_ptr<ContractionProblem>, VAL>
```
instead of:
```
std::unordered_map<ContractionProblem, VAL>
```
so that the copying of `ContractionProblem` keys is a lightweight `std::shared_ptr` copy. We don't need the copies to be mutable, just persistent when referenced anywhere. And a copy of a `std::shared_ptr` is often just a move, when the copy can/must be elided or an rvalue `std::shared_ptr` is used.

Basically, I'm "boxing" the problem inside a smart pointer, so that standard containers and algorithms can be operated on it, without suffering the expensive costs of copying it.

I'll look at it when you've checked this in.

sdquiring: 1. This is not part of the `solve()` function which is the current bottleneck.  Are there any other changes requested?
2. Read my earlier comment about `shared_ptr`:

    std::unordered_map<std::shared_ptr<ContractionProblem>, VAL>

is invalid because the `ContractionProblem` is created by the user who could subsequently modify the underlying object, which would cause undefined behaviour.

leekillough: I was talking about using `std::shared_ptr` to store our **own copy** of the object, not a pointer to the user's copy.  This way we only incur the copy cost once, when adding it to a cache, and we do not suffer any copying costs later on or during the STL's own reordering of the map, since all copies are shallow from then on.

sdquiring:
@wbgilmartin Can you provide more info? Can we get the issues resolved?

PR #960, opened 2020-05-04
Wider sload

aazz44ss:
using s_load_dwordxN to load kernel argument, can save absolutely 3 us.


aazz44ss:
> Did we re-arrange sgpr definition so they align to x8,x16 sload instructions?

yes, I have re-arrange sgpr definition to align to argument memory in this commit.
https://github.com/ROCmSoftwarePlatform/Tensile/pull/960/commits/616b591395c01043d1d8a25bc78aa6d74af17459

PR #962, opened 2020-05-04
Use clang-offload-bundler to access hipcc-built source kernels

AlexBrownAMD:


AlexBrownAMD:
This replaces building an so and using extractkernel

PR #965, opened 2020-05-05
Fix solution state KeyError issue in multi-problem-type

solaslin:
Can be re-produced by "Tensile/Configs/rocblas_sgemm_asm_only.yaml"

For a ProblemSizeGroup satisfying: 
1) JoinParameters: MacroTile
2) ForkParameters
3) Is second or later ProblemType

A KeyError 'MatrixInstruction' will triggered when "updating solution database", shown as below:
```
# Updating Solution Database
  File "SolutionStructs.py", line 1656, in assignProblemIndependentDerivedParameters
    if state["MatrixInstruction"]:
KeyError: 'MatrixInstruction'
```
This is due to the defaultXXXXParameters are cleared by the first benchmarkProcess & step.
Fixed by replacing copy with deepcopy


zaliu:
Please merge at your earliest convenience.  Thanks.

PR #969, opened 2020-05-07
Fix SIA3 LW waitcnt calculation

aazz44ss:
fix localWrite waitcnt calculation and schedule.
original count localWriteCode Module as 1 localwrite to skip in waitcnt. However, there might have several localwrite instruction in same localWriteCode Module and cause localwrite calculation and schedule incorrect.


aazz44ss:
> looks good.. wondering if you have any seen any failures with sia3 in 16x16x4x1 instruction type..

the waitcnt calculation is not correct in sia3 without this fix when there is more than 1 LW instruction in module and more than 1 mfma per iteration, so it might have chance to have failures.

I have tested this PR with sgemm 32x32x2x1, 32x32x1x2, 16x16x4x1, 16x16x1x4, NT, TN, NN and passed all validation.

PR #974, opened 2020-05-12
Fix some linter warnings from mfma refactor

AlexBrownAMD:


rosenrodt:
@AlexBrownAMD I think we should have a separate pipeline for static code lint, as it is hardware agnostic and shouldn't prevent pytest from running

leekillough:
> 
> 
> @AlexBrownAMD I think we should have a separate pipeline for static code lint, as it is hardware agnostic and shouldn't prevent pytest from running.

I don't disagree, but this somewhat relates to #982 and automatic code reformatting. We need to choose [YAPF](https://pypi.org/project/yapf/) or a similar solution, instead of imposing draconian Lint checks as part of QA. If lint failures can be removed by virtue of a reformatter, then this issue becomes moot.

PR #978, opened 2020-05-14
clean up and add comment to local read function

TonyYHsieh:
Add detail comment to generated assembly code local read address function.
We also clean up code for two local read similar functions.
lraTileAssignmentMFMA
lraFinalOffsetVALU

### DIFF
--- a/Tensile/KernelWriterAssembly.py
+++ b/Tensile/KernelWriterAssembly.py
@@ -4789,9 +4789,9 @@ class KernelWriterAssembly(KernelWriter):
       if kernel["UnrollMajorLDS%s" % tc]:
         lds_stride = kernel["DepthU"] + LdsPad
         kStr += inst("v_mul_u32_u24", vgpr(destVgpr), hex(lds_stride), vgpr(tP["gpr"]["lwoT"]), \
-            "lw%s%s**(MT%s + PAD)"%(tP["tensorChar"], self.unrollChar, tP["tensorChar"]))
+            "lw%s%s**(DepthU + PAD)"%(tP["tensorChar"], self.unrollChar))
         kStr += inst("_v_add_lshl_u32", vgpr(destVgpr), vgpr(uReg), vgpr(destVgpr), hex(log2(tP["bpe"])), \
-            "lwFO%s = (lw%s%s + lw%s%s*(MT%s+PAD))*bpe" % (tc, tc, tc, tc, self.unrollChar, tP["tileChar"]) )
+            "lwFO%s = (lw%s%s + lw%s%s*(DepthU+PAD))*bpe" % (tc, tc, tc, tc, self.unrollChar) )
       else:
         lds_stride = kernel["MacroTile%s"%tP["tensorChar"]] + LdsPad
         kStr += inst("v_mul_u32_u24", vgpr(destVgpr), hex(lds_stride), vgpr(uReg), \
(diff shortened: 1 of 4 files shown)


Tensile/KernelWriterAssembly.py
```diff
@@ -11880,15 +11784,15 @@ def vectorStaticRemainder(qReg, rReg, dReg, divisor, tmpVgpr, tmpSgpr):
...
-    kStr += inst("_v_sub_co_u32", vgpr(rReg), "vcc", vgpr(dReg), vgpr(tmpVgpr), "vectorStaticDiv: remainder = dividend - product")
+    kStr += inst("s_mov_b32", sgpr(tmpSgpr), hex(magic), comment)
+    kStr += inst("v_mul_hi_u32", vgpr(tmpVgpr+1), vgpr(dReg), sgpr(tmpSgpr), comment)
+    kStr += inst("v_mul_lo_u32", vgpr(tmpVgpr+0), vgpr(dReg), sgpr(tmpSgpr), comment)
+    kStr += inst("s_mov_b32", sgpr(tmpSgpr), hex(shift), comment)
+    kStr += inst("v_lshrrev_b64", vgpr(tmpVgpr,2), sgpr(tmpSgpr), vgpr(tmpVgpr,2), comment)
+    kStr += inst("v_mov_b32", vgpr(qReg), vgpr(tmpVgpr), comment)
+    kStr += inst("s_mov_b32", sgpr(tmpSgpr), hex(divisor), comment)
+    kStr += inst("v_mul_lo_u32", vgpr(tmpVgpr), vgpr(qReg), sgpr(tmpSgpr), comment)
+    kStr += inst("_v_sub_co_u32", vgpr(rReg), "vcc", vgpr(dReg), vgpr(tmpVgpr), comment)
```
AlexBrownAMD, line 11795: Minor detail, but this probably doesn't need the same comment on every line. I'd probably just put the comment at the start and then something different to mark the end of the vectorStaticDiv?

TonyYHsieh: Hi Alex,

Sometime we will put /*......*/ at beginning of code block, ex: local read, which shows purpose of code block.
If we put /*...*/ at beginning of Reminder/Divide, it will confuse us with code block.
So I choose to add same comment in tail comment of instruction.
We know those comment is for divide and didn't break code block.
Does that work for you??

AlexBrownAMD: Yea, that works. Whatever you think will be most readable, and whatever helps speed up debugging the generated kernels.

TonyYHsieh:
@ramjana  I add comment and clean up some function in this PR. Does it works for you?

Tensile/KernelWriterAssembly.py
```diff
@@ -4789,9 +4789,9 @@ def lwaFirstOffset(self, kernel, tP):
       if kernel["UnrollMajorLDS%s" % tc]:
         lds_stride = kernel["DepthU"] + LdsPad
         kStr += inst("v_mul_u32_u24", vgpr(destVgpr), hex(lds_stride), vgpr(tP["gpr"]["lwoT"]), \
-            "lw%s%s**(MT%s + PAD)"%(tP["tensorChar"], self.unrollChar, tP["tensorChar"]))
+            "lw%s%s**(DepthU + PAD)"%(tP["tensorChar"], self.unrollChar))
```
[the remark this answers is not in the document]
TonyYHsieh: Hi Raman,

TransposeLDS can support different type of padding.
1. normal padding
2. block padding

I isolate block padding and transposeLDS feature.
In order to do that, we use block padding as switch to use block padding or normal padding.
If block padding == 0, normal padding else block padding.

That's why you see 0 out padding when block padding == 0.
Calculation for block padding is in different place.
You can find if block padding != 0: xxxxxx,  where do calculation for block padding.

PR #983, opened 2020-05-19
implement WaveSeperateGlobalRead feature

TonyYHsieh:
sync previous global memory pattern when transposeLDS is enabled.
This feature ask each wave to load separate region.
Original global read pattern is [w0,w1,w2,w3,w0,w1,w2,w3,w0,w1,w2,w3,w0,w1,w2,w3].
When this feature is enabled, read pattern becomes [w0,w0,w0,w0,w1,w1,w1,w1,w2,w2,w2,w2,w3,w3,w3,w3].

### DIFF
--- a/Tensile/Common.py
+++ b/Tensile/Common.py
@@ -271,6 +271,13 @@ validParameters = {
     "GlobalReadCoalesceVectorA":  [        True ], # FIXME =False worked before the vector refactor; fixing requires re-ordering load/store indices; but they aren't the faster option so not worth time right now
     "GlobalReadCoalesceVectorB":  [        True ],
 
+    # original global read to lds is interlace, [w0,w1,w2,w3,w0,w1,w2,w3,w0,w1,w2,w3,w0,w1,w2,w3]
+    # when WaveSeparateGlobalRead is enabled, LDS is divided to number of waves part.
+    # each wave load a block memory to lds,     [w0,w0,w0,w0,w1,w1,w1,w1,w2,w2,w2,w2,w3,w3,w3,w3]
+    # -1 is selected by logic, 0 disable, 1 enable.
+    "WaveSeparateGlobalReadA":    [ 0, 1 ],
+    "WaveSeparateGlobalReadB":    [ 0, 1 ],
+
     "PrefetchGlobalRead":         [ False, True ], # prefetch / double-buffer reads from global memory -> vgprs -> lds. Requires 2X LDS space, and VGPRs for buffering data on way into LDS
     "PrefetchLocalRead":          [ 0,1,2,3], # prefetch / double-buffer reads from lds (or 2 for triple-buffer, 3 for quad-buffer).  Increases size of ValuA/ValuB registers.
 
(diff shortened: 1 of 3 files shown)


Tensile/SolutionStructs.py
```diff
@@ -2068,6 +2073,27 @@ def parameterWrapper(state):
...
+  # check and calculate Wave Seperate Global Read
+  @staticmethod
+  def checkAndAssignWaveSeperateGlobalRead(state, tc):
+    # assign WaveSeperateGlobalRead
+    if state["WaveSeperateGlobalRead%s"%tc] == -1:
+      state["WaveSeperateGlobalRead%s"%tc] = 1 if state["UnrollMajorLDS%s"%tc] else 0
+
+    # check can we use WaveSeperateGlobalRead
+    numOfWaves = state["NumThreads"] // globalParameters["WavefrontWidth"]
+    if state["WaveSeperateGlobalRead%s"%tc]:
```
ramjana, line 2086: Lets not turn-on WaveSeparateGlobalRead for transposeLDS and find fractionalLoad=1 and reject solution.  user didnt ask and code turning it on and rejecting solution for fractionalLoad=1

TonyYHsieh: Hi Raman,
 Do you means to add below two conditions???
   1. set WaveSeparateGlobalRead to 0 when transposeLDS == 0
   2. reject fractionalLoad=1

Tensile/Common.py
```diff
@@ -271,6 +271,13 @@
     "GlobalReadCoalesceVectorA":  [        True ], # FIXME =False worked before the vector refactor; fixing requires re-ordering load/store indices; but they aren't the faster option so not worth time right now
     "GlobalReadCoalesceVectorB":  [        True ],
 
+    # original global read to lds is interlace, [w0,w1,w2,w3,w0,w1,w2,w3,w0,w1,w2,w3,w0,w1,w2,w3]
```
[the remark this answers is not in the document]
TonyYHsieh: Hi Raman,
This idea is from original algorithm.
I found that we use wave separate read pattern on TransposeLDS mode.
This change is add it back and use a flag to turn on or off.

A and B are independent buffer which are difference in size too.
And we see memory bandwidth is not always faster when this feature is on or off.
Win some lose some in different size.
That's why we separate  A & B.
Let A, B to find faster memory access method by search.

Tensile/SolutionStructs.py
ramjana: do not set WveSeparateGlobalRead=1 for Tlds=1 if WaveSeparateGLobalRead=-1; let it be controlled by user

Tensile/Common.py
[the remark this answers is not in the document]
ramjana: do they outperform wabveSeparateGLobalRead=0?.   we will come back and review to see if we still need separate control . may be there are other factors ( like utl1 TLB issue) for larger MT size that might explain the difference.

TonyYHsieh:
> Please make sure your code is pyflakes-clean. It isn't right now. Thanks.

Hi @zaliu ,

I re-base my change and pass lint test case.
Does pyflakes-clean means change can pass lint test case??

By the way, my previous base has lint problem in base which is fixed by AlexB changes.

PR #987, opened 2020-05-20
Tensile create libraries custom command

cgmillette:
These changes are meant to defer cmake calls of TensileCreateLibraries script to 'make' time. This will reduce the burden of cmake changes causing a call regenerate the libraries every time. Currently, changes only take effect when building for the new tensile client. Otherwise same behaviour.

Additional parameter --generate-manifest-and-exit will create a manifest file with the expected yaml/library outputs on cmake run, and then exit. This manifest list is used to create a custom command and set the outputs as dependencies for project targets. Target deps are checked at make time and then the custom command will execute as necessary.

Copy targets were added to make sure that libs are created and present when dependencies need them, especially for gtest builds that test out the targets (e.g. TensileTests).


sdquiring:
Looks good.  I had issues with `make -j` so I suspect we may have to update the CI commands.  Will have to see once #992 is in.

PR #988, opened 2020-05-21
Component system, and Navi10 FP16 non-HPA

sdquiring:
This introduces a component system to the code generator.  A component is a piece of code that is chosen among compatible components based on the current hardware capabilities and/or kernel options.

The class hierarchy is automatically used to detect which type of component a given class belongs to.  For example, all the MAC components should inherit from the MAC class.

Most components should be able to get away with defining their requirements via the class properties (e.g.):

```python
class FMA_NonPacked(MAC):
    asmCaps = {"v_fma_f16": True,
               "v_pk_fma_f16": False}
    #archCaps = {}
    kernel = {"ProblemType": {"DataType": DataType(DataType.half),
                              "HighPrecisionAccumulate": False}}
```

Values in the dictionaries can be lambdas for more advanced logic:

```python
class FMA_HPA_MAD_MIX(MAC):
    asmCaps = {"v_mad_mix_f32": True}
    #archCaps = {}
    kernel = {"ProblemType": {"DataType": DataType(DataType.half),
                              "HighPrecisionAccumulate": True},
              "LocalDotLayout": lambda ldl: ldl > 1
             }
```

Any more advanced logic should be implemented by overriding the `matches()` method.

Components are found by calling `Component.<subtype>.find(writer)` where `writer` is a KernelWriter object:

```python
    component = Component.MAC.find(self)
    if component:
      return component(self, m, innerUnroll)

    # No component was found, fall back to existing code...
```

With this fallback mechanism, components can be added one at a time.

Components can be categorized in different files in the `Tensile/Components` directory.

Feedback is encouraged.  I believe this will help make the KernelWriter files much easier to read and maintain.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,4 +55,6 @@ tags
 \#*\#
 *~
 
+# tox virtual environments
 .tox
+
--- a/HostLibraryTests/CMakeLists.txt
+++ b/HostLibraryTests/CMakeLists.txt
@@ -65,6 +65,9 @@ endif()
 
 set(INSTALL_GTEST OFF CACHE BOOL "Install GTest")
 add_subdirectory(googletest)
+if(NOT gtest_SOURCE_DIR)
+    message(FATAL_ERROR "googletest not found.  Run git submodule update --init")
+endif()
 
 find_package(Boost COMPONENTS filesystem)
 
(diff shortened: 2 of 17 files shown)


leekillough:
I added a [flake8] section to `tox.ini`, which is another place where it can be configured.

PR #989, opened 2020-05-21
Load TensileLibrary code object file if source kernels exist

amdkila:


saadrahim:
@amdkila @sdquiring is this a stale PR?

PR #991, opened 2020-05-25
fix  warning: anonymous non-C compat type

TorreZuk:
* fixes many warnings when compiling kernels in rocBLAS

### DIFF
--- a/Tensile/Source/KernelHeader.h
+++ b/Tensile/Source/KernelHeader.h
@@ -32,23 +32,22 @@
 __device__ inline int GenDot4(int a, int b, int c)
 {
 #if(__hcc_workweek__ >= 19092) || __HIP_CLANG_ONLY__
-    typedef union
+    union
     {
         int32_t i;
         char4   z;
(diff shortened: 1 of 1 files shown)


TorreZuk:
Linkage symbol probably not required anyway so just made local instances of the union.  Typedefs probably would not be declared inside helper function but in separate file with some form of naming standardization.

amdkila:
This looks fine to me, but I will let Scott review this as the code you've edited is not one that I've worked on.

leekillough:
This is why, for example:

https://github.com/ROCmSoftwarePlatform/rocBLAS-internal/blob/acdac7474b1148ef5672042b2158d2b85a650c84/library/include/rocblas-types.h#L62

has `struct rocblas_half` and `typedef ... rocblas_half` both declared.

leekillough:
The warning is that a `struct` or `union` without a tag is anonymous, and thus not C-compatible. Simply supply a tag the same name as the type, and it will be happy.

PR #992, opened 2020-05-25
Partially restore CI

sdquiring:
There is still the issue with the extended tests taking a long time.  Some work still needs to be done to figure out what state is not being properly reset in between tests.

Groovy:
 - Get rid of logic that was hiding the true exceptions
 - Fix the issues with variables not being defined

Pytest:
 - Fix some issues with the timing code
 - Update package versions

C++:
 - Improve performance of convolution reference code.


Tensile/Source/client/source/Reference.cpp
```diff
@@ -411,126 +411,118 @@ namespace Tensile
...
 
-            // Loops always traverse in same order but addressing in memory can be
-            // flexible to support different activation and filter formats
-            std::vector<size_t> s(ConvolutionProblem::MaxNumSpatialDims, 0);
-            std::vector<size_t> f(ConvolutionProblem::MaxNumSpatialDims, 0);
+            // Loops always traverse in same order but addressing in memory can be flexible to support different activation
+            // and filter formats
+            size_t spatialCoordCount = CoordCount(counts.scount.begin(), counts.scount.end());
+#pragma omp parallel for collapse(3)
             for(size_t cout = 0; cout < counts.coutCount; cout++)
```
cgmillette, line 418: I would hesitate to use 'cout' as a variable name, for the most obvious reason being if anyone using std::cout would find this out in a hurry.

PR #993, opened 2020-05-26
new feature StoreRemap to optimize MatrixInstruction store performance

jichangjichang:
This new feature is used to optimize MatrixInstruction store patterns to enhance performance.
MI output data between each threads are along N dims. But global memory is along M dim continous. That mean global write between each threads are not continous. Therefore, store performance for MI instruction is poor.

How StoreRemap works in final store stage:
1. Put all thread output data into LDS.
2. All thread read data from LDS along M dims. (match global Memory continous direction)
3. All thread write out data into global memory.

"StoreRemapVectorWidth":      [0,1,2,4,8]
0:   Disable StoreRemap (default)
1~8: Enable StoreRemap and set the global write vector width

### DIFF
--- a/Tensile/Common.py
+++ b/Tensile/Common.py
@@ -609,6 +609,21 @@ validParameters = {
     # If empty, do not use these instructions
     "MatrixInstruction":          validMatrixInstructions,
 
+    # StoreRemap: Optimize MatrixInstruction store patterns to enhance performance.
+    #             MI output data between each threads are along N dims.
+    #             But global memory is along M dim continous.
+    #             That mean global write between each threads are not continous.
+    #             Therefore, store performance for MI instruction is poor.
+    # How StoreRemap works in final store stage:
+    #             1. Put all thread output data into LDS.
+    #             2. All thread read data from LDS along M dims.
+    #                (match global Memory continous direction)
+    #             3. All thread write out data into global memory.
+    # 0:   Disable StoreRemap (default)
+    # 1~8: Enable StoreRemap and set the global write vector width
+    # Suggest optimum value: fp32 = [2,4], fp16 or bf16 = [4,8] (dwordx2 and dowrdx4)
+    "StoreRemapVectorWidth":      [0,1,2,4,8],
+
     # Disable overlapping AB-tile vgpr and read/write addr vgprs with C-tile vgprs
     # Valid only for MatrixInstruction enabled kernels, which by default overlaps
     # C-tile w/ AB-tile until it's due for v_accvgpr_read before the writeback. Illustrated below:
@@ -919,6 +934,7 @@ defaultBenchmarkCommonParameters = [
     {"ReplacementKernel":         [ False ] },
     {"MinVgprNumber":             [0]},
     {"MaxVgprNumber":             [256]},
+    {"StoreRemapVectorWidth":     [ 0 ] },
     ]
 # benchmark these solution independently
 defaultForkParameters = []
--- a/Tensile/KernelWriterAssembly.py
+++ b/Tensile/KernelWriterAssembly.py
@@ -6413,7 +6413,10 @@ class KernelWriterAssembly(KernelWriter):
         # TODO - maybe this should be encapulated in the SS setup code
         tmpSgpr = self.getTmpSgpr(2).idx()
         if ss.optSingleColVgpr:
-          tmpVgpr= None # don't need tmps for other uses?
+          if not kernel["StoreRemapVectorWidth"]:
+            tmpVgpr= None # don't need tmps for other uses?
+          else:
+            tmpVgpr= self.vgprPool.checkOut(1, "NLL store remap tmps")
         else:
           tmpVgpr= self.vgprPool.checkOut(3, "NLL address tmps")
         assert(len(kernel["PackedC1IndicesX"]) == 1)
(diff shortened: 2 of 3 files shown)


Tensile/KernelWriterAssembly.py
```diff
@@ -9155,6 +9192,274 @@ def notLocalFullTileElements(self, kernel, edge):
         return self.notLocalFullTileElementsMFMA(kernel, edge)
     else:
         return self.notLocalFullTileElementsVALU(kernel, edge)
+  ##############################################################################
+  # Store Remap: Local Write
+  ##############################################################################
+  def storeRemapAddLocalWrite(self, kernel, ss, addrCalc, sumIdx):
```
ramjana, line 9198: write offset address do they write MT0 elements contiguously  or write vgprC as  xdlops outputed into LDS and read gather MT0 elements?

i imagine reducing bankconflicts in read critical for global store that depends on read

Tensile/KernelWriterAssembly.py
```diff
@@ -9155,6 +9192,274 @@ def notLocalFullTileElements(self, kernel, edge):
...
+    gwvw = kernel["StoreRemapVectorWidth"]
+    startIdx = self.storeRemapStartSumIdx
+    endIdx = self.storeRemapEndSumIdx
+    bpe = kernel["ProblemType"]["DataType"].numBytes()
+    bps = kernel["ProblemType"]["DataType"].numBytes() * gwvw
+    rpe = kernel["ProblemType"]["DataType"].numRegisters()
+    rpv = kernel["ProblemType"]["DataType"].numRegisters() * gwvw
+
+    src = vgpr(self.storeRemapLR)
+    for i in range (startIdx, endIdx, gwvw):
```
ramjana, line 9242: i wonder b2b ds_read_xx affect inversely for mi32x32 with bigger MT0?  given that there isnt much math left to do to hide wrte->read-> global-write..

ramjana: can you think of using existing store code and add waitcnt() when needed?. may be for next optimization..

Tensile/Common.py
```diff
@@ -602,6 +602,20 @@
...
+    #             That mean global write between each threads are not continous.
+    #             Therefore, store performance for MI instruction is poor.
+    # How StoreRemap works in final store stage:
+    #             1. Put all thread output data into LDS.
+    #             2. All thread read data from LDS along M dims.
+    #                (match global Memory continous direction)
+    #             3. All thread write out data into global memory.
+    # 0:   Disable StoreRemap (default)
+    # 1~8: Enable StoreRemap and set the global write vector width
+    "StoreRemapVectorWidth":      [0,1,2,4,8],
```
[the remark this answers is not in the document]
jichangjichang: My original implementation ever considers this. 
But for "StoreRemap disable" case, store vector width is always 4 for MatrixInstruction. it should not use other vector width.

If I just add "enable/disable"for StoreRemap, and adjust different Store Vector width, for example:
StoreRemap= [0,1]
StoreVectorWidth = [1,2,4]
6 cases (2x3) would be generated.

But we only want use below 4 cases:
1. StoreRemap =  disable, storeVectorWidth = [4]
2. StoreRemap =  enable,  storeVectorWidth = [1]
3. StoreRemap =  enable,  storeVectorWidth = [2]
4. StoreRemap =  enable,  storeVectorWidth = [4]
we will never want use below 2 cases :
1. StoreRemap =  disable, storeVectorWidth = [1]
2. StoreRemap =  disable, storeVectorWidth = [2]

That's why i use one parameter to determine which storevectorwidth we should use and only for storeRemap. It should not touch and change for storeRemap disabled case.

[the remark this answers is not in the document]
ramjana: with store re-order,  vw4 (for f32) vw2 (for half-precision) should be optimal value for all problem size ?  minimum number of rows per wave for xdlops >=4. default should be set to 4/2 for store-reorder ; with comments in common.py user should be able to use optimum value in all cases.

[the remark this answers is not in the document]
jichangjichang: After re-order, best vw for f32 is 4 and 2.
Best vw for half-precision is 8 and 4. (Dwordx4 and Dwordx2)
which one is better depends on wave block shape and ldd.
I will add some comments about optimal values into Common.py

Tensile/KernelWriterAssembly.py
jichangjichang: It write vgprC same as xdlops output into LDS.
each thread write vw4 into different lds column.

we have lds padding to prevent the bandconflict.

Tensile/KernelWriterAssembly.py
jichangjichang: > i wonder b2b ds_read_xx affect inversely for mi32x32 with bigger MT0? given that there isnt much math left to do to hide wrte->read-> global-write..

It wouldn't really happen due to LDS resource has limitation.
For example: 
fp16 MT256x256 with mi32x32 and 4 waves, it would consume (256(MT0)+4 padding) x 32 (MI32x32)x 4 (waves) * 2 (bpe) = 66560.bytes lds.
This kernel will be rejected by solutionStruction.py

jichangjichang: > can you think of using existing store code and add waitcnt() when needed?. may be for next optimization..

I ever did this in my implementation.
But I think that the code is better separated cleanly with existing store code.
For storeRemap store code, it needs to consider "local read" and "edge" and "waitcnt".
It would be more complicated and original store code is very simple.
Separate those would be more easily for reading and maintenance.

Thanks for your suggestions anyway.

Tensile/Common.py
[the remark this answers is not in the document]
jichangjichang: > with store re-order, vw4 (for f32) vw2 (for half-precision) should be optimal value for all problem size ? minimum number of rows per wave for xdlops >=4. default should be set to 4/2 for store-reorder ; with comments in common.py user should be able to use optimum value in all cases.

Comments are just  added into common.py
Please check it, thanks.

PR #994, opened 2020-05-26
Fix kernel segfaults using staggerU with tensor sizes > 2^32

cgmillette:
For these kernels, the calculation of the stagger offsets need to be augmented to 64 bit calculations to ensure that tensors in col-major format (N) with sizes > 2^32 can be handled.

This PR will address the 64 bit calculations in both unroll loop and tail section addressing updates of the kernels. Will also add tests for regression covering these cases where skinny tensors can get > 2^32 in size.

### DIFF
--- a/Tensile/Contractions.py
+++ b/Tensile/Contractions.py
@@ -229,7 +229,7 @@ class ProblemType:
         if includeOperation:
             predicates.append(ProblemPredicate("OperationIdentifierEqual", value=self.operationIdentifier))
             if not self.useBeta:
-                predicates.append(ProblemPredicate("BetaZero"));
+                predicates.append(ProblemPredicate("BetaZero"))
 
         if includeType:
             predicates.append(ProblemPredicate("TypesEqual", value=(self.aType, self.bType, self.cType, self.dType)))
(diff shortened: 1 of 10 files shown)


zaliu:
This PR broke rocBLAS build.  The instruction s_mul_hi_i32 is *not* supported on gfx803.  Please either fix this issue right away or revert this PR for now.  Feature Complete day is Wednesday.  We can't afford to have Tensile breaking rocBLAS build.  Thanks!

rosenrodt:
It looks like we really need some TensileCreateLibrary CI test against some subset of `asm_ci` or `asm_full`.

zaliu:
> It looks like we really need some TensileCreateLibrary CI test against some subset of `asm_ci` or `asm_full`.

Indeed.  Currently we encourage all Tensile developers to make sure their siginficant PRs can build rocBLAS (and pass some subset of rocblas-test) before merging, which is a pain.  At any rate, in this case, testing only `asm_ci` won't catch the build failure because it's related to gfx803, which isn't covered by `asm_ci`.  For that matter, as @sdquiring pointed out, no CI tests gfx803 - other than when building rocBLAS with `asm_full`.

PR #995, opened 2020-05-27
disable pre_checkin/regression/persistent_kernel.yaml BoundsCheck for now

zaliu:
Disable BoundsCheck for pre_checkin/regression/persistent_kernel.yaml temporarily to avoid SEGV.


amdkila:
Do the pre-checkin tests pass on hip-clang after this fix?

zaliu:
> Do the pre-checkin tests pass on hip-clang after this fix?

pre_checkin tests pass on hcc after this fix.  Haven't reached hip-clang yet.  A step at a time.

PR #997, opened 2020-05-29
Revert "Partial support of tuning strided batched GEMMs"

sdquiring:
Strangely, the new test introduced in this PR is failing in CI but not when run locally.  Revert until we can figure out why this is the case.


rosenrodt:
Seems to be related to #995. Try adding `BoundsCheck: False` in the test yaml in question. I will detail why later

zaliu:
> It looks like same kind of failure to what #995 was trying to fix. The test suite seems to silently set `BoundsCheck=true` in ClientParameters.ini even when `NumElementsToValidate != -1`, which violates the condition that BoundsCheck can only be turned on when you validate all the elements. I cannot reproduce the same behavior if the test is directly run against using `Tensile/bin/Tensile`.
> 
> I will open another PR with the BoundsCheck explicitly set false or cross out the number of validations—see if it helps

OK, we'll hope off on this revert until your experiment is done.  Thanks!

zaliu:
Close for now.  Can re-open if needed.

PR #1010, opened 2020-06-03
ROCm 3.6 merge develop into master

zaliu:


zaliu:
> > Seems like on hcc, we have a test failure with the ReplacementKernel bad_file test. Other tests pass
> 
> Thanks, updated with a more permanent replacement kernel name this time.

Not a good idea.  The short kernel name doesn't exist in COv3 subdirectory.  Oh well, back to the ever-changing kernel name.

PR #1012, opened 2020-06-03
Fix leading dims in tuning scripts and other minor fixes

amdkila:
- Wait until after branching to merge


tuning/automation/GenerateTuningConfigurations.py
```diff
@@ -95,7 +95,7 @@ def processFile(headerFileName, key, configDefinitionList, configurationPath, wo
 
 def SetDefaultStrides(problemDefinition, m, n, k):
     if problemDefinition["transposeB"] == "T":
-        return [k, k, n]
+        return [m, n, m]
```
zaliu, line 98: If transposeA is true here, default lda would be k, not m, [k, n, m] should be returned.  Instead it returns [m, n, m] in error.

amdkila: are you referring to TT sizes? I've made some assumptions that we don't encounter TT sizes

zaliu: > are you referring to TT sizes? I've made some assumptions that we don't encounter TT sizes

Then please add a brief comment to SetDefaultStrides stating the assumption.  Thanks.

amdkila: Good idea. I added that in my next commit

PR #1013, opened 2020-06-03
Config file for testing bert sizes

AlexBrownAMD:


AlexBrownAMD:
> Can you share the length of time needed to run through this tuning configuration file? It's hard to figure how many valid configurations were actually used. Thanks.

I'm not sure how long it takes to run in this format, I ran parts of this incrementally to measure performance changes with each feature. There is a rather large number of valid configurations though, and running the full file would take a number of hours. @amdkila is currently running this with additional sizes. Do you need precise numbers for run-time or configurations included?

zaliu:
> > Can you share the length of time needed to run through this tuning configuration file? It's hard to figure how many valid configurations were actually used. Thanks.
> 
> I'm not sure how long it takes to run in this format, I ran parts of this incrementally to measure performance changes with each feature. There is a rather large number of valid configurations though, and running the full file would take a number of hours. @amdkila is currently running this with additional sizes. Do you need precise numbers for run-time or configurations included?

I assume you're covering a small subset of the BERT sizes in this file, right?  Are you saying even with the small subset of the sizes you picked, it's going to take hours to run through this file?  That's the concern I had.  Normally we ran a tuning configuration, looked at the winners (and almost winners) to decide which options offered very little, if any, benefit, then we prune them to reduce tuning time.  Oh well, as long as @amdkila isn't complaining, I'm happy.

amdkila:
> > > Can you share the length of time needed to run through this tuning configuration file? It's hard to figure how many valid configurations were actually used. Thanks.
> > 
> > 
> > I'm not sure how long it takes to run in this format, I ran parts of this incrementally to measure performance changes with each feature. There is a rather large number of valid configurations though, and running the full file would take a number of hours. @amdkila is currently running this with additional sizes. Do you need precise numbers for run-time or configurations included?
> 
> I assume you're covering a small subset of the BERT sizes in this file, right? Are you saying even with the small subset of the sizes you picked, it's going to take hours to run through this file? That's the concern I had. Normally we ran a tuning configuration, looked at the winners (and almost winners) to decide which options offered very little, if any, benefit, then we prune them to reduce tuning time. Oh well, as long as @amdkila isn't complaining, I'm happy.

I have all the sizes in a larger yaml file available [here](https://github.com/amdkila/Tensile/commit/441526df33886f1064c7df5b29c28044a9d766fd). I can add these to the PR, but I haven't had the chance to run through all the kernels yet.

PR #1015, opened 2020-06-04
modify logic of ldsPad=-1

aazz44ss:
we need to pad lds to avoid bank conflict on both ds_read and ds_write
In MatrixInstruction and TLDS,
we'll have MIInput element per ds_read and GRVW element per ds_write


aazz44ss:
not upload wider local read yet,
once we have wider local read, ldsPad should be max(GRVW,LRVW)

PR #1016, opened 2020-06-04
updated transformer model configurations

ramjana:
pruned configurations files to keep number solutions for each data formats small
should take ~2 for all three data types..


ramjana:
> Since FractionalLoad defaults to [0], instead of removing the line, you want to always force it to be [1]. Thanks.

fractionalLoad =1  helps to use wider global load when MT is small and not every threads are used to fetch elements. I ran with FractionalLoad=0 && FractionalLoad==1 and hardly seen any different in performance. I think its fine to just keep the default for transformer model

zaliu:
> > Since FractionalLoad defaults to [0], instead of removing the line, you want to always force it to be [1]. Thanks.
> 
> fractionalLoad =1 helps to use wider global load when MT is small and not every threads are used to fetch elements. I ran with FractionalLoad=0 && FractionalLoad==1 and hardly seen any different in performance. I think its fine to just keep the default for transformer model

Some configurations aren't compatible with FractionalLoad=0, and you'll miss opportunities FractionalLoad=1 affords you.  Hey, it's your call.

PR #1018, opened 2020-06-05
host-side assertion to check bufferload/bufferstore offset limit

solaslin:
Add new predicates to check if a problem will cause bufferload offset >= 2^32 for a kernel. 
if DID_NOT_SATISFY_ASSERTS, then the kernel will not be used for the problem to avoid incorrect gemm result.
The yamls under Tests/extended/bufflerload_limit show these kind of problems.


solaslin:
@cgmillette thanks for the recommendation. Now I use a Packet struct to do this. It took me some time figuring our how to handle the iot.mapRequired.

PR #1021, opened 2020-06-09
Add new merge script

amdkila:
The new merge script has the following features: 

1) **Default behavior:** Merge in tuned kernels for previously tuned sizes only if the efficiency has improved. This also **removes unused kernels**

e.g. python avoid-regressions.py  asm_full  incremental_logic_dir  output_dir

2) **Old behavior:** Merge in tuned kernels for previously tuned sizes regardless of improvements in efficiency

e.g. python avoid-regressions.py  asm_full  incremental_logic_dir  output_dir  true

**Recommended usage:** Use the first option always. Only use the 2nd option if you're working on a GPU with lower peak performance than another GPU with the same architecture, i.e. MI50 vs. MI60 tuning


rosenrodt:
@amdkila does the new merge script preserve leading dimension info in logic yaml? Because old script reduces 8-tuple problem key to 4-tuple upon merge.

amdkila:
@rosenrodt It doesn't, but I can change that if needed. From my understanding, kernel selection ignores leading dimensions. So that's why it's ok for the old merge script wiping out the leading dimensions. For that reason, I decided to intentionally do the same with the new script.

rosenrodt:
@amdkila Given the fact that LDs have good amount of impact on kernel selection, the Tensile client's behavior might change in the future. Especially given the Tensile's ability to tune for strided batched gemms (though with limited support). Erasing the strides/LDs info for now may be OK, but will probably come back and bite us in the future.

amdkila:
This PR should be good to go. Now there isn't a need to merge the base file into incremental logic, and all solutions are removed as expected

PR #1023, opened 2020-06-09
Fix stagger u offset calculation

cgmillette:
Revive fix for staggerU offset calculation

- Adjusted for the case when mulhi instructions are not supported on target architecture.


sdquiring:
@zaliu can you help test this with rocBLAS to verify that the issue is fixed?

zaliu:
> @zaliu can you help test this with rocBLAS to verify that the issue is fixed?

On gfx906, old-client based rocBLAS builds fine, and rocblas-test passed quick and pre_checkin.

zaliu:
> @zaliu can you help test this with rocBLAS to verify that the issue is fixed?

In hipclang docker 2723 on gfx906, rocBLAS with this Tensile PR built fine and rocblas-test passed quick and pre_checkin.

PR #1028, opened 2020-06-15
BF16/FP16 GSU global accumulation

TonyYHsieh:
use a extra global buffer for GlobalSplitU accumulation to implement BF/FP high precision GSU.
rocblas also need to related change due to interface change.
Need a extra buffer which is passed through Tensile interface.

RocBLAS PR: https://github.com/ROCmSoftwarePlatform/rocBLAS-internal/pull/132

### DIFF
--- a/Tensile/TensileCreateLibrary.py
+++ b/Tensile/TensileCreateLibrary.py
@@ -304,12 +304,11 @@ def buildKernelSourceAndHeaderFiles(results, outputPath, kernelsWithBuildErrs, \
 ################################################################################
 # Write Solutions and Kernels for BenchmarkClient or LibraryClient
 ################################################################################
-def writeSolutionsAndKernels(outputPath, CxxCompiler, problemTypes, solutions, kernels, kernelsBetaOnly, \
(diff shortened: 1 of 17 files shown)


Tensile/TensileCreateLibrary.py
```diff
@@ -1117,18 +1153,32 @@ def TensileCreateLibrary():
         masterLibraries[architectureName].version = args.version
 
   # create solution writer and kernel writer
-  kernels = []
-  kernelsBetaOnly = []
+  kernels            = []
+  kernelsBetaOnly    = []
+  kernelsGlobalAccum = []
```
sdquiring, line 1158: Can we collectively call these "helper" kernels and have one list?

TonyYHsieh:
> Since these kernels require additional memory allocation on the part of the user, they need to be opt-in.
> 
> There needs to be:
> 
> * A flag in the `ContractionProblem` which allows the user to specify whether a temporary buffer is available.
> * A predicate attached to the relevant solutions which looks at this flag and prevents those solutions from being selected if the flag is not set.
> * A method in the `ContractionSolution` which calculates the required size of the temporary buffer, based on the `ContractionProblem`.

Hi @sdquiring ,

Thanks for your feedback.
We will modify this change according to your feedback.

PR #1031, opened 2020-06-15
GSU support DGEMM

ramjana:
- Added GSU support for DGEMM
- added yaml files for BDAS workloads (kmeans, pca)


zaliu:
Full rocblas-test passed on gfx906.

PR #1039, opened 2020-06-17
Support GRVW up to 8 for fp16/bf16 NT B matrix MI kernel

jichangjichang:
- support GRVW8 for NT B matrix

shift-ptr correction for MI kernel B matrix can be larger then MIOutputVectorWidth size.
(For A matrix shift-ptr design, data are corrected in single thread. It is limited by MIOutputVectorWidth size.  
But for B matrix, the data are corrected by re-calculating coordination. It doesn't have MIOutputVectorWidth limitation.)
Now GLVB=8 is rejected by MIOutputVectorWidth condition check.
Modify reject condition and GLVA calculation to support GRVW=8 for B matrix.

### DIFF
--- a/Tensile/SolutionStructs.py
+++ b/Tensile/SolutionStructs.py
@@ -1761,7 +1761,7 @@ class Solution:
   #  state[GlobalLoadVectorWidth*]
   #  state[NumLoads*] # only used in SolutionStructs, with classic alg
   @staticmethod
-  def setGlobalLoadVectorWidth(state, tc, totalVectors):
+  def setGlobalLoadVectorWidth(state, tc, totalVectors, grvw):
     validDepthU = True
     if totalVectors < state["NumThreads"]:
       # Try to reduce size of vector so every thread has a load to do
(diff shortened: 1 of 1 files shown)


Tensile/SolutionStructs.py
```diff
@@ -2499,11 +2499,24 @@ def assignDerivedParameters(state):
...
+        if not Solution.setGlobalLoadVectorWidth(state, "A", tva, state["GlobalReadVectorWidth"]):
           validDepthU = False
-        if not Solution.setGlobalLoadVectorWidth(state, "B", tvb):
+        if not Solution.setGlobalLoadVectorWidth(state, "B", tvb, state["GlobalReadVectorWidth"]):
           validDepthU = False
 
+        if state["EnableMatrixInstruction"] and state["GlobalLoadVectorWidthA"]:
+          partialA = state["ProblemType"]["TLUA"] and (state["AssertFree0ElementMultiple"]%state["GlobalLoadVectorWidthA"] != 0)
+          if partialA and state["GlobalLoadVectorWidthA"] > state["MIOutputVectorWidth"]:
+            #reduce GLVA if GLVA larger than MIOVW
```
ramjana, line 2510: In partial case, tensile should choose mioutputvectorWidth over GLVW; is this for to avoid partial altogether or write shift-ptr logic to work?

jichangjichang: Yes.
For A matrix, shift-ptr only work when GLVW  <= MIoutputVectorWidth.
For B matrix, shift-ptr can work when GLVW <= MatrixInstN.

This code to reduce GLVWA size to MIOutputVectorWidth. Later shift-ptr checking will not reject this solution.
It let GLVWB has chance to use higher GLVW. Such as GLVWB=8.

jichangjichang: > In partial case, tensile should choose mioutputvectorWidth over GLVW; is this for to avoid partial altogether or write shift-ptr logic to work?
If without this patch, tensile only reject the solution that GLVW > MIOVW. It wouldn't choose MIOVW over GLVW.

PR #1042, opened 2020-06-19
Tensile create libraries custom command

cgmillette:
- Revive PR 987 https://github.com/ROCmSoftwarePlatform/Tensile/pull/987
- Defer the full call to TensileCreateLibrary to 'make' time instead of 'cmake' time
- Allow the script to generate a manifest file and use the contents to create new target dependencies
- To be used with new client only (see HostLibraryTests CMake for usage example)
- Projects migrating to the new client should use TensileCreateLibraryFiles from TensileConfig.cmake
- Results in less time wasted regenerating library files when not necessary


Tensile/TensileCreateLibrary.py
```diff
@@ -1165,22 +1306,68 @@ def TensileCreateLibrary():
...
                                              kernelWriterSource, kernelWriterAssembly)
 
+  sanityCheck0 = set(codeObjectFiles) - set(sourceLibPaths + asmLibPaths)
+  sanityCheck1 = set(sourceLibPaths + asmLibPaths) - set(codeObjectFiles)
+
+  if len(sanityCheck0) > 0:
+    print1("Warning: Manifest is missing library files: {}".format(sanityCheck0))
+
+  if len(sanityCheck1) > 0:
+    print1("Warning: Manifest has unexpected library files: {}".format(sanityCheck1))
```
sdquiring, line 1368: What are the situations where this could happen? Should this be an assert? Should it throw an error?

PR #1043, opened 2020-06-20
Add library-format flag and fix Tensile issues

amdkila:
- Makes yaml default for Tensile, and msgpack default for TensileCreateLibrary
- Adds library-format flag, and new Tensile_LIBRARY_FORMAT cmake variable (will need to be added to rocBLAS)
- Works with NewClient: 0, 1, and 2 for both msgpack and yaml


amdkila:
Never mind, looks like TensileCreateLibrary is broken since I fixed the merge conflicts. Something about LibraryFormat being unrecognized. Will leave that for Monday

Tensile/Common.py
```diff
@@ -164,7 +164,7 @@
 globalParameters["MaxFileName"] = 128 # If a file name would be longer than this, shorten it with a hash.
 globalParameters["SupportedISA"] = [(8,0,3), (9,0,0), (9,0,6), (9,0,8), (10,1,0), (10,1,1)]             # assembly kernels writer supports these architectures
 globalParameters["ClientBuildPath"] = "0_Build"                   # subdirectory for host code build directory.
-globalParameters["NewClient"] = 1                                 # 1=Run old+new client, 2=run new client only (All In)
+globalParameters["NewClient"] = 2                                 # 1=Run old+new client, 2=run new client only (All In)
```
sdquiring, line 167: Is this what we want as the new default?

Tensile/LibraryLogic.py
```diff
@@ -1444,7 +1444,7 @@ def main(  config ):
     logicTuple = analyzeProblemType( problemType, problemTypes[problemType], \
         analysisParameters)
 
-    LibraryIO.configWriter(globalParameters["YAML"]).writeLibraryLogicForSchedule(globalParameters["WorkingPath"], \
+    LibraryIO.configWriter(globalParameters["LibraryFormat"]).writeLibraryLogicForSchedule(globalParameters["WorkingPath"], \
```
sdquiring, line 1447: I don't think we want this to follow LibraryFormat.  This should be YAML no matter what, correct?

amdkila: Yeah, you're right. The next commit makes both of the changes you've requested

Tensile/Source/TensileCreateLibrary.cmake
```diff
@@ -30,6 +30,7 @@ function(TensileCreateLibraryCmake
     Tensile_COMPILER
     Tensile_CODE_OBJECT_VERSION
     Tensile_ARCHITECTURE
+    Tensile_LIBRARY_FORMAT
     Tensile_MERGE_FILES
     Tensile_SHORT_FILE_NAMES
     Tensile_LIBRARY_PRINT_DEBUG )
```
TorreZuk, line 36: I'll also point out that you are changing the order of arguments to the function CreateLibraryCmake so all callers must modify their calls.   Note the comment in ARGN handling below, about breaking the old API, it is easy enough to update rocBLAS but not sure about other callers outside tensile.

PR #1046, opened 2020-06-23
directToLDS: implement UseInstOffsetForGRO

TonyYHsieh:
implement UseInstOffsetForGRO:
1. use instruction offset instead of update m0 for global read to LDS.
2. move global read address to (base - 4096) to prevent negative GRO sgpr offset

### DIFF
--- a/Tensile/KernelWriterAssembly.py
+++ b/Tensile/KernelWriterAssembly.py
@@ -1047,6 +1047,8 @@ class KernelWriterAssembly(KernelWriter):
     buffer_load_short = MemoryInstruction("buffer_load_short_d16", 1, 0, 0, 0.5, \
         "UNUSED %s, %s, %s, %s offen offset:0 %s" )
 
+    self.buff_load_inst_offset_max = 4096
+
     ########################################
     # Global Write
     flat_store_dwordx4 = MemoryInstruction("flat_store_dwordx4",  1, 0, 0, 4, \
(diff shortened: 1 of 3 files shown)


TonyYHsieh:
Hi @ramjana , 

May you help to review this change???

Thanks, Tony.

ramjana:
InstructionOffset feature will fail if offset value > 16K.  is there assertion built-in for that?. I understand stall cycle introduced between mov  m0, offset  and buffer_load. we need to live that.

TonyYHsieh:
> InstructionOffset feature will fail if offset value > 16K. is there assertion built-in for that?. I understand stall cycle introduced between mov m0, offset and buffer_load. we need to live that.

Hi Raman, in order to live with this limitation, we separate offset to two part. 

1. (offset/4096)*4096 => sgpr or vgpr offset
2. offset%4096 => instruction offset

We still have to update m0 after instruction offset is larger than 4K, but we don't have to update for each instruction.

By the way, according to "Gfx9_Shader_Programming p.77", instruction offset is a 12-bit field.
So I use 4096 as dividend.

ramjana:
(offset/4096)*4096 => sgpr or vgpr offset
offset%4096 => instruction offset

VMEM instruction usually takes 4 cycle to issue.
using SGPR for offset in LDS_DMA instruction results in longer instruction issue latency.  CU require two SGPR reads one for resource and another for SGPR offset.  4 -cycles for SRD and 4 -cycle for SGPR offset read. with that instruction takes 8 cycles and halving issue rate of vmem.

ramjana:
did i close this accidentally?

Tensile/KernelWriterAssembly.py
```diff
@@ -1005,6 +1005,8 @@ def initKernel(self, kernel, tPA, tPB ):
     buffer_load_short = MemoryInstruction("buffer_load_short_d16", 1, 0, 0, 0.5, \
         "UNUSED %s, %s, %s, %s offen offset:0 %s" )
 
+    self.buff_load_inst_offset_max = 4096
```
ramjana, line 1008: would be good idea to define offset limit variable in memory Instruction and use that here. so in future easier to update

zaliu:
@TonyYHsieh, what's your plan on this PR?

TonyYHsieh:
Re-base PR.  We can merge this PR after CI testing and review.

PR #1051, opened 2020-06-24
Use LLVM 9.0 on CentOS

AlexBrownAMD:
Tell cmake to prefer using LLVM 9.0 over 11.0 on CentOS to avoid header incompatibilities. This fixes linker error when building rocBLAS.


AlexBrownAMD:
> So this 9.0 is for Centos 8, did SLES need another version or we need to wait and see?

Looking into SLES next, will post an update once I've tried the build there.

PR #1052, opened 2020-06-25
Use LLVM 7.0 on SLES

AlexBrownAMD:
Add option to find LLVM 7 package (version installed on our SLES builds).


AlexBrownAMD:
> Good for now. Not sure why cmake doesn't supported a version request list

Yea, spent a while searching for an alternative to the nested list of versions. The closest I found is cmake allows changing the sort order of found configs (natural asc des). But since we want any version between 6 and 9, there doesn't seem to be a tidier way to implement that with find_package.

PR #1055, opened 2020-06-27
Component for function signature for code object V2 and V3

AlexBrownAMD:


zaliu:
There are some flake8 issues:
$ flake8 Tensile
Tensile/Components/Signature.py:345:9: F841 local variable 'kernArgBytes' is assigned to but never used
Tensile/Components/Signature.py:389:9: F841 local variable 'cptAlign' is assigned to but never used

AlexBrownAMD:
> If there are many globalParameter values that will need to be accessed this way then yes we should include that here, otherwise we should consider including code object version as a kernel parameter.

Added CodeObjectVersion to kernel

PR #1057, opened 2020-06-30
Make msgpack support optional

sdquiring:
This can wait until after the release.

This makes YAML and msgpack support each options that can be selected independently (either, both, or neither).


Tensile/Source/lib/CMakeLists.txt
```diff
@@ -101,6 +110,13 @@ if(TENSILE_USE_LLVM)
     target_include_directories(TensileHost PRIVATE ${LLVM_INCLUDE_DIRS})
 endif()
 
+if(TENSILE_USE_MSGPACK)
+    target_compile_definitions(TensileHost PUBLIC -DTENSILE_MSGPACK=1)
+
+    find_package(msgpack)
+    target_link_libraries(TensileHost PUBLIC msgpackc)
```
TorreZuk, line 117: This is now duplicate and outdated content so can be removed and the target_compile_definition moved up to the earlier if (TENSILE_USE_MSGPACK)

amdkila:
I'm testing this now. How do I select neither?

Tensile/Source/lib/CMakeLists.txt
sdquiring: I'll resolve this in #1071 to avoid having to rebase.

PR #1058, opened 2020-06-30
only add dependency when on and links as private static

TorreZuk:
This overlaps slightly with PR#1057 so maybe can be integrated into it, but seems to allow hiding the shared library dependency from propagating down the application chain by statically linking msgpackc into TensileHost.  Please comment

### DIFF
--- a/Tensile/Source/lib/CMakeLists.txt
+++ b/Tensile/Source/lib/CMakeLists.txt
@@ -77,8 +77,10 @@ add_library (TensileHost STATIC ${tensile_sources})
 
 target_include_directories(TensileHost INTERFACE "${CMAKE_CURRENT_SOURCE_DIR}/include")
 
-find_package(msgpack)
-target_link_libraries(TensileHost PUBLIC msgpackc)
+if(Tensile_LIBRARY_FORMAT MATCHES "msgpack")
+    find_package(msgpack)
+    target_link_libraries(TensileHost PRIVATE msgpackc.a)
+endif()
 
 if(TENSILE_USE_LLVM)
     target_compile_definitions(TensileHost PUBLIC TENSILE_DEFAULT_SERIALIZATION)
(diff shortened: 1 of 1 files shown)


TorreZuk:
Use of TENSILE_USE_MSGPACK would mean it comes after in which case the only addition is static linking to hide the dependency from cascading.   Thoughts on that ?  Could be added in 1057.   Either PR should help rocSOLVER and hipBLAS for current yaml default, but don't you recommend we change to msgpack as default ?   Thus could hopefully hide msgpack which is likely build from source on many platforms (see https://github.com/ROCmSoftwarePlatform/rocBLAS-internal/pull/151 )

Tensile/Source/lib/CMakeLists.txt
```diff
@@ -77,8 +77,10 @@ add_library (TensileHost STATIC ${tensile_sources})
 
 target_include_directories(TensileHost INTERFACE "${CMAKE_CURRENT_SOURCE_DIR}/include")
 
-find_package(msgpack)
-target_link_libraries(TensileHost PUBLIC msgpackc)
+if(Tensile_LIBRARY_FORMAT MATCHES "msgpack")
+    find_package(msgpack)
+    target_link_libraries(TensileHost PRIVATE msgpackc.a)
```
pfultz2, line 82: `msgpackc-static` should be used. Using `msgpackc.a` will not include the usage requirements.

pfultz2: Actually this wont work when building msgpackc with `-DBUILD_SHARED_LIBS=Off`, as msgpack doesn't provide `msgpackc-static`. You need to check for static, like this:

```cmake
if(TARGET msgpackc-static)
    target_link_libraries(TensileHost PRIVATE msgpackc-static)
else()
    target_link_libraries(TensileHost PRIVATE msgpackc)
endif()
```

PR #1059, opened 2020-06-30
Move fp16 code entirely to components

sdquiring:
- Refactoring fp16 code to use format strings instead of % operator
 - Adding Priority component which can raise/lower workgroup priority or be disabled.
 - Head level component selectors can be callable.


Tensile/Code.py
```diff
@@ -563,7 +563,8 @@ def __str__(self):
               C[1] = A[1]*B[1]+D[1]
               """
         else:
-          printExit("Half-precision not supported for arch=%u" % self.version )
+          import pdb; pdb.set_trace()
```
sdquiring, line 566: Remove debug statement.

Tensile/Common.py
```diff
@@ -1339,12 +1350,56 @@ def gfxArch(name):
...
-# can override them, those overridings happen here
 ################################################################################
 def assignGlobalParameters( config ):
+  """
+  Assign Global Parameters
+  Each global parameter has a default parameter, and the user
+  can override them, those overridings happen here
+  """
 
   global globalParameters
```
cgmillette: @ Line 1409 - Message about Benchmark.yaml might be confusing here as we might not always use this function in that context.

Tensile/Common.py
```diff
@@ -1253,21 +1253,29 @@ def locateExe( defaultPath, exeName ): # /opt/rocm/bin, hip-clang
...
+  rv["SupportedISA"]    = tryAssembler(isaVersion, "")
+  rv["HasExplicitCO"]   = tryAssembler(isaVersion, "v_add_co_u32 v0,vcc,v0,1")
+  rv["HasExplicitNC"]   = tryAssembler(isaVersion, "v_add_nc_u32 v0,v0,1")
+
+  rv["HasDirectToLds"]  = tryAssembler(isaVersion, "buffer_load_dword v40, v36, s[24:27], s28 offen offset:0 lds")
+  rv["HasAddLshl"]      = tryAssembler(isaVersion, "v_add_lshl_u32 v47, v36, v34, 0x2")
+  rv["HasSMulHi"]       = tryAssembler(isaVersion, "s_mul_hi_u32 s47, s36, s34")
+  rv["HasCodeObjectV3"] = tryAssembler(isaVersion, "", False, "-mno-code-object-v3")
+
+  rv["HasMFMA"]         = tryAssembler(isaVersion, "v_mfma_f32_32x32x2bf16 a[0:31], v32, v33, a[0:31]")
```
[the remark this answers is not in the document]
sdquiring: Yes, just naming each key as the instruction mnemonic is the preferred way going forward so that we don't have to come up with new names for everything.

PR #1060, opened 2020-07-01
Pipelined store for MI kernels

rosenrodt:
Status: local `pre_checkin` passed. `rocblas-test` integration test gfx900/gfx908 passed

Optimize VGPR consumption for MI kernels in the store path. Allowing opportunity for more parameters combinations (eg MT256x256) or more occupancy if not LDS-bound:
- Removed declaring C-tile vgprs to allow for generating extreme cases such as MT256x256 kernels
- Opportunistically use all VGPRs until the next occupancy level in C-tile store stage. Shown to improve efficiency

Minors:
- Refactored code gen so optimized NoLoadLoop uses the same pipelined store just as the ordinary code path
- All store paths will reuse registers from the register pool for storing C-tiles
- Added more test problem sizes to catch edge under Tensile/Tests/pre_chechin/mfma and remove duplicates

|Kernel|VGPR|VGPR(old)|
|-|-|-|
Cijk_Ailk_Bjlk_HBH_MT128x256x32_MI32x32x4x2_SE_IU1_K1_TT2_64 | 128 | 139
Cijk_Ailk_Bjlk_HBH_MT128x256x32_MI32x32x4x2_SE_IU2_K1_TT2_64 | 128 | 139
Cijk_Ailk_Bjlk_HBH_MT128x256x32_MI32x32x4x2_SE_IU4_K1_TT2_64 | 175 | 175
Cijk_Ailk_Bjlk_HBH_MT256x256x32_MI32x32x4x2_SE_IU1_K1_TT4_64 | 256 | 267
Cijk_Ailk_Bjlk_HBH_MT256x256x32_MI32x32x4x2_SE_IU2_K1_TT4_64 | 256 | 267
Cijk_Ailk_Bjlk_HBH_MT256x256x32_MI32x32x4x2_SE_IU4_K1_TT4_64 | 256 | 267
Cijk_Ailk_Bjlk_HBH_MT64x128x32_MI32x32x4x2_SE_IU1_K1_TT1_32 | 45 | 45
Cijk_Ailk_Bjlk_HBH_MT64x128x32_MI32x32x4x2_SE_IU2_K1_TT1_32 | 61 | 61
Cijk_Ailk_Bjlk_HBH_MT64x128x32_MI32x32x4x2_SE_IU4_K1_TT1_32 | 93 | 93


rosenrodt:
> would be good idea to use C/D vgprs count based on MT and xdlop choice (should be same for all instruction type) and optimize VGPR allocation for local / global reads, offset() and other register usage to fit into C/D register count of xdlops()?
> 
> understand some optimization parameter might require more VGPR(s) than C/D register count. in that case allocate additional VGPR(s) as needed.

Store part requires at the minimum `StoreVectorWidth`'s worth of C/D vgprs. The idea is to perform only so much acc->vgpr copying in order to issue one store instruction. 

For example, if we aim for dwordx4 store, then there will be as few as 4 C/D vgprs that alias the inactive LR/LW/GR/AB vgprs for the "acc->vgpr->store" loop. In practice, the number of C/D vgprs is determined by available spare vgprs in the `RegisterPool` class instance.

I wanted to add that MT256x256 does not actually need 256 C/D vgprs as the table suggests—there's still plenty of vgpr headroom—it merely says 256 acc's are needed. Those numbers however are straight from kernel header (albeit a bit misleading due to acc/vgpr resource granularity). I should update the table to reflect true usage for each of the vgpr/acc resource. The extra vgpr headroom should allow us to explore other optimization opportunities.

> do we allocate additional LDS for storereorder optimization?

That is #993 Store Remapping. Pipelined store does not need extra LDS memory.

rosenrodt:
After rebase, `pre_checkin` on gfx908 passed and `rocblas-test` on gfx900/908 passed. Merging

PR #1071, opened 2020-07-08
Improve serialization performance.

sdquiring:
This is branched from #1057.  Once that is merged, the changes here should be reduced.

This improves the library loading times significantly.  On my test machine, the msgpack version of a recent rocBLAS library loads in under 2.5 seconds, the YAML version loads in around 6.5 seconds.

Changes made to improve performance:

 - Save results of `objectToMap()` to reduce redundant work.
 - Merge language and instruction predicates to reduce total object count.
 - Use unordered versions of containers.
 - Limit consistency checks to debug mode.


Tensile/Source/lib/include/Tensile/msgpack/MessagePack.hpp
```diff
@@ -112,13 +113,19 @@ namespace Tensile
...
+            {
+                error.push_back(msg);
+            }
+
             template <typename T>
             void mapRequired(const char* key, T& obj)
             {
-                auto map = objectToMap(object);
+                if(objectMap.empty())
+                    objectMap = objectToMap(object);
```
TorreZuk, line 127: Could just pass the map as reference second argument?  Would double check the assembly for this assignment

zaliu:
Like @TorreZuk I was unable to use the browser (Chrome) to review the code changes.  In fact the BUI showed the number of Files Changed as infinity.

sdquiring:
This should show correctly now.  There were changes to some large files that were causing problems with GitHub; I've compressed them which should also speed checkouts in the future.

Tensile/Source/lib/include/Tensile/msgpack/MessagePack.hpp
sdquiring: Great suggestion, this is even faster.

PR #1081, opened 2020-07-15
Benchmark tuning parts

wbgilmartin:
updates to tensile to support a modular based approach to tuning using separates steps in the tuning process and provides functions (api's) which enable these steps. 

Step one is the creation of the tensile client.
Step two is the creation of a tensile library which includes the solutions which are used for bench-marking.
Step three runs the actual bench-marks using the which are inputs to the api.

The set of solutions is defined by a configuration. Here I striped down the generation of the solutions based on how we do the tuning in practice, this only includes the common and forked parameters.

The unit and pre_checking tests pass on my local machine.


saadrahim:
@wbgilmartin you are failing some linting tests. There is now a separate static analysis job to show this.

Tensile/Common.py
```diff
@@ -1579,21 +1580,28 @@ def assignParameterRequired(destinationDictionary, key, sourceDictionary):
 # Push / Pop Working Path
 # store a WorkingPath where to write files (like benchmark files)
 ################################################################################
+def ensurePath( path ):
```
sdquiring, line 1583: Moving this function here puts the comment above in the wrong place.

Tensile/ClientWriter.py
```diff
@@ -135,11 +131,57 @@ def main( config ):
 
   return returncode
 
+################################################################################
+# Write New Client Run Script
+################################################################################
+def writeNewClientRunScript(scriptPath, clientParametersPath, clientExePath = None): #, forBenchmark):
```
sdquiring, line 137: I think we should avoid this pattern of creating scripts to run tasks.  We've been trying to avoid adding more code generation where possible.

If we want to be able to rerun the client, it should be exposed via a separate Python script or via a subcommand of the main Tensile script.

Tensile/ClientWriter.py
```diff
@@ -135,11 +131,57 @@ def main( config ):
...
+  runScriptName = writeNewClientRunScript(scriptPath, clientParametersPath, clientBuildDir)
+
+  with ClientExecutionLock():
+    process = subprocess.Popen(runScriptName, cwd=scriptPath)
+    process.communicate()
+
+  if process.returncode:
+    printWarning("ClientWriter Benchmark Process exited with code %u" % process.returncode)
+  return process.returncode
+
```
sdquiring, line 183: We should try to use the [`subprocess.run()`](https://docs.python.org/3/library/subprocess.html#using-the-subprocess-module) call wherever possible.

Returning error codes is also a bad pattern.

Also, where is this function called? I can't find it in these changes.

Tensile/ClientWriter.py
```diff
@@ -513,6 +556,8 @@ def problemSizeParams(solution, problem):
 
     return rv
 
+def problemSizeParams(solution, problem):
+  return problemSizeParamsImpl(solution.problemType, problem)
```
sdquiring, line 560: What's the reason for this?

Tensile/ClientWriter.py
```diff
@@ -551,21 +594,12 @@ def dataInitParams(problemType):
...
             ('init-beta',  dataInitName(initBeta))]
 
-def writeClientConfig(forBenchmark, solutions, problemSizes, stepName, stepBaseDir, newLibrary, codeObjectFiles, tileAwareSelection = False):
-
-    if tileAwareSelection:
-      filename = os.path.join(globalParameters["WorkingPath"], "ClientParameters_Granularity.ini")
-    else:
-      filename = os.path.join(globalParameters["WorkingPath"], "ClientParameters.ini")
-
+def writeClientConfigImpl(problemSizes, problemType, sourceDir, codeObjectFiles, resultsFileName, parametersFilePath):
```
sdquiring, line 597: I like this separation.  I wonder if we could come up with a better name than just adding 'Impl' to the function name?

Tensile/TensileCreateLibrary.py
```diff
@@ -924,6 +922,179 @@ def writeSolutionCall(solutionName, problemType):
...
+    permutation = {}
+    permutation.update(basePermutation)
+    permutation.update(newPermutation)
+    updatedHardcodedParameters.append(permutation)
+  return updatedHardcodedParameters
+
+##############################################################################
+# assigenParameters
+##############################################################################
+def assigenParameters(problemTypeConfig, configBenchmarkCommonParameters, configForkParameters):
```
sdquiring, line 941: 'assign'?

Tensile/TensileCreateLibrary.py
```diff
@@ -1264,53 +1505,48 @@ def TensileCreateLibrary():
...
-      if kernel not in kernels:
-        kernels.append(kernel)
-    solutionHelperKernels = solution.getHelperKernelObjects()
-    for ko in solutionHelperKernels:
-      kname = ko.getKernelName()
-      if kname not in kernelHelperNames:
-        kernelHelperOjbs.append(ko)
-        kernelHelperNames.add(kname)
-
+#<<<<<<< HEAD
```
sdquiring, line 1508: Is this a merge conflict that's unresolved?

Tensile/TensileCreateLibrary.py
```diff
@@ -924,6 +922,179 @@ def writeSolutionCall(solutionName, problemType):
   s += ", stream, numInputEvents, inputEvents, outputEvent )"
   return s
 
+##############################################################################
```
sdquiring, line 925: Why is all this code being added to TensileCreateLibrary?
Looks like this code is being copied from elsewhere - now we have 2 copies of this code to maintain?

Tensile/Common.py
wbgilmartin: Resolved in last commit

Tensile/ClientWriter.py
wbgilmartin: resolved in last commit

Tensile/ClientWriter.py
wbgilmartin: How about writeClientConfigIni?

Tensile/TensileCreateLibrary.py
wbgilmartin: resolved in last commit

Tensile/TensileCreateLibrary.py
wbgilmartin: Resolve in last commit

Tensile/TensileCreateLibrary.py
wbgilmartin: This code provides functionality to separate out the generation of solutions from the full Benchmark process. As part of a conversation we do not in practice use all the bench making steps; this code simplifies the overhead and complexity involved in the full process for use in the modernization by providing a means of solution generation without building a complex object to host the such generation.

wbgilmartin: Yes. I still have not addressed this. My comment was an explanation of the purpose of exposing the Benchmarking Internals.

wbgilmartin: I fixed this  in the latest commits

Tensile/BenchmarkProblems.py
```diff
@@ -43,6 +43,41 @@
 from .SolutionWriter import SolutionWriter
 from .TensileCreateLibrary import writeSolutionsAndKernels, writeCMake, buildObjectFileNames
 
+
+def generateForkedSolutions (problemType, forkedParameters, commonSolutionParameters):
```
sdquiring, line 47: I'd like to see this function replace the loop below (line 203/238 of this file) instead of duplicating that functionality.

Tensile/BenchmarkProblems.py
```diff
@@ -43,6 +43,41 @@
...
+  
+  solutions = []
+  numForked = len(forkedParameters)
+
+  ############################################################################
+  # this creates a set or solutions based on the forked parameters using
+  # a set of common parameters from which to fork from
```
sdquiring, line 54: These kinds of header comments should be in docstrings at the beginning of functions so they are more accessible.

Tensile/BenchmarkProblems.py
wbgilmartin: This is actually something I somehow missed in my refactoring. I look into using the Python generator functions.

I want to minimize the amount of refactoring here, we can do more extensive refactoring in future iterations, as time permits.

Tensile/Common.py
```diff
@@ -1617,14 +1618,22 @@ def pushWorkingPath( foldername ):
...
       os.path.split(globalParameters["WorkingPath"])[0]
+  else:
+    globalParameters["WorkingPath"] = workingDirectoryStack.pop()
 def ensurePath( path ):
   try:
     os.makedirs(path)
   except OSError:
     pass
   return path
+def setWorkingPath( fullPathName ):
```
sdquiring, line 1632: I thought this is how `pushWorkingPath` actually worked.  Reviewing the calls to `pushWorkingPath`, I think this implementation could replace the old one, especially if we set the new directory to `os.path.join(cwd, newPath)` because if newPath is an absolute path it will just return `newPath`.

sdquiring (approved):

PR #1091, opened 2020-07-20
GSU for BF16/FP16 high precision accumulation

TonyYHsieh:
In order to use GSU for BF16/FP16 high precision accumulation,
we need user to provide workspace memory for accumulation.

1. add workspace parameter in ContractionProblem
2. add required workspace method, so user can query
   how much wospace each solution required
3. filter kernel by available worksapce for findBestSolution.
4. GSU kernel has beta kernel, normal kernel and conversion kernel.
   beta kernel: handle kernel part
   normal kernel: accumulate F32 data to workspace
   conversion kernel: convert WS F32 to D FP16/BF16 buffer


Tensile/Source/client/include/DataInitialization.hpp
```diff
@@ -286,6 +286,11 @@ namespace Tensile
                 }
             }
 
+            size_t workspaceSize() const
```
jichangjichang, line 289: suggest revising naming as getWorkspaceSize()

Tensile/KernelWriterAssembly.py
```diff
@@ -8071,7 +8072,8 @@ def computeStoreSrdStart(self, kernel):
         strideC = "StrideC%s"%self.indexChars[i]
         kStr += self.s_mul_u64_u32(sgpr(tmpS0), sgpr(tmpS1), coord, sgpr(strideC), "CScale %s by Stride"%coord)
         #kStr += assert_no_shift_of(tmpS1, log2(self.bpeCexternal), "Need temp")
-        kStr += inst("s_lshl_b64", sgpr(tmpS0,2), sgpr(tmpS0,2), log2(self.bpeCexternal), "scale by bpe")
+        tmpBpe = self.bpr if kernel["_GlobalAccumulation"] else self.bpeCexternal
```
rosenrodt, line 8075: can we use `bpeCinternal` here to denote the size requirement for accumulation?

TonyYHsieh: update. use bpeCinternal to replace self.bpr

Tensile/BenchmarkProblems.py
```diff
@@ -497,10 +502,16 @@ def writeBenchmarkFiles(stepBaseDir, solutions, problemSizes, stepName, filesToC
         kernelNames.add(kName)
     solutionKernelsBetaOnly = solution.getKernelsBetaOnly()
     for kernel in solutionKernelsBetaOnly:
-      kName = Solution.getNameFull(kernel)
+      kName = KernelWriter.getKernelNameBetaOnly(kernel)
       if kName not in kernelNamesBetaOnly:
         kernelsBetaOnly.append(kernel)
         kernelNamesBetaOnly.add(kName)
+    solutionKernelsGlobalAccum = solution.getKernelsGlobalAccum()
+    for kernel in solutionKernelsGlobalAccum:
```
sdquiring, line 510: Could we generalize the helper kernels here instead of creating another special case?  That is, could we group the BetaOnly and GlobalAccum kernels as 'helper' kernels so we don't have to have special logic in so many places?

TonyYHsieh: Hi @sdquiring,

May you describe more detail about generalize kernel??

Purpose of BetaOnly and GlobalAccum is different, so they have different source body.
We need different function or class to generate their source code.

We can put kernel name or instance of Beta/GlobalAccum together, but for kernel generator, we have to use different method to handle it.

Thanks,
Tony.

Tensile/Source/client/include/DataInitialization.hpp
TonyYHsieh: I follow naming in Tesnile

Tensile/BenchmarkProblems.py
TonyYHsieh: @sdquiring,

What do you think about putting generalization to another event???

Tensile/Source/lib/include/Tensile/ContractionProblem.hpp
```diff
@@ -505,6 +506,20 @@ namespace Tensile
             return getOperationDescription();
         }
 
+        static constexpr size_t HPA_GSU_WORKSPACE_SIZE_GRANULARITY = 256;
+
+        void setWorkspaceSize(size_t size)
+        {
+            m_workspaceSize = ~size ? (size / HPA_GSU_WORKSPACE_SIZE_GRANULARITY)
```
sdquiring, line 513: How often is the workspace size going to change?  How often will it change by a value that is <256?

It seems to me it will have a few fairly different values: 0, numeric_limits<size_t>::max(), and D elements * sizeof(float).  It may also later have another larger value if more allocations are done, but the chances that we need to increase the workspace size by some value > 0 and < 256 seems pretty small to me.

This seems to increase complexity at what is not likely to be a meaningful gain.

Tensile/Source/lib/include/Tensile/ContractionProblem.hpp
```diff
@@ -554,6 +569,8 @@ namespace Tensile
         size_t m_allocatedElementsNonBatchA;
         size_t m_allocatedElementsNonBatchB;
 
+        size_t m_workspaceSize = 0;
```
sdquiring, line 572: This new member needs to be included in the hash and comparison functions in https://github.com/ROCmSoftwarePlatform/Tensile/blob/develop/Tensile/Source/lib/include/Tensile/ContractionProblem_Detail.hpp .

TonyYHsieh: hash and comparison function is updated.

Tensile/Source/lib/include/Tensile/ContractionProblem.hpp
TonyYHsieh: remove this patch.

Tensile/Source/client/main.cpp
```diff
@@ -355,6 +355,56 @@ namespace Tensile
             return args;
         }
 
+        size_t getMaxWorkspace(std::shared_ptr<MasterSolutionLibrary<ContractionProblem>>& library,
```
sdquiring, line 358: Does this code create extra output?

I think we could get away with growing the workspace as needed, as we iterate.

TonyYHsieh: GPU buffer allocation is in DataInitialization.
So I do iteration in getMaxWorkspace before running iteration.
Do we have plan to do growing workspace in running iteration??

sdquiring: If `DataInitialization::prepareGPUInputs()` also takes a solution, it could look at the current required workspace and grow it if necessary.

TonyYHsieh: https://github.com/ROCmSoftwarePlatform/Tensile/blob/develop/Tensile/Source/client/include/DataInitializationTyped.hpp#L181
Here shows it only allocate GPU buffer at first call with max buffer size.
Looks like, it didn't re-allocate GPU buffer for each problem which is good behavior for training.

If we really want to put max workspace size calculation in DataInitialization, we can pass solution set in MasterSolutionLibrary as input to DataInitialization and calculate max workspace on constructor of DataInitialization.
Do we need to do that??

PR #1100, opened 2020-07-24
Fix: reusable vgpr up to next occupancy threshold

rosenrodt:
Fixes reusable vgpr count up until the next occupancy threshold. Especially the lesser seen solutions where workgroup size > 256 threads.

Unit tests have also been added for the functions that have been fixed.

### DIFF
--- a/Tensile/KernelWriterAssembly.py
+++ b/Tensile/KernelWriterAssembly.py
@@ -395,7 +395,29 @@ class ZeroPadReg:
   def isMatch(self, perp, sPerp, para, sPara):
     return self.perp==perp and self.sPerp==sPerp and self.para==para and self.sPara==sPara
 
+class VgprOccupancyCurve:
+  def __init__(self):
+    self._vgprOccupancy = [0]*(256+1)
+    for i in range(0,   24+1): self._vgprOccupancy[i] = 10
+    for i in range(25,  28+1): self._vgprOccupancy[i] = 9
+    for i in range(29,  32+1): self._vgprOccupancy[i] = 8
+    for i in range(33,  36+1): self._vgprOccupancy[i] = 7
+    for i in range(37,  40+1): self._vgprOccupancy[i] = 6
+    for i in range(41,  48+1): self._vgprOccupancy[i] = 5
+    for i in range(49,  64+1): self._vgprOccupancy[i] = 4
+    for i in range(65,  84+1): self._vgprOccupancy[i] = 3
+    for i in range(85, 128+1): self._vgprOccupancy[i] = 2
+    for i in range(129,256+1): self._vgprOccupancy[i] = 1
+  def __call__(self, index):
+    if index < 0 or index > 256:
+      return 0
+    return self._vgprOccupancy[index]
+  def __getitem__(self, item):
+    return self(item)
+  def __len__(self):
+    return len(self._vgprOccupancy)
 
+vgprOccupancy = VgprOccupancyCurve()
 
 ################################################################################
 # Assembly Kernel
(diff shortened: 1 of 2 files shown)


Tensile/KernelWriterAssembly.py
```diff
@@ -579,36 +589,35 @@ def getLinkCodeObjectArgs(self, objectFileNames, coFileName, *moreArgs):
...
+    initOccupancy = vgprOccupancy[vgprs]
     if accvgprs > 0:
-      initOccupancy = min(initOccupancy, accvgprOccupancy)
+      initOccupancy = min(initOccupancy, vgprOccupancy[accvgprs])
     if ldsSize > 0:
-      initOccupancy = min(initOccupancy, self.getLdsLimitedOccupancy(ldsSize))
+      initOccupancy = min(initOccupancy, KernelWriterAssembly.getLdsLimitedOccupancy(ldsSize))
     lastVgprs = vgprs
-    while vgprs < len(self.vgprOccupancy)-1:
+    while vgprs < len(vgprOccupancy)-1:
```
imcarsonliao, line 618: using if condition or a static array should be able to get the largest vgpr within the same occupancy criteria rather than loop search.

Tensile/KernelWriterAssembly.py
```diff
@@ -579,36 +589,35 @@ def getLinkCodeObjectArgs(self, objectFileNames, coFileName, *moreArgs):
...
-
-    vgprOccupancy = self.vgprOccupancy[vgprs] if vgprs <= 256 else 0
-    accvgprOccupancy = self.vgprOccupancy[accvgprs] if accvgprs <= 256 else 0
-    initOccupancy = vgprOccupancy
+  @staticmethod
+  def getMaxRegsForOccupancy(vgprs, ldsSize, accvgprs=0):
+    initOccupancy = vgprOccupancy[vgprs]
     if accvgprs > 0:
-      initOccupancy = min(initOccupancy, accvgprOccupancy)
+      initOccupancy = min(initOccupancy, vgprOccupancy[accvgprs])
```
imcarsonliao, line 614: no need to check this if condition

rosenrodt: addressed in latest commit

Tensile/KernelWriterAssembly.py
rosenrodt: @imcarsonliao I would prefer the vgpr#-occupancy curve to be the single source of truth. I simply want to avoid defining another lookup table for this specific use case.

Tensile/Tests/unit/test_KernelWriterAsssembly.py
```diff
@@ -0,0 +1,30 @@
...
+# FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+# COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+# IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNE-
+# CTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+################################################################################
+
+from Tensile.KernelWriterAssembly import KernelWriterAssembly
+
+def test_max_regs():
+    assert KernelWriterAssembly.getMaxRegsForOccupancy(10,65536) == 256
```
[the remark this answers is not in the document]
rosenrodt: addressed in latest commit

PR #1104, opened 2020-07-27
Modify reject test to avoid false rejection

solaslin:
Update some rejection test code which causes false rejection:
False rejection means:
- Reject a solution which already has the final derived parameters, and should not be rejected.
- Happens on some auto-assigned parameters (Value = -1)
- Happens when a solution is re-assignDerivedParams() : 
  - Which means the rejection test is correct for the first time, but incorrect if the rejection is tested again.

### DIFF
--- a/Tensile/SolutionStructs.py
+++ b/Tensile/SolutionStructs.py
@@ -2714,20 +2714,23 @@ class Solution:
       return
 
     # Default LocalReadVectorWidth
+    supportWiderLR = state["TransposeLDS"] and \
+                     state["ProblemType"]["TLUA"] == False and \
+                     state["ProblemType"]["TLUB"] == False
     if state["LocalReadVectorWidth"] == -1:
       if state["EnableMatrixInstruction"]:
         state["LocalReadVectorWidth"] = state["ProblemType"]["DataType"].numMIInput()
       else:
         state["LocalReadVectorWidth"] = state["VectorWidth"]
-    else:
-      if not state["EnableMatrixInstruction"]:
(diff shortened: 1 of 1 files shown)


Tensile/SolutionStructs.py
```diff
@@ -2714,20 +2714,21 @@ def assignDerivedParameters(state):
...
-      if state["ProblemType"]["TLUA"] or state["ProblemType"]["TLUB"]:
-        reject(state, "LocalReadVectorWidth requires TN TransposeLDS=1")
       if state["LocalReadVectorWidth"] < state["ProblemType"]["DataType"].numMIInput():
         reject(state, "LocalReadVectorWidth < %u" %(state["ProblemType"]["DataType"].numMIInput()))
+    else:
+      if state["LocalReadVectorWidth"] == -1:
+        state["LocalReadVectorWidth"] = state["VectorWidth"]
+      if state["LocalReadVectorWidth"] != state["VectorWidth"]:
+        reject(state, "LocalReadVectorWidth %u Must == VectorWidth %u when not (EnableMatrixInstruction && TransposeLDS=1 && TN)" \
+          %(state["LocalReadVectorWidth"], state["VectorWidth"]))
```
solaslin, line 2732: Can use test case " LRVW= -1 / EnableMatrixInst=False"
- Original:
  - 1st time (Train): LRVW = VectorWidth, done
  - 2nd time (Lib):  LRVW != -1, but EnableMatrixInst = False, reject

solaslin:
Fixed and approved

zaliu:
> Fixed and approved

Indeed.  Please squash-and-merge.  Thanks!

PR #1109, opened 2020-07-29
Temporarily disable tests that don't build on hipcc

AlexBrownAMD:


AlexBrownAMD:
> Are these extended tests?

Yes, 2 don't build, 1 fails (but passes on hcc)

PR #1110, opened 2020-07-29
Don't generate legacy components for new client builds

cgmillette:
- Previously caused issues when the MERGE_FILES is OFF because it tries to create very long solution source file names. These legacy components should be skipped when using TensileCreateLibraryFiles in cmake as it is designed for the new client only


amdkila:
Hmmm, looks like this change has broke our CI. I think you just have to pass this --no-legacy-components to the hip/CMakeLists.txt file as well

PR #1116, opened 2020-08-05
New BoundsCheck design: guard page

jichangjichang:
[Background]
Tensile and Rocblas-test memory allocation behavior are different with real framework use cases. It would allocate a huge memory for max problems size one time.  And ABCD matrix could be allocated continuously. Not easy to trigger memory fault when out-of-boundary access.  Better to have a way to detect all out of boundary cases for each size and each problems.

[New Design purpose]
Let MMU detect out-of-boundary access automatically for each problems. Even if single byte access before Matrix or after Matrix also can be detected for all problems.
BoundsCheck = 2 : Front side guard page
BoundsCheck = 3 : Back side guard page
BoundsCheck = 4 : both Front side and Back side guard page. Needs 2 benchmark runs per solution. (Recommended)

[How This new Design work]
Make sure all matrix allocated page aligned memory. (2MB by default)
Let the memory addressing between different matrix are un-allocated. (free guard page)
Detect front side: Put the Matrix start address to allocated memory start address.
Detect back side: Put the Matrix end address to allocated memory end address. 
Any single byte overrun would trigger GPU memory segmentation fault.

### DIFF
--- a/HostLibraryTests/client/DataInitialization_test.cpp
+++ b/HostLibraryTests/client/DataInitialization_test.cpp
@@ -79,7 +79,7 @@ public:
         return rv;
     }
 
-    void RunDataContaminationTest(bool cEqualD, bool pristineGPU, bool boundsCheck)
+    void RunDataContaminationTest(bool cEqualD, bool pristineGPU, int boundsCheck)
     {
         using val = po::variable_value;
 
--- a/HostLibraryTests/client/DataInitialization_test.cpp
+++ b/HostLibraryTests/client/DataInitialization_test.cpp
@@ -79,7 +79,7 @@ public:
         return rv;
     }
 
-    void RunDataContaminationTest(bool cEqualD, bool pristineGPU, int boundsCheck)
+    void RunDataContaminationTest(bool cEqualD, bool pristineGPU, BoundsCheckMode boundsCheck)
     {
         using val = po::variable_value;
 
@@ -161,42 +161,42 @@ TYPED_TEST_SUITE(DataInitializationTest, InputTypes);
 
 TYPED_TEST(DataInitializationTest, Contamination_false_false_false)
 {
-    this->RunDataContaminationTest(false, false, false);
+    this->RunDataContaminationTest(false, false, BoundsCheckMode::Disable);
 }
 
 TYPED_TEST(DataInitializationTest, Contamination_false_true_false)
 {
-    this->RunDataContaminationTest(false, true, false);
+    this->RunDataContaminationTest(false, true, BoundsCheckMode::Disable);
 }
 
 TYPED_TEST(DataInitializationTest, Contamination_true_false_false)
 {
-    this->RunDataContaminationTest(true, false, false);
+    this->RunDataContaminationTest(true, false, BoundsCheckMode::Disable);
 }
 
 TYPED_TEST(DataInitializationTest, Contamination_true_true_false)
 {
-    this->RunDataContaminationTest(true, true, false);
+    this->RunDataContaminationTest(true, true, BoundsCheckMode::Disable);
 }
 
 TYPED_TEST(DataInitializationTest, Contamination_false_false_true)
 {
-    this->RunDataContaminationTest(false, false, true);
+    this->RunDataContaminationTest(false, false, BoundsCheckMode::NaN);
 }
 
 TYPED_TEST(DataInitializationTest, Contamination_false_true_true)
 {
-    this->RunDataContaminationTest(false, true, true);
+    this->RunDataContaminationTest(false, true, BoundsCheckMode::NaN);
 }
 
 TYPED_TEST(DataInitializationTest, Contamination_true_false_true)
 {
-    this->RunDataContaminationTest(true, false, true);
+    this->RunDataContaminationTest(true, false, BoundsCheckMode::NaN);
 }
 
 TYPED_TEST(DataInitializationTest, Contamination_true_true_true)
 {
-    this->RunDataContaminationTest(true, true, true);
+    this->RunDataContaminationTest(true, true, BoundsCheckMode::NaN);
 }
 
 template <typename T>
--- a/Tensile/Source/client/include/DataInitialization.hpp
+++ b/Tensile/Source/client/include/DataInitialization.hpp
@@ -57,6 +57,16 @@ namespace Tensile
             Count
         };
 
+        enum class BoundCheckMode
+        {
+            Disable = 0,
+            NaN,
+            GuardPageFront,
+            GuardPageEnd
+        };
+
+        const int pageSize = 2 * 1024 * 1024;
+
         static bool IsProblemDependent(InitMode const& mode)
         {
             return mode == InitMode::SerialIdx || mode == InitMode::SerialDim0
@@ -67,6 +77,11 @@ namespace Tensile
 
         std::ostream& operator<<(std::ostream& stream, InitMode const& mode);
         std::istream& operator>>(std::istream& stream, InitMode& mode);
+
+        inline bool operator==(const int& lhs, const BoundCheckMode& rhs)
+        {
+            return lhs == static_cast<int>(rhs);
+        }
 
         template <typename TypedInputs>
         class TypedDataInitialization;
@@ -320,7 +335,7 @@ namespace Tensile
             /// If true, we will initialize all out-of-bounds inputs to NaN, and
             /// all out-of-bounds outputs to a known value. This allows us to
             /// verify that out-of-bounds values are not used or written to.
-            bool m_boundsCheck = false;
+            int m_boundsCheck = 0;
 
             /// If true, the data is dependent on the problem size (e.g. serial)
             /// and must be reinitialized for each problem. Pristine copy on GPU
--- a/Tensile/Source/client/include/DataInitialization.hpp
+++ b/Tensile/Source/client/include/DataInitialization.hpp
@@ -57,16 +57,6 @@ namespace Tensile
             Count
         };
 
-        enum class BoundCheckMode
-        {
-            Disable = 0,
-            NaN,
-            GuardPageFront,
-            GuardPageEnd
-        };
-
-        const int pageSize = 2 * 1024 * 1024;
-
         static bool IsProblemDependent(InitMode const& mode)
         {
             return mode == InitMode::SerialIdx || mode == InitMode::SerialDim0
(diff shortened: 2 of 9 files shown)


Tensile/Source/lib/include/Tensile/ContractionSolution.hpp
```diff
@@ -34,6 +34,10 @@
 #include <Tensile/DataTypes.hpp>
 #include <Tensile/Predicates.hpp>
 
+#include <boost/program_options.hpp>
```
sdquiring, line 37: Boost cannot be used inside the lib directory as that imposes boost as a dependency on all users.

Tensile/Source/lib/include/Tensile/ContractionSolution.hpp
```diff
@@ -148,18 +152,22 @@ namespace Tensile
         /**
    * Generate a set of kernel calls to solve a particular problem.
    */
-        virtual std::vector<KernelInvocation>
-            solve(Problem const& problem, Inputs const& inputs, Hardware const& hardware) const;
+        virtual std::vector<KernelInvocation> solve(Problem const&           problem,
```
sdquiring, line 155: Adding arguments here needs to be very carefully considered, as this is part of the main interface to Tensile.  We can't add any 'debug' type arguments here.

Tensile/Source/lib/include/Tensile/ContractionProblem.hpp
```diff
@@ -588,8 +588,16 @@ namespace Tensile
...
         TypedContractionInputs();
-        TypedContractionInputs(
-            A const* _a, B const* _b, C const* _c, D* _d, Alpha _alpha, Beta _beta);
+        TypedContractionInputs(A const* _a,
+                               B const* _b,
+                               C const* _c,
+                               D*       _d,
+                               Alpha    _alpha,
+                               Beta     _beta,
+                               size_t   _aElements,
```
sdquiring, line 597: Adding these arguments is a fundamental change to the interface to Tensile.  We can't do this just for internal debugging purposes.

Tensile/Source/lib/include/Tensile/Utils.hpp
```diff
@@ -34,6 +34,8 @@
 #include <sstream>
 #include <type_traits>
 
+#define PAGE_SIZE (2097152)
```
sdquiring, line 37: We have been very careful not to add any symbols outside the Tensile namespace.  This can't be a macro.

sdquiring:
Any functionality which will cause kernels to be called more than what the user has requested needs to be **outside** of the lib directory.

jichangjichang:
> Any functionality which will cause kernels to be called more than what the user has requested needs to be **outside** of the lib directory.

@sdquiring  I remove all changes of lib directory. Now the debug codes are only in client codes.
Please check if this ok for you.

Tensile/Source/client/include/DataInitialization.hpp
```diff
@@ -57,6 +57,16 @@ namespace Tensile
             Count
         };
 
+        enum class BoundCheckMode
```
sdquiring, line 60: For consistency, this should be `BoundsCheckMode`

Tensile/Source/client/include/DataInitialization.hpp
```diff
@@ -57,6 +57,16 @@ namespace Tensile
...
+        {
+            Disable = 0,
+            NaN,
+            GuardPageFront,
+            GuardPageEnd
+        };
+
+        const int pageSize = 2 * 1024 * 1024;
```
sdquiring, line 68: Are we sure the page size is always exactly 2MB?

Is there a way to query for the page size?

Tensile/Source/client/include/DataInitialization.hpp
```diff
@@ -320,7 +335,7 @@ namespace Tensile
             /// If true, we will initialize all out-of-bounds inputs to NaN, and
             /// all out-of-bounds outputs to a known value. This allows us to
             /// verify that out-of-bounds values are not used or written to.
```
sdquiring, line 337: This comment needs to be updated.

Tensile/Source/client/include/ReferenceValidator.hpp
```diff
@@ -133,7 +133,7 @@ namespace Tensile
             bool m_printValids;
             int  m_printMax;
             int  m_validationStride;
-            bool m_boundsCheck;
+            int  m_boundsCheck;
```
sdquiring, line 136: Should be of type `BoundsCheckMode`.

HostLibraryTests/client/DataInitialization_test.cpp
```diff
@@ -79,7 +79,7 @@ class DataInitializationTest : public ::testing::Test
         return rv;
     }
 
-    void RunDataContaminationTest(bool cEqualD, bool pristineGPU, bool boundsCheck)
+    void RunDataContaminationTest(bool cEqualD, bool pristineGPU, int boundsCheck)
```
sdquiring, line 82: Should be of type `BoundsCheckMode`

Tensile/Source/client/main.cpp
```diff
@@ -528,19 +530,18 @@ int main(int argc, const char* argv[])
                             auto kernels = solution->solve(problem, *inputs, *hardware);
 
                             size_t       warmupInvocations = listeners.numWarmupRuns();
```
sdquiring, line 532: If ReferenceValidator::numWarmupRuns returns 2, we could run both the Front and Back modes in one process.

Tensile/Source/client/include/DataInitialization.hpp
```diff
@@ -57,6 +57,16 @@ namespace Tensile
             Count
         };
 
+        enum class BoundCheckMode
+        {
+            Disable = 0,
+            NaN,
+            GuardPageFront,
+            GuardPageEnd
```
sdquiring, line 65: Should be either `Front`/`Back` or `Begin`/`End`.

Tensile/Source/client/include/DataInitialization.hpp
jichangjichang: @sdquiring done

Tensile/Source/client/include/DataInitialization.hpp
jichangjichang: @sdquiring No way to query the actual page size. Or I can add PageSize globalParameters config to let user set the pagesize? This may be better than hard-code design. Is that ok for you?

Tensile/Source/client/include/DataInitialization.hpp
jichangjichang: @sdquiring comments updated

Tensile/Source/client/include/ReferenceValidator.hpp
jichangjichang: @sdquiring done

HostLibraryTests/client/DataInitialization_test.cpp
jichangjichang: @sdquiring done

Tensile/Source/client/include/DataInitialization.hpp
jichangjichang: @sdquiring done

zaliu:
@amdkila was the clang-format issue in this PR fixed by you in another PR?  If so, perhaps a (normally not needed) merge from develop will get the StaticAnalysis portion cleared?  Thanks.

amdkila:
@zaliu No, these format failures are in three files that have been modified in this PR (ReferenceValidator, and two DataInitialization files). I didn't change the githooks script in develop, so a merge of develop into this PR won't fix the issue.

zaliu (dismissed):
Please clear all clang-format issues.  Thanks.

jichangjichang:
> Please clear all clang-format issues. Thanks.
Done

Tensile/Source/client/main.cpp
jichangjichang: @sdquiring Please check PR's third commit. 
I add a new mode "GuardPageAll". When config yaml set boundsCheck as "GuardPageAll", clientWriter would set numWarmups as 2. In new client main function, first run of warmups would go "GuardPageFront" checking and second run would go "GuardPageBack" checking.

Tensile/Source/client/include/DataInitialization.hpp
sdquiring: No, I think this is fine.  If we need to have this configurable I think it should be through the `Hardware` class.

Tensile/Source/client/main.cpp
sdquiring: This new change breaks the encapsulation of the DataInitialization class.  It should be possible to implement this in the DataInitialization class.  If you need more info to be sent in (warmup number) then do that.

Tensile/Source/client/main.cpp
```diff
@@ -537,11 +547,38 @@ int main(int argc, const char* argv[])
                             for(int i = 0; i < warmupInvocations; i++)
                             {
                                 listeners.preWarmup();
-                                if(gpuTimer)
-                                    adapter.launchKernels(
-                                        kernels, stream, warmupStartEvents[i], warmupStopEvents[i]);
+
+                                if(i == 1 && boundsCheck == BoundsCheckMode::GuardPageAll)
```
sdquiring, line 551: This breaks the encapsulation of the DataInitialization class.  If you need to send `i` into the prepareGPUInputs function, then do that.

sdquiring (changes requested):

Tensile/Source/client/main.cpp
jichangjichang: @sdquiring If we put i into prepareGPU, every kernel launch need to re-call prepareGPUInputs() and solve() one time. launch index "i" is meaningless for DataInitialization class.
For "GuardPageAll", it needs construct 2 sets DataInitialization. Different "launch kernel" needs to call different dataInit. It would break encapsulation of the DataInitialization class no matter we put the implementation inside or outside of DataInitialization class.
I would like to discard my third commit "GuardPageAll" and just keep the implementation "GuardPageFront" and "GuardPageBack".
Is that ok?

sdquiring: If we can validate front and back in one run, this feature is much more likely to be used.

What if this mode caused there to be two iterations of the `while(listeners.needMoreRunsInSolution())` loop?  This could be controlled from the ReferenceValidator, or we could make the DataInitialization class be a listener.

jichangjichang:
@sdquiring As your suggestion, I make the DataInitialization class be a listener. If ::GuardPageAll is set, needMoreRunsInSolution() of DataInitialization class would trigger 2 runs for front side checking and back side checking. main() of new client now only needs 1 line change. 
Please help review current design. Thanks.

sdquiring (approved):
Great!

PR #1122, opened 2020-08-07
Print kernel name

JeremyAdamHart:
Added a TENSILE_DB=0x4000 flag which prints the kernel name when the solve function is called.

### DIFF
--- a/Tensile/Source/lib/source/Debug.cpp
+++ b/Tensile/Source/lib/source/Debug.cpp
@@ -47,11 +47,6 @@ namespace Tensile
         return m_value & 0x10;
     }
 
-    bool Debug::printLookupEfficiency() const
-    {
-        return m_value & 0x4000;
-    }
-
     bool Debug::printCodeObjectInfo() const
     {
         return m_value & 0x20;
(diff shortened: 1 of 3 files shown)


Tensile/Source/lib/source/Debug.cpp
```diff
@@ -95,6 +95,11 @@ namespace Tensile
         return m_value & 0x2000;
     }
 
+    bool Debug::printWinningKernelName() const
+    {
+        return m_value & 0x4000;
```
rosenrodt, line 100: does the debug bit for `printWinningKernelName()` intentionally collide with `printLookupEfficiency()`? They both AND on the same mask 0x4000

JeremyAdamHart: Thanks for catching that, the rest were set up in increasing order and I missed that one. I'll change it to 0x6000 and maybe move printLookupEfficiency just before it to keep the ordering

JeremyAdamHart: That last comment should have said 0x8000, I've applied that change in the newest commit

PR #1123, opened 2020-08-07
Short program for timing hipModuleLoad function

JeremyAdamHart:
Compares the execution time of hipModuleLoad to the time required to the time to load the file into memory


JeremyAdamHart:
Intended to be run on the TensileLibrary .co files, and .hsaco files, which can be found in the Tensile/library directory inside a rocBLAS build

PR #1125, opened 2020-08-10
PersistentKernel with Prefetch

solaslin:
* Persistent Kernel for MI, support StoreRemap, FP16/BF16 HPA, WiderLocalRead.
* PersistentKernelAlongBatch: Persistent Kernel WG can also distributed along batch dimension
* Implement PrefetchAcrossPersistent


Tensile/KernelWriterAssembly.py
```diff
@@ -11124,9 +11231,15 @@ def closePrefetchAcrossPersistent(self, kernel):
   def functionEnd(self, kernel, addLabel=True):
     imod = Code.Module()
     if kernel["PersistentKernel"]:
+
+      if kernel["StoreRemapVectorWidth"]:
+        imod.addInst("s_barrier", "")
```
rosenrodt, line 11236: Would it be better to put the barrier immediately before the local writes instructions in the next loop? So that you can always have buffer loads (for the next loop) sooner before hitting the barrier

solaslin: Yes, thanks for the review. I've changed this, now the barrier located after ShadowInit and before the first ds_write

Tensile/KernelWriter.py
```diff
@@ -945,34 +959,42 @@ def kernelBody( self, kernel, tensorParametersA, tensorParametersB ):
...
-      if self.enable["Wait"]:
-        kl.append(self.wait(kernel, tensorParametersA, tensorParametersB, 0, -1, -1, "8wait for global read"))
-
-        # These cases loop back and run the prefetch loop again
-        # we need an extra barrier to ensure that the ds_reads from previous iteration
-        # have finished before we generate the prefetch for the next summation index.
-        if kernel["PrefetchAcrossPersistent"] or self.actualSummationLoops>1:
-          kl.append(self.syncThreads(kernel))
+      else: # no doShadowInit
+        if self.enable["Wait"]:
```
imcarsonliao, line 994: change the behavior of non-pk and would shift wait global load earlier

Tensile/KernelWriter.py
```diff
@@ -945,34 +959,42 @@ def kernelBody( self, kernel, tensorParametersA, tensorParametersB ):
         # can't do shadow initC with multiple summation since this resets the ValuC counters
         # on each unroll iteration.
         self.doShadowInit = 1 # 1 is just store setup
+
+    # in HPA mode, can do the alpha, beta conversion f16->f32 only once
+    kl.append( self.checkAlphaBetaForHPA(kernel))
```
imcarsonliao, line 964: any reason to move this code block earlier

Tensile/KernelWriter.py
imcarsonliao, line 994: change the behavior of non-pk and would shift wait global load earlier

Tensile/KernelWriter.py
solaslin: This is for HPA issue, originally, the conversion from f16 to f32 is done in each tile loop, and stores the f32 value back to alpha / beta everytime, if we keep this conversion inside the pk-loop, then we need to test if this is the first time doing conversion. (if not the first time, then skip the conversion). 

And also, in the original version, this conversion code is scattered all over the different case (OptNNL, non-OptNNL, SRVW, Non-SRVW) which makes the maintenance difficult. 

Put this conversion before the PK-loop so that we can do it only once.

solaslin: Done, I've made the change, now this is done earlier only when we have PK

Tensile/KernelWriter.py
solaslin: Yes, I found this is redundant modification. Fixed.

solaslin:
WIP, need to rebase with latest PR. And also need to resolve the HostLibraryTests data conflict.

Tensile/KernelWriterAssembly.py
```diff
@@ -5842,6 +5927,14 @@ def openSumAtLeastUnroll(self, kernel, prefetch, isPap, isOptNLL):
       kStr += inst("s_cbranch_scc0 %s"%skipOptNLL, \
           "skip if tail loop required")
 
+      # The prefetch across persistent for OptNLL case
+      if self.prefetchAcrossPersistent:
+        kStr += str(self.openPrefetchAcrossPersistent(kernel, isOptNLL=True))
+        newTileCodes = self.setupNewTile(kernel, self.tPA, self.tPB, isPap=True, isOptNLL=True)
```
ramjana, line 5933: Offset Ptrs for new next PK tile is same as current one. Can we define  another set of SRDA,SRDB for next tile and do all SRDs code after we issue prefetch for the current tile in setupNewTile code?? 

noloadLoop() should be same as unroll loop with global loads & lwr in unroll iteration is for next Tile. 

may be that's next optimization for PK

solaslin:
@ramjana thanks for reviewing. I'll merge this version first and keep improving this as your suggestion. But it's about the code freeze date so I'll do it at the beginning of next release.

PR #1132, opened 2020-08-17
BF16/FP16 HPA GSU with per GSU temp buffer

TonyYHsieh:
1. write each GSU output to independent temp buffer
    2. use reduce kernel to accumulate all GSU output
    3. stride of temp buffer is equal to size of temp buffer
    4. support SRVW for BF16/FP16 HPA GSU


Tensile/SolutionStructs.py
```diff
@@ -2871,7 +2879,7 @@ def assignDerivedParameters(state):
       if state["PersistentKernel"]:
         reject(state, "storeRemap doesn't support persist kernel yet")
         return
-      if state["GlobalSplitU"] > 1:
+      if (state["GlobalSplitU"] > 1) and (state["_GlobalAccumulation"] != 2):
```
imcarsonliao, line 2882: should modify reject reason

Tensile/KernelWriterAssembly.py
```diff
@@ -8459,7 +8472,7 @@ def cleanupGlobalWrite(self, kernel):
   ##############################################################################
   # Return max global write vector width, in elements
   def maxGwvw(self, kernel):
-    atomic = kernel["GlobalSplitU"] > 1
+    atomic = (kernel["GlobalSplitU"] > 1) and (kernel["_GlobalAccumulation"] != 2)
```
imcarsonliao, line 8475: would it be better to specifically choose _GlobalAccumulation ==1 in case GSUalgo is growing? what can be considered is put atomic condition to a reusable function as some other places use this as well

Tensile/Tests/pre_checkin/mfma/hpa_hgemm_asm.yaml
```diff
@@ -428,3 +515,32 @@ BenchmarkProblems:
...
+          - [ 2, 32 ]
+        - WorkGroup:
+          - [  32, 8, 1 ]
+          - [  64, 4, 1 ]
+        - WorkGroupMapping: [8]
+        - GlobalSplitU: [1]
+        - InnerUnroll: [2]
+        - DepthU: [32]
+        - VectorWidth: [2]
+        - GlobalSplitUAlogrithm: [1, 2]
```
imcarsonliao, line 539: typo

Tensile/Common.py
```diff
@@ -323,6 +323,11 @@
     # GSU applies only to the unroll summation dimension
     "GlobalSplitU":               list(range(1, 1024+1)),
 
+    # choose how to do GlobalSplitU
+    # 1: use atomic operation to accumlate on one buffer
+    # 2: each GSU group write to each own buffer and accumulate by another kernel
+    "GlobalSplitUAlogrithm":      [1, 2],
```
imcarsonliao, line 329: typo

Tensile/KernelWriterAssembly.py
TonyYHsieh: 0 and 1 need to use atomic

Tensile/SolutionStructs.py
TonyYHsieh: update

Tensile/Tests/pre_checkin/mfma/hpa_hgemm_asm.yaml
```diff
@@ -428,3 +515,32 @@ BenchmarkProblems:
...
+          - [32, 32, 8, 1]
+        - PrefetchLocalRead: [1]
+        - PrefetchGlobalRead: [True]
+        - ThreadTile:
+          - [ 2, 32 ]
+        - WorkGroup:
+          - [  32, 8, 1 ]
+          - [  64, 4, 1 ]
+        - WorkGroupMapping: [8]
+        - GlobalSplitU: [1]
```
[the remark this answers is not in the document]
TonyYHsieh: update

Tensile/Tests/pre_checkin/mfma/hpa_hgemm_asm.yaml
TonyYHsieh: update

Tensile/Common.py
TonyYHsieh: update

imcarsonliao (approved):

TonyYHsieh:
@sdquiring  @ramjana 
May you help to review this change??

Here is performance compare. Tab new apply this PR.
image

imcarsonliao:
@AlexBrownAMD please help review as well.

Tensile/Source/lib/source/ContractionSolution.cpp
```diff
@@ -610,15 +642,14 @@ namespace Tensile
...
-        rv.args    = KernelArguments(debug);
+        rv.args = KernelArguments(T_Debug);
 
         rv.args.reserve(512, 64);
 
         rv.kernelName = outputConversionKernelName(problem, inputs, hardware);
 
-        rv.workGroupSize.x = 8;
-        rv.workGroupSize.y = 8;
+        rv.workGroupSize.x = 256;
```
AlexBrownAMD, line 651: Total workgroup size and number of work groups changes here - not sure if there are any side effects of this.

AlexBrownAMD (approved):
LGTM with a couple notes/questions. This could also probably use an extra review from someone.

Tensile/Source/lib/source/ContractionSolution.cpp
TonyYHsieh: We modify kernel as well.
Using 1-D work makes global read in 1-D that can improves performance by avoid bank conflict.

Tensile/SolutionWriter.py
```diff
@@ -558,7 +576,13 @@ def getProblemSourceString(self, problemType, solution, kernelsWithBuildErrs):
...
           s += "%sdataC,\n" % (t)
           # strides
-          for i in range(0,numStridesC*2):
+          if kernel["_GlobalAccumulation"]:
+            for i in range(0, numStridesC):
+              s += "%s%s,\n" % (t, WSstrides[i])
+          else:
+            for i in range(0, numStridesC):
+              s += "%s%s,\n" % (t, self.strideList[i])
+          for i in range(numStridesC, numStridesC*2):
```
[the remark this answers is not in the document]
TonyYHsieh: optimize BetaOnly kernel for GSU algorithm.
This change also remove duplicate code in here.

HostLibraryTests/sample_library.yaml
```diff
@@ -8,7 +8,7 @@
 solutions:
   - name:            'Cijk_Ailk_Bljk_SB_MT128x128x08_K1'
     sizeMapping:
-      globalAccumulation: false
+      globalAccumulation: 0
```
cgmillette, line 11: I'm not opposed to using numerical IDs to distinguish an algorithm, however I might suggest updating the implementation to use an enum wrapper class for the ID.
To a maintainer, globalAccumulation: GlobalAccumulateAlgo.MyAlgo may be a bit more verbose, but is more explicit and refactor-safe than globalAccumulation: 0, where meaning of 0 must be looked up, and potentially meaning of 0 could change in the future.

Tensile/Common.py
```diff
@@ -347,6 +347,11 @@
     # GSU applies only to the unroll summation dimension
     "GlobalSplitU":               list(range(1, 1024+1)),
 
+    # choose how to do GlobalSplitU
+    # 1: use atomic operation to accumlate on one buffer
+    # 2: each GSU group write to each own buffer and accumulate by another kernel
+    "GlobalSplitUAlgorithm":      [1, 2],
```
cgmillette, line 353: +0 = no GSU?

Tensile/Contractions.py
```diff
@@ -316,8 +316,9 @@ def CompoundPredicates(cls, state, problemType):
         if "KernelLanguage" in state:
             rv += [cls("KernelLanguageCompatible", value=state["KernelLanguage"])]
 
-        if 'GlobalSplitU' in state and state['GlobalSplitU'] > 1:
-            rv += [cls("DeterministicMode", value = False)]
+        if ('GlobalSplitU' in state) and (state['GlobalSplitU'] > 1):
```
cgmillette, line 319: Why use _GlobalAccumulation instead of GlobalSplitUAlgorithm?

Tensile/Contractions.py
```diff
@@ -316,8 +316,9 @@ def CompoundPredicates(cls, state, problemType):
...
+            if ('_GlobalAccumulation' not in state) or (state['_GlobalAccumulation'] != 2):
+                rv += [cls("DeterministicMode", value = False)]
```
cgmillette, line 321: Does the _GlobalAccumulation 2 algo use atomics ? If yes then DeterministicMode predicate for 'false' should be included.

Tensile/Common.py
cgmillette: I'm not opposed to using numerical IDs to distinguish an algorithm, however I might suggest updating the implementation to use an enum wrapper class for the ID.
To a maintainer, globalAccumulation: GlobalAccumulateAlgo.MyAlgo may be a bit more verbose, but is more explicit and refactor-safe than globalAccumulation: 0, where meaning of 0 must be looked up, and potentially meaning of 0 could change in the future.

Tensile/KernelWriterAssembly.py
```diff
@@ -1113,28 +1113,27 @@ def initKernel(self, kernel, tPA, tPB ):
 
     # registers per element
     self.bpr = 4 # all registers are 32bit
-    self.bpeAB = int(self.bpr*\
-        kernel["ProblemType"]["DataType"].numRegisters())
-    self.bpeCexternal = int(self.bpr*\
-        kernel["ProblemType"]["DataType"].numRegisters())
+    self.bpeAB = int(self.bpr * kernel["ProblemType"]["DataType"].numRegisters())
+    self.bpeCexternal = int(self.bpr * kernel["ProblemType"]["DataType"].numRegisters())
+    self.bpeDexternal = int(self.bpr * kernel["ProblemType"]["DataType"].numRegisters())
```
cgmillette, line 1118: bpeCexternal / bpeDexternal should use kernel["ProblemType"]["DestDataType"] ? Treat as though A/B may be different type than C/D ?

Tensile/KernelWriterAssembly.py
```diff
@@ -1113,28 +1113,27 @@ def initKernel(self, kernel, tPA, tPB ):
...
+
     #jgolds Need to check device for support
     if kernel["ProblemType"]["HighPrecisionAccumulate"]:
         if kernel["ProblemType"]["DataType"].isHalf():
```
cgmillette, line 1122: Can perhaps combine half and bfloat 16 cases as they are identical

Tensile/KernelWriterAssembly.py
cgmillette: Also what is the difference between bpeCexternal and bpeDexternal? Do we support different types of C / D ?

TonyYHsieh: Hi @cgmillette,
We have a special case FP16 high precision accumulation.
Which compute datatype is FP16, but accumulation type is FP32.
I believe this is a legacy issue.
So we have to use bpeCexternal / bpeDexternal instead of use kernel["ProblemType"]["DestDataType"].
I hope we can correct special case FP16 and remove high precision accumulation, but we don't

In our HPA BF16 GSU algorithm, we use workspace which type is FP32 for temporarily D and C buffer which type is FP16.
So we need bpeCexternal / bpeDexternal for C & D.

Tensile/Contractions.py
TonyYHsieh: GlobalSplitUAlgorithm is used to decide which algorithm you want to use if GSU is not 0.
_GlobalAccumulation is used to decide do we have to do accumulation in workspace.
GlobalSplitUAlgorithm is a user define value, _GlobalAccumulation is a derived value.
I don't want to modify GlobalSplitUAlgorithm value by Tensile logic.

state["_GlobalAccumulation"] = 0
    if ((dataType.isBFloat16() or dataType.isHalf())
      and state["ProblemType"]["HighPrecisionAccumulate"] \
      and state["GlobalSplitU"] > 1 \
      and state["EnableMatrixInstruction"]):
      state["_GlobalAccumulation"] = state["GlobalSplitUAlgorithm"]

Tensile/Common.py
TonyYHsieh: no GSU is decided by GlobalSplitU.
GlobalSplitUAlgorithm is a sub parameter of GlobalSplitU.
It only take effect when GlobalSplitU != 1

TonyYHsieh: can we use "string" as a key for GlobalSplitUAlgorithm.
Enum is not good for yaml file.
I upload a commit to use string to replace int for GlobalSplitUAlgorithm.

Tensile/KernelWriterAssembly.py
TonyYHsieh: done

Tensile/Common.py
TonyYHsieh: Hi @cgmillette,
Does that work for you??

PR #1135, opened 2020-08-18
Kernel short circuit alpha zero

cgmillette:
Set summation dimension to zero when alpha == 0.

- Throw exception for A or B  == nullptr when alpha != 0
- A or B may be nullptr when alpha = 0
- Short circuit improves performance when alpha = 0
- Refactor GEMMKernelTest to allow testing of different datatypes.
- Only Float tests enabled until other kernel types are added to tests.
- Added short circuit tests when alpha = 0
- Added nullptr tests when alpha = 0


Tensile/Source/lib/include/Tensile/DataTypes_Int8x4.hpp
```diff
@@ -33,19 +33,29 @@ namespace Tensile
  */
     struct Int8x4
     {
+        Int8x4()
```
cgmillette, line 36: Added default ctor so can be used in vector container

HostLibraryTests/testlib/include/TestUtils.hpp
```diff
@@ -42,7 +42,7 @@ namespace Tensile
         template <typename RNG, typename... Args>
         T operator()(RNG& rng, Args&&...)
         {
-            return dist(rng);
+            return static_cast<T>(dist(rng));
```
cgmillette, line 45: Use explicit casting so that bfloat16 can be used

Tensile/KernelWriterAssembly.py
```diff
@@ -3033,6 +3033,30 @@ def allocateResources(self, kernel):
     kStr += inst("v_readfirstlane_b32", sgpr("WaveId"), vgpr(tmpVgpr), "WaveId")
     self.vgprPool.checkIn(tmpVgpr)
 
+    # Check alpha == 0
+    if self.do["ApplyAlpha"]:
```
sdquiring, line 3037: It's possible for a floating point value to be -0, in which case the comparison will fail.  I don't see a scalar floating point comparison instruction so I guess we either have to use a vector instruction or do some bit manipulation.

sdquiring:
Great! We should probably include this in the testing.  You should be able to create a -0 with [`std::copysign(0.0f, -1.0f)`](https://en.cppreference.com/w/cpp/numeric/math/copysign).

Tensile/KernelWriterAssembly.py
cgmillette: Found a way to do this without needing extra vector registers. Done!

cgmillette:
Just tested again with the rocblas pre_checkin, no failures :)

zaliu (approved):
LGTM

sdquiring (approved):

PR #1140, opened 2020-08-20
New client takes LogLevel from config.

solaslin:
* Write the log level to "log-level" of new client .ini (Error/Terse/Versobe/Debug), originally it's always Debug
* Change "SolutionProgress" to "Idx/TotalSize" instead of "Idx/LastIdx", which is aligned with "ProblemProgress"
* When DID_NOT_SATISFY_ASSERT, print only the failed predicates to make it clean.

### DIFF
--- a/Tensile/ClientWriter.py
+++ b/Tensile/ClientWriter.py
@@ -28,9 +28,29 @@ import os
 import subprocess
 from shutil import copy as shutil_copy
 from shutil import rmtree
+from enum import Enum
 
 from .Contractions import FreeIndex
 
+class DataInitName(Enum):
+  Zero = 0
+  One = 1
+  Two = 2
+  Random = 3
+  NaN = 4
+  Inf = 5
+  BadInput = 6
+  BadOutput = 7
+  SerialIdx = 8
+  SerialDim0 = 9
+  SerialDim1 = 10
+  Identity = 11
+
+class ClientLogLevel(Enum):
+  Error = 0
+  Terse = 1
+  Verbose = 2
+  Debug = 3
 
 ################################################################################
 # Main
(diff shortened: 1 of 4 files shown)


TonyYHsieh (commented):
LGTM

aazz44ss (approved):
LGTM

Tensile/ClientWriter.py
```diff
@@ -551,6 +551,12 @@ def dataInitParams(problemType):
             ('init-alpha', dataInitName(initAlpha)),
             ('init-beta',  dataInitName(initBeta))]
 
+def clientLogLevelName(num):
```
rosenrodt, line 554: Consider Enum class:

```python3
from enum import Enum
class LogLevel(Enum):
  Error = 0
  Terse = 1
  Verbose = 2
  Debug = 3

print(LogLevel.Debug.name)  # prints Debug
print(LogLevel.Debug.value) # prints 3
```

Tensile/Source/client/source/SolutionIterator.cpp
```diff
@@ -141,7 +141,7 @@ namespace Tensile
         {
             m_reporter->report(ResultKey::SolutionIndex, m_currentSolutionIdx);
             m_reporter->report(ResultKey::SolutionProgress,
-                               concatenate(m_currentSolutionIdx, "/", m_lastSolutionIdx));
+                               concatenate(m_currentSolutionIdx, "/", m_lastSolutionIdx - m_firstSolutionIdx + 1));
```
sdquiring, line 144: Won't this end up printing e.g. 12/10 if m_firstSolutionIdx > 0?

solaslin: Looks like possible. 
Somehow I just want to make the ProblemProgress (Idx/Size) consistent with SolutionProgress(Idx/LastIdx).
So I guess it's also possible for ProblemProgress to print 12/10 ?
How about unify them both to Idx/LastIdx ?

Tensile/ClientWriter.py
solaslin: Adopted. And also change the dataInitName to the same method for consistency

sdquiring:
Looks good! There's an issue with the format check.  You can ignore the ubuntu16 ones the ubuntu18 one should be fixed.

PR #1144, opened 2020-08-24
Refactor runtime contraction input type checks

cgmillette:
- Currently contraction input types as determined from runtime inputs use a cascading branching structure e.g.  if (typeA == A && typeB == B && typeC == C && typeD ==D && typeAlpha == alpha && typeBeta == beta)... else if (typeA ==  ....)... 
- Worst case scenario there are 6 * numCases comparisons, which grows unfeasibly as the support for more input cases increases.
- We can reduce this to 1 single check by assembling a bit field and using it with a switch table and compile time meta-info.
- DataType::Count is small enough to allow us to combine enums for AType, BType, CType, DType, AlphaType and BetaType into a single 32 bit bitfield, with no collisions.
- This can help reduce a lot of pressure on CPU branch prediction and stands to improve performance in this area.
- Added more aliases to commonly used TypedContractionInputs. Helps to improve readability and reduce error prone repetitive template instantiations.


Tensile/Source/lib/include/Tensile/ContractionProblem.hpp
```diff
@@ -584,6 +584,25 @@ namespace Tensile
     {
         ContractionInputs();
         virtual ~ContractionInputs();
+
+        constexpr static uint32_t typeId(DataType aType,
```
sdquiring, line 588: Static functions should be capitalized.

Tensile/Source/lib/include/Tensile/Utils.hpp
```diff
@@ -152,6 +152,31 @@ namespace Tensile
         return stream;
     }
 
+    constexpr uint32_t bitsNeeded(uint32_t maxVal)
```
sdquiring, line 155: This function name could use more context.  Maybe these functions could be grouped into a namespace or class?

Tensile/Source/lib/include/Tensile/ContractionProblem.hpp
```diff
@@ -618,10 +637,35 @@ namespace Tensile
 
         Alpha alpha = static_cast<Alpha>(0);
         Beta  beta  = static_cast<Beta>(0);
+
+        constexpr static uint32_t typeId()
+        {
+            return ContractionInputs::typeId(TypeInfo<A>::Enum,
```
sdquiring, line 643: Not sure if which is the better way but we should consider an enum class for the type IDs

```cpp
enum class ContractionTypeID: uint32_t
{
    Float = FloatContractionInputs::TypeID(),
    Double = DoubleContractionInputs::TypeID(),
    // ...
}
```
The advantage would be that if there's a type missing in any of the switch statements, there would be a compiler warning (which could be configured as an error).

The disadvantage would be that the switch statements would be a bit less hard to follow:

```cpp
switch(contractionInputsTypeId)
{
    case ContractionTypeID::Float:
    {
        auto const& typedInputs = dynamic_cast<FloatContractionInputs const&>(inputs);
        return solveTyped(problem, typedInputs, hardware);
    }
    // ...
}
```
we've lost the consistency between the case line and the body of the case.

cgmillette: When I read it, I do tend to prefer the explicitness of checking the TypedInputs::TypeId() as a reference to the following cast within the case. I think the extra layer of ID's would obfuscate it a bit.

PR #1146, opened 2020-08-26
Tensile integration test

rosenrodt:
**(Updated 11 Nov 2020: rebased & moved to independent Jenkins pipeline)**

The patch introduces integration test to ensure (to an extent) that TensileCreateLibrary, with its various build options, does not regress when invoked on up-to-date rocBLAS logic yamls.

1. Logic yamls are pulled from public rocBLAS Git repo. To make tests quick, only selected archs (gfx900/906/908) and selected problems (SGEMM/HGEMM/BF16-GEMM) are built
2. TensileCreateLibrary is tested with different build options: `--[no-]merge-files`, `[--no-legacy-components]`, `--library-format [yaml or msgpack]`, `--[no]-short-file-names`
3. For each of the above build combinations, a Tensile client will run a few problems (up to 16 for each type) sampled from the logic yaml to make sure the build artifacts' integrity

Known fail cases that've been marked xfail:
- `--library-format msgpack`: Tensile client segfaults when loading up the library
- `--short-file-name`: Can be reproduced by setting global parameter `ShortNames = True`

Misc:
- Enables benchmarking tuned solution with new client by defining `LibraryClient:` in config yaml

~~Some question(s) remain open:~~
- ~~The test is currently under CI extended test to avoid resource drain. But as discussed before, we can put integration into a separate pipeline in parallel to pre_checkin. The integration test runs about 20m with prebuilt client and 4 worker processes. @zaliu @saadrahim your opinion is appreciated here~~ **This is now on independent "Integration" pipeline. Thanks @saadrahim!**
- ~~Should we test all available TensileCreateLibrary build options? In that case I will manually trim the size of logic yaml to make each round of test shorter~~

For reference, running pre_checkin plus integration doesn't take long:

```bash
$ tox -e py35 -- Tensile/Tests -m 'pre_checkin or integration'
```
|Arch|Time|
|--|--|
|gfx906| 78 passed, 15 skipped, 4 xfailed in 1410.32s (0:23:30)
|gfx908| 86 passed, 7 skipped, 4 xfailed in 2633.28s (0:43:53)


Tensile/ClientWriter.py
```diff
@@ -95,6 +97,29 @@ def main( config ):
...
+
+  # Note: this step is needed for retrieving code object and tensile library yaml paths
+  # TODO Do away with this workaround. TensileCreateLibrary.py should take function arguments and return values directly
+  #      Refactor TensileCreateLibrary (not .py) to take CLI args instead and pass to underlying python script
+  globalParameters["GenerateManifestAndExit"] = True
+  createLibraryScript = getBuildNewClientLibraryScript(stepBaseDir, libraryLogicPath)
+  subprocess.run(shlex.split(createLibraryScript), cwd=stepBaseDir)
+  coList = []
+  yamlList = []
+  with open(os.path.join(stepBaseDir,"library","TensileManifest.txt"), "r") as f:
```
sdquiring, line 109: This should be configurable.

Tensile/ClientWriter.py
```diff
@@ -95,6 +97,29 @@ def main( config ):
...
+    lines = f.read().split("\n")
+    coList = [line for line in lines if "co" in line]
+    yamlList = [line for line in lines if "yaml" in line]
+  globalParameters["GenerateManifestAndExit"] = False
+  createLibraryScript = getBuildNewClientLibraryScript(stepBaseDir, libraryLogicPath)
+  subprocess.run(shlex.split(createLibraryScript), cwd=stepBaseDir)
```
sdquiring, line 115: Should turn on check=True in case this fails.

Tensile/Tests/integration/test_integration.py
```diff
@@ -0,0 +1,127 @@
...
+      "vega10_Cijk_Ailk_Bjlk_SB.yaml"
+    ]
+  }
+
+  files = [prefix + e for e in testData[schedule] ]
+  destDir = os.path.join(baseDir, "logic_yaml", schedule)
+  Common.ensurePath(destDir)
+
+  import glob
+  for f in glob.glob(destDir+"/*"):
```
sdquiring, line 37: `os.listdir` would be simpler.

Tensile/Tests/integration/test_integration.py
```diff
@@ -0,0 +1,127 @@
...
+from Tensile import ClientWriter, LibraryIO, Common
+from Tensile.SolutionStructs import ProblemType, Problem
+
+# 1. Call TensileCreateLibrary
+# 2. Get client instance from ClientExecutable()
+# 3. Write some problem types/sizes to ClientParameters with writeClientConfig()
+# 4. Run client with small subset of GEMM problems
+# $ pytest Tensile/Tests -m integration --capture=tee-sys --builddir /tmp/pytest
+
+def getLogicFileDir(baseDir, schedule):
```
sdquiring, line 11: The entire rocBLAS zip package is only 7.3 mb.  Maybe that's a simpler way to get these files.

Tensile/Tests/integration/test_integration.py
```diff
@@ -0,0 +1,127 @@
...
+                           CMakeBuildType="Debug"
+                           ):
+    Common.ensurePath(outputDir)
+
+    # Note: this step is needed for retrieving code object and tensile library yaml paths
+    # TODO Do away with this workaround. TensileCreateLibrary.py should take function arguments and return values directly
+    #      Refactor TensileCreateLibrary (not .py) to take CLI args instead and pass to underlying python script
+    Common.globalParameters["GenerateManifestAndExit"] = True
+    createLibraryScript = ClientWriter.getBuildNewClientLibraryScript(outputDir, logicFileDir)
+    subprocess.run(shlex.split(createLibraryScript), cwd=outputDir)
```
sdquiring, line 86: These `run()` calls should have check=True.  That will also give a better error message than just `assert(returncode=0)`

Tensile/Tests/integration/test_integration.py
```diff
@@ -0,0 +1,127 @@
...
+                           MergeFiles=mergeFiles,
+                           LibraryFormat=libraryFormat,
+                           LegacyComponents=legacyComponents,
+                           CMakeBuildType="Debug"
+                           ):
+    Common.ensurePath(outputDir)
+
+    # Note: this step is needed for retrieving code object and tensile library yaml paths
+    # TODO Do away with this workaround. TensileCreateLibrary.py should take function arguments and return values directly
+    #      Refactor TensileCreateLibrary (not .py) to take CLI args instead and pass to underlying python script
```
sdquiring, line 83: You've duplicated a lot of TensileCreateLibrary here.   Why not just use a subprocess to call the script?  That way we test the entire process.

Tensile/Tests/integration/test_integration.py
rosenrodt: that's a great suggestion

Tensile/ClientWriter.py
rosenrodt: You mean the TensileCreateLibrary input args? The integration test is already parametrized by a few of the available options of the library creation script.

rosenrodt: I am mistaken—you were actually referring to the ClientWriter. Could you be more specific about what should be configurable?

Tensile/Tests/integration/test_integration.py
rosenrodt: That duplicated code is now removed. Tensile will now always generate manifest file so no extra run with `GenerateManifestAndExit=OFF` is needed.

The `getBuildNewClientLibraryScript()` is adopted here because it already does the job of translating global param to TensileCreateLibrary input args.

Tensile/ClientWriter.py
sdquiring: I mean the location of the manifest file.

rosenrodt: The relative path to the manifest file (`library/TensileManifest.txt`) is not changeable from user's perspective. Same goes to TensileConfig.cmake, which has the exact hardcoded path. Or if you meant the parent folder `stepBaseDir`, in which case it is already configurable via `globalParameters["LibraryClientPath"]`

rosenrodt: @sdquiring ping. further comment is appreciated.

rosenrodt:
To give an update on the open questions:

1. As per our discussion, placing this test in "extended" is fine and no need for a separate CI pipeline for now
2. The build option `--msgpack` is still unclear to me if it actually works, therefore remains "xfail" 

Any comment is welcomed

zaliu:
> For reference, running pre_checkin plus integration doesn't take long:
> 
> ```shell
> $ tox -e py35 -- Tensile/Tests -m 'pre_checkin or integration'
> ```
> 
> Arch	Time
> gfx906	78 passed, 15 skipped, 4 xfailed in 1410.32s (0:23:30)
> gfx908	86 passed, 7 skipped, 4 xfailed in 2633.28s (0:43:53)

Well, without #1146 pre_checkin time is actually much shorter on standalone desktops:

    gfx906   72 passed, 9 skipped in 507.12s (0:08:27)
    gfx908   80 passed, 1 skipped in 1153.11s (0:19:13)

So the new pre_checkin time with #1146 is respectively 278% and 228% of the existing pre_checkin time, which is very significant increase.  We need to think very hard on whether to put this in pre_checkin, extended or a separate pipeline as @saadrahim indicated in an email communication.  A separate pipeline has the flexibility of future growth, which seems to be more desirable.

saadrahim:
Does this test need to be run on multiple GPU architectures? If not, it is best suited to a separate pipeline. Can it be run on a CPU only system?

zaliu:
@rosenrodt can confirm.  Since this is conceptually a subset of the rocBLAS build process, as long as a simulated target.lst is present enumerating the gfx architectures supported, Tensile is supposed to just work - unless this PR added gfx-specific tests which would require the presence of a GPU.

saadrahim:
> 
> 
> @rosenrodt can confirm. Since this is conceptually a subset of the rocBLAS build process, as long as a simulated target.lst is present enumerating the gfx architectures supported, Tensile is supposed to just work - unless this PR added gfx-specific tests which would require the presence of a GPU.

Based on what I reading, I recommend a separate test job.

Tensile/ClientWriter.py
sdquiring: I suppose making this fully configurable is outside the scope of this PR.

sdquiring:
I agree, a separate test pipeline is the best choice for this.

rosenrodt:
> Does this test need to be run on multiple GPU architectures? If not, it is best suited to a separate pipeline. Can it be run on a CPU only system?

Since this is an end-to-end integration test and not just for the build part, it still needs a GPU to complete kernel validation. It is to ensure our Tensile client implemented with public Tensile APIs correctly loads code objects from either HIP-source or ASM, under a plethora of TensileCreateLibrary build combinations.

That being said, this test is not meant to be arch-specific. The question: which one suits better for our needs, run the test (as a separate pipeline) on whatever machine available at the time, or decide early that we run on a specific gfx-arch machine? How should I configure the jenkins file based on the decision we make? @saadrahim I need your advice here. Thanks!

sdquiring (changes requested):
This should set the `TENSILE_TEST_LIBRARY` environment variable to the test library that is built so that we can run the HostLibraryTests on the newly built library.

zaliu:
@rosenrodt Can we please resolve the conflict on Tensile/ClientWriter.py?  Thanks.

rosenrodt:
> This should set the `TENSILE_TEST_LIBRARY` environment variable to the test library that is built so that we can run the HostLibraryTests on the newly built library.

@sdquiring Integration test in this PR does not adopt the test framework under HostLibraryTests but instead the tensile client. I have disabled HostLibraryTests-related tests in the groovy script to reflect that

saadrahim (approved):
code looks good from the CI side.

rosenrodt:
@sdquiring Any further feedback? Thanks!

sdquiring (approved):

PR #1148, opened 2020-08-27
Fix the incorrect CSV-merge function

solaslin:
bug-fix for csv exporter:
* Using string.find() instead of string.find_first_of(), the wrong find-result causes assert() fail
  - string key = "solution-idx";
  - key.find_first_of("Size") will return 5, which is position of "i"
  - key.find("Size") will return std::npos, means not found. This is what I expected.
* When readCurrentRow(), only get the fields that are in the headers


aazz44ss (approved):
LGTM

PR #1155, opened 2020-09-02
Add parameters to help training with BF16/FP16 GSU

TonyYHsieh:
use MaxWorkspaceSize to limitation workspace size in training.

### DIFF
--- a/Tensile/Source/client/main.cpp
+++ b/Tensile/Source/client/main.cpp
@@ -463,8 +463,7 @@ int main(int argc, const char* argv[])
     size_t maxWorkspaceSizeLimit = args["max-workspace-size"].as<size_t>();
     size_t maxWorkspaceSize
         = getMaxWorkspace(library, hardware, args, problems, firstProblemIdx, lastProblemIdx);
-    maxWorkspaceSize
-        = (maxWorkspaceSize < maxWorkspaceSizeLimit) ? maxWorkspaceSize : maxWorkspaceSizeLimit;
+    maxWorkspaceSize = std::min(maxWorkspaceSize, maxWorkspaceSizeLimit);
 
     auto dataInit = DataInitialization::Get(args, problemFactory, maxWorkspaceSize);
 
--- a/Tensile/Source/client/main.cpp
+++ b/Tensile/Source/client/main.cpp
@@ -213,6 +213,7 @@ namespace Tensile
                 ("log-level",                po::value<LogLevel>()->default_value(LogLevel::Debug),                "Log level")
                 ("exit-on-failure",          po::value<bool>()->default_value(false), "Exit run early on failed kernels.")
                 ("selection-only",           po::value<bool>()->default_value(false), "Don't run any solutions, only print kernel selections.")
+                ("max-workspace-size",       po::value<size_t>()->default_value(32*1024*1024), "Max workspace for training")
                 ;
             // clang-format on
 
@@ -459,8 +460,11 @@ int main(int argc, const char* argv[])
         lastSolutionIdx = firstSolutionIdx + numSolutions - 1;
     }
 
+    size_t maxWorkspaceSizeLimit = args["max-workspace-size"].as<size_t>();
     size_t maxWorkspaceSize
         = getMaxWorkspace(library, hardware, args, problems, firstProblemIdx, lastProblemIdx);
+    maxWorkspaceSize
+        = (maxWorkspaceSize < maxWorkspaceSizeLimit) ? maxWorkspaceSize : maxWorkspaceSizeLimit;
 
     auto dataInit = DataInitialization::Get(args, problemFactory, maxWorkspaceSize);
 
--- a/Tensile/Source/lib/include/Tensile/ContractionProblemPredicates.hpp
+++ b/Tensile/Source/lib/include/Tensile/ContractionProblemPredicates.hpp
@@ -986,6 +986,50 @@ namespace Tensile
                 }
             };
 
+            struct GlobalSplitUCheckMinK
+                : public Predicate_CRTP<GlobalSplitUCheckMinK, ContractionProblem>
+            {
+                enum
+                {
+                    HasIndex = false,
+                    HasValue = true
+                };
+
+                size_t value;
+
+                GlobalSplitUCheckMinK() = default;
+                GlobalSplitUCheckMinK(size_t value)
+                    : value(value)
+                {
+                }
+
+                static std::string Type()
+                {
+                    return "GlobalSplitUCheckMinK";
+                }
+
+                virtual bool operator()(ContractionProblem const& problem) const override
+                {
+                    return problem.boundSize(0) >= value;
+                }
+
+                virtual std::string toString() const override
+                {
+                    return concatenate(this->type(), "(value:", value, ")");
+                }
+
+                virtual bool debugEval(ContractionProblem const& problem,
+                                       std::ostream&             stream) const override
+                {
+                    bool rv = (*this)(problem);
+
+                    stream << *this << ": (" << problem.boundSize(0) << " >= " << value
+                           << ") == " << rv;
+
+                    return rv;
+                }
+            };
+
         } // namespace Contraction
 
         /**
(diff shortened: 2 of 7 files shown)


aazz44ss (approved):
LGTM

Tensile/Source/lib/include/Tensile/ContractionProblemPredicates.hpp
```diff
@@ -959,6 +959,54 @@ namespace Tensile
...
 
+            struct GlobalSplitUCheckK
+                : public Predicate_CRTP<GlobalSplitUCheckK, ContractionProblem>
+            {
+                enum
+                {
+                    HasIndex = false,
+                    HasValue = true
+                };
+                GlobalSplitUCheckKPacket value;
```
sdquiring, line 970: There's no reason the two values here couldn't just be members of this class, we would just need a custom serialization traits class for this predicate.

sdquiring (changes requested):

Tensile/Source/lib/include/Tensile/ContractionProblemPredicates.hpp
TonyYHsieh: hi @sdquiring ,

Current Property in Tensile/Properties.py only accept one index and one value.
If we want to pass two values to new client predictor, we have to pack it to one value.

So how to  achieve " custom serialization traits class for this predicate" ??

sdquiring: > custom serialization traits class

This means implementing the serialization code manually instead of simply inheriting from `AutoMappingTraits` in ContractionPredicates.hpp.  You just need to implement:

```cpp
static void mapping(IO& io, Object& obj)
```

You would also need to ensure that your properties get written out by including them in the dictionary returned by `state` in the Property base class in Python:

```python
def state(self):
```

TonyYHsieh: Hi @sdquiring ,

I modify Properties.py to add one more value parameter and extend AutoMapping to accept 2 values if needed.
That is similar to your suggestion and should be more flexible than customize case.
How do you think about this change??

Thanks,
Tony.

sdquiring: Sorry, this slipped past me in my inbox.

No, I don't like this solution.  If there are two values, what is the meaning of each variable? It can't be named something meaningful, just `value` and `value2`.

TonyYHsieh: ok, we discard modification for check K in each GSU work.
Only keep workspace limitation for Tensile training.
How about that??

zaliu:
@TonyYHsieh please resolve the file conflict.  Thanks.

Tensile/Source/lib/include/Tensile/ContractionProblemPredicates.hpp
sdquiring: Your solution to "do this a different way" is to not do it at all?

Tensile/Source/lib/include/Tensile/ContractionProblemPredicates.hpp
```cpp
struct GlobalSplitUCheckK: public Predicate_CRTP<GlobalSplitUCheckK, ContractionProblem>
{
    enum
    {
        HasIndex = true,
        HasValue = true
    };
    size_t index;
    size_t value;
    size_t k;
    // ... Need to implement manual toString() in addition to Type() and operator()().

}
```

Tensile/Source/lib/include/Tensile/Serialization/ContractionPredicates.hpp
```cpp
template <typename IO>
struct MappingTraits<Predicates::Contraction::GlobalSplitUCheckK, IO>
{
    using Predicate = Predicates::Contraction::GlobalSplitUCheckK;
    using iot     = IOTraits<IO>;

    static void mapping(IO& io, Predicate& pred)
    {
        iot::mapRequired(io, "index", pred.index);
        iot::mapRequired(io, "value", pred.value);
        iot::mapRequired(io, "k", pred.k);
    }
};

```

TonyYHsieh:
Hi @sdquiring ,

I modified change according to your suggestion.
How do you think about new modification??

Thanks,
Tony.

Tensile/Source/client/main.cpp
```diff
@@ -456,8 +457,11 @@ int main(int argc, const char* argv[])
         lastSolutionIdx = firstSolutionIdx + numSolutions - 1;
     }
 
+    size_t maxWorkspaceSizeLimit = args["max-workspace-size"].as<size_t>();
     size_t maxWorkspaceSize
         = getMaxWorkspace(library, hardware, args, problems, firstProblemIdx, lastProblemIdx);
+    maxWorkspaceSize
```
cgmillette, line 463: Just use std::min(maxWorkspaceSize, maxWorkspaceSizeLimit)

Tensile/Source/lib/include/Tensile/ContractionProblemPredicates.hpp
```diff
@@ -986,6 +986,54 @@ namespace Tensile
                 }
             };
 
+            struct GlobalSplitUCheckK
```
cgmillette, line 989: Just as a matter of naming, I would recommend GlobalSplitUCheckMinK . IMO just a bit more descriptive

cgmillette (changes requested):

Tensile/Source/lib/include/Tensile/ContractionProblemPredicates.hpp
```diff
@@ -986,6 +986,54 @@ namespace Tensile
...
+            {
+                enum
+                {
+                    HasIndex = false,
+                    HasValue = false
+                };
+
+                int globalSplitU;
+                int minKInGSU;
+
```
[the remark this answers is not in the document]
TonyYHsieh: Hi @cgmillette,

we need to pass two values instead of one value.
Do you have any suggest to pass one value??

Tensile/Source/lib/include/Tensile/ContractionProblemPredicates.hpp
```diff
@@ -986,6 +986,54 @@ namespace Tensile
...
+                static std::string Type()
+                {
+                    return "GlobalSplitUCheckK";
+                }
+
+                virtual bool operator()(ContractionProblem const& problem) const override
+                {
+                    return (globalSplitU == 1)
+                           || ((problem.boundSize(0) / globalSplitU) >= minKInGSU);
+                }
```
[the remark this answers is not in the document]
TonyYHsieh: update

Tensile/Source/lib/include/Tensile/ContractionProblemPredicates.hpp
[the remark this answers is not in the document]
TonyYHsieh: update

Tensile/Source/lib/include/Tensile/ContractionProblemPredicates.hpp
```diff
@@ -986,6 +986,54 @@ namespace Tensile
...
 
+            struct GlobalSplitUCheckK
+                : public Predicate_CRTP<GlobalSplitUCheckK, ContractionProblem>
+            {
+                enum
+                {
+                    HasIndex = false,
+                    HasValue = false
+                };
+
```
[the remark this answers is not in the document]
TonyYHsieh: update

Tensile/Properties.py
```diff
@@ -22,33 +22,38 @@
 from .Utils import hash_objs, state
 
 class Property:
-    def __init__(self, tag=None, index=None, value=None):
+    def __init__(self, tag=None, index=None, value=None, minKInGSU=None):
         self._tag = tag
         self._index = index
         self._value = value
+        self._minKInGSU = minKInGSU
```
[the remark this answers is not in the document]
TonyYHsieh: update

Tensile/Source/lib/include/Tensile/ContractionProblemPredicates.hpp
TonyYHsieh: Hi @cgmillette,
Thanks for you suggestion.
We change our patch according your comment.
Please help to review again.
By the way, _GlobalAccumulation contains condition of GSU > 1.

Tensile/Source/client/main.cpp
TonyYHsieh: update

cgmillette (approved):
Looks good! Please make sure Tensile and rocblas pre_checkin pass before merge.

TonyYHsieh:
hi @sdquiring ,

Is this change ok to you??

Thanks,
Tony.

solaslin (approved):
LGTM

sdquiring (approved):

PR #1156, opened 2020-09-02
Option to do kernel selection only, without running kernels

AlexBrownAMD:


sdquiring (approved):

AlexBrownAMD:
Added a new debug flag so other apps like rocblas-bench can skip kernel execution. To enable this mode, set the environment variable TENSILE_DB2=0x1 (to see selection, also set TENSILE_DB=0x8000). This flag is in a separate environment variable since it alters the behaviour of the library rather than just adding debug output. Also prints a warning "DEBUG: Skipping kernel execution" so it is clear when this mode is turned on. Currently, this option only skips kernel execution, everything else is done as normal - I may also need to make it skip allocating/initializing matrices to help reduce execution time.

zaliu (approved):
I ran a variant of scripts/performance/hgemm_transformer.sh in rocBLAS (by lowering iters to 1 and set cold_iters to 0) thusly:

    TENSILE_DB=0x8000 TENSILE_DB2=0x1 ./rocblas-bench --yaml hgemm_transformer.sh > hgemm.log 2>&1

it did exactly what was told: skip kernel execution and report the names of the kernels, yet it took almost 4.4 hours to finish.  I chose this list because there are a large number of sizes (4282).  My feeling  is that since kernel execution is a relatively minor portion of a rocblas-bench run, if we don't modify rocblas-bench to skip memory allocation and copying when TENSILE_DB2=0x1, we won't reach our goal of super-fast kernel name collection.  On the other hand, ~3.37 seconds/size was impressive already.  Thanks!

PR #1159, opened 2020-09-08
Revert "Benchmark tuning parts"

zaliu:
Reverts ROCmSoftwarePlatform/Tensile#1081; breaks rocblas-test


zaliu:
Not the cause of rocblas-test failures

PR #1163, opened 2020-09-14
Fix: Hold back some kick-out rules for rocBLAS library

solaslin:
Disable the kernel rejection/kick-out rules for small MT and high occupancy.
* Although this two rules may skip some benchmarks that are not likely having gain for persistent-kernel. 
However it's not proved yet for all kinds of data-type (H,S,D,BF...) so it's better to hold back this hypothesis for now.
* There are a few existing dgemm kernels in rocBLAS having PK on. We have to prevent those kernels from being kicked-out and not complied to .co files.


solaslin:
Have passed the test: TensileCreateLibrary on asm_full (rocBLAS)

zaliu (approved):

PR #1166, opened 2020-09-21
Added --generate-sources-and-exit option

Artem-B:
When specified, TensileCreateLibray generates asm and source kernels
and exists without compiling them. Assembly sources will be placed in
the output directory.

Also added -v and -j to control output verbosity and the number of
threads launched by the script.

### DIFF
--- a/Tensile/KernelWriter.py
+++ b/Tensile/KernelWriter.py
@@ -3355,18 +3355,23 @@ for codeObjectFileName in codeObjectFileNames:
 
     try:
       if kernel["KernelLanguage"] == "Assembly":
-        self.writeByteArrayScript()
-
         asmPath = self.getAssemblyDirectory()
-        coFile = self.getSingleCodeObjectFile(kernel)
         kernelName = self.getKernelName(kernel)
 
-        if globalParameters["CodeFromFiles"] or globalParameters["NewClient"] > 1:
-          # I guess in this case we are making sure that the code object file exists by executing the code
-          # above but we aren't placing it into the source.
+        if globalParameters["GenerateSourcesAndExit"]:
+          # only create the assembly file.
+          self.getKernelObjectAssemblyFile(kernel)
           return (0, "")
+        else:
+          self.writeByteArrayScript()
+          coFile = self.getSingleCodeObjectFile(kernel)
 
-        return (0, self.getFileCobaDefinition(kernelName, os.path.join(asmPath, coFile)))
+          if globalParameters["CodeFromFiles"] or globalParameters["NewClient"] > 1:
+            # I guess in this case we are making sure that the code object file exists by executing the code
+            # above but we aren't placing it into the source.
+            return (0, "")
+
+          return (0, self.getFileCobaDefinition(kernelName, os.path.join(asmPath, coFile)))
 
       else:
         return (0, self.getKernelSource(kernel))
--- a/Tensile/TensileCreateLibrary.py
+++ b/Tensile/TensileCreateLibrary.py
@@ -424,8 +424,9 @@ def writeSolutionsAndKernels(outputPath, CxxCompiler, problemTypes, solutions, k
     if kernelHeaderFile:
       kernelHeaderFile.close()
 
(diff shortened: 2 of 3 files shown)


Artem-B:
I didn't test the patch with the master branch, so I'm not sure if it works there.
Original version of the patch is on rocm-3.7.x branch and can be found here:
https://github.com/Artem-B/Tensile/commit/6893c8dedcbc30f282b04fe23301098e286692bb

Tensile/bin/TensileCreateLibrary
```diff
@@ -21,6 +21,15 @@
 ################################################################################
 # This script only gets called by CMake
 
+import google3
```
sdquiring, line 24: What is this module?

Tensile/KernelWriter.py
```diff
@@ -3268,18 +3268,23 @@ def getSourceFileString(self, kernel):
...
-        self.writeByteArrayScript()
-
         asmPath = self.getAssemblyDirectory()
-        coFile = self.getSingleCodeObjectFile(kernel)
         kernelName = self.getKernelName(kernel)
 
-        if globalParameters["CodeFromFiles"] or globalParameters["NewClient"] > 1:
-          # I guess in this case we are making sure that the code object file exists by executing the code
-          # above but we aren't placing it into the source.
+        if globalParameters["GenerateSourcesAndExit"]:
```
sdquiring, line 3274: A new global parameter should get a default value by assigning it in [Common.py](https://github.com/ROCmSoftwarePlatform/Tensile/blob/develop/Tensile/Common.py#L53).

Tensile/TensileCreateLibrary.py
```diff
@@ -1280,6 +1282,12 @@ def TensileCreateLibrary():
                           default=False, help="Output manifest file with list of expected library objects and exit.")
   argParser.add_argument("--library-format", dest="LibraryFormat", choices=["yaml", "msgpack"], \
       action="store", default="msgpack", help="select which library format to use")
+  argParser.add_argument("--generate-sources-and-exit",   dest="GenerateSourcesAndExit", action="store_true",
+                          default=False, help="Output source files only and exit.")
+  argParser.add_argument("-j",   dest="CpuThreads", type=int,
+                          default=-1, help="Number of parallel jobs to launch.")
+  argParser.add_argument("-v",   dest="PrintLevel", type=int,
+                          default=-1, help="Set printout verbosity level.")
```
sdquiring, line 1290: Good idea to make this more canonical.

sdquiring (changes requested):
Generally looks good.  I switched this to merge into develop since that fits with our development process.

Our CI will not run successfully without the default value and with the `google3` import.

cgmillette (commented):
I will sanity test this tomorrow and make sure our regular workflow isn't impacted. I see CI is failing, looks like due to uninitialized globalParameters entry. Should be easy fix.

So far, looks good minus the package import. Looks like something for bazel?

Tensile/bin/TensileCreateLibrary
Artem-B: This is an unintentional local change needed to run in our environment. I'll remove it.

Tensile/KernelWriter.py
Artem-B: Will do.

Tensile/TensileCreateLibrary.py
Artem-B: Do you mean to make them `--jobs` and `--verbose`? Or do you prefer them to match `CpuThreads`/`PrintLevel`?

Tensile/KernelWriter.py
Artem-B: Done.

Tensile/bin/TensileCreateLibrary
Artem-B: Done.

Tensile/TensileCreateLibrary.py
Artem-B: Updated to accept both `--jobs|-j` and `--verbose|-v`.

cgmillette (approved):

sdquiring (approved):

PR #1170, opened 2020-09-24
Solution fitness metric

JeremyAdamHart:
Moved content of findBestSolution functions into findBestSolutionWithFitness, which also returns a value indicating how close the solution is to a pretuned problem size. When the fitness is equal to zero, the problem size has been pretuned for. For the Euclidean Distance selection metric, this is just the euclidean distance. The findBestSolution function now calls findBestSolutionWithFitness and discards the fitness value.

I also added a test to verify that the fitness is 0 for a pretuned size and isn't 0 for a non-pretuned size.

This has been tested on Ubuntu 18.04 with a Radeon VII graphics card.


JeremyAdamHart:
I noticed the error with the static analysis, it looks like it's just a missing newline character at the end. I'll bundle it with the next commit if there's any changes that need to be made.

leekillough:
> 
> 
> I noticed the error with the static analysis, it looks like it's just a missing newline character at the end. I'll bundle it with the next commit if there's any changes that need to be made.

`.githooks/pre-commit --reformat` should fix any formatting errors.  You need to run `ln -fs ../../.githooks/pre-commit .git/hooks/` to install the `pre-commit` script so that it is automatically run before every commit.

JeremyAdamHart:
> > I noticed the error with the static analysis, it looks like it's just a missing newline character at the end. I'll bundle it with the next commit if there's any changes that need to be made.
> 
> `.githooks/pre-commit --reformat` should fix any formatting errors. You need to run `ln -fs ../../.githooks/pre-commit .git/hooks/` to install the `pre-commit` script so that it is automatically run before every commit.

Thanks! I'll make sure I set that up for the next time

sdquiring:
Does this change introduce any performance regressions? Running `HostLibraryTests/llvm/LibraryPerformance_test.cpp` should give a good sense.  There are no acceptance criteria for those tests but the reported test runtimes can help to judge a change like this.

Tensile/Source/lib/include/Tensile/CachingLibrary.hpp
```diff
@@ -157,34 +157,36 @@ namespace Tensile
...
+        virtual std::tuple<std::shared_ptr<MySolution>, double>
+            findBestSolutionWithFitness(MyProblem const& problem, Hardware const& hardware) const override
         {
             try
             {
                 auto const& amdgpu = dynamic_cast<AMDGPU const&>(hardware);
 
-                auto rv = m_cache.find(problem, amdgpu);
+                double fitness = std::numeric_limits<double>::max();
+                std::shared_ptr<MySolution> rv;
```
sdquiring, line 176: `rv` stands for return value.  Maybe this variable should be renamed to 'solution'?

JeremyAdamHart:
I modified findBestSolution to now return the fitness through a pointer argument instead of a tuple, and this managed to fix the performance drop.

sdquiring (approved):

PR #1171, opened 2020-09-28
Fixing the logic for testing if (alpha == 1) for OptNLL

solaslin:
Fixing the logic for comparing Alpha with 1.0 (F32) or 0x3c003c00 (Packed F16)? (HGEMM <H,H,H>)
* NO HPA: Clearly should cmp with 0x3c003c00 
* HPA = True, 
  * If non-PK: The cvt_f32_f16 happens very late when GlobalWrite, so should cmp with 0x3c003c00 when testing OptNLL (before GW)
  * If PK: The cvt_f32_f16 is done very early only once (right after the arguments are loaded), so should cmp with 1.0 (already cvt-ed)

### DIFF
--- a/Tensile/KernelWriterAssembly.py
+++ b/Tensile/KernelWriterAssembly.py
@@ -5990,9 +5990,14 @@ class KernelWriterAssembly(KernelWriter):
       # check alpha
       if self.do["ApplyAlpha"]:
         if kernel["ProblemType"]["DataType"].isHalf():
-          # when HPA=True and ComputeDataType=Half --> Alpha/Beta is already converted to F32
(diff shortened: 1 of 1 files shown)


solaslin:
This should be merge as soon before 3.10, or it may block some kernels from entering OptNLL

imcarsonliao (approved):

rosenrodt (approved):
Looks good. But I worry how we can catch these kind of program errors that cause perf regression but yet pass validation.

Other similar kind of perf regression I can think of is if we consume a few more vgpr/sgpr/lds under the same kernel parameter, by accident, but enough to bring down occupancy. That would be tough to catch.

PR #1174, opened 2020-09-29
Add support for hpa fp16 in fp32 out

cgmillette:
**Summary of changes:**
- Add kernels for fp16 in to fp32 out with some tests.
- Add some additional data type checks to allow ASM half/bfloat input kernels to perform calculations in fp32 with HPA and keep fp32 as output.
- Add some additional data type checks to allow SOURCE half/bfloat input kernels to perform calculations in fp32 with HPA and keep fp32 as output.
- Insert generator checks to allow fp16->fp32 asm kernel generation
-  Allow tensile_client to call for fp16->fp32 kernels
- Allow tensile_client to validate fp16->fp32 kernels
- **Updated: ** kernel naming to include characters representing <Input/Dest/Compute> type format, ONLY for BSS and HSS:

          Cijk_Ailk_Bljk_BSS_BH
                         ^^^ ^^
          OperationType_<Bfloat16/FP32/FP32>_<UseBeta/HPA>

*This is due to naming conflicts with other existing kernels of the same input type. Cannot blanket change this for every kernel now because we would have to change existing logic and avoid duplicating kernels. 

- Adjust LDS occupancy calculations to take into acct SRVW when staging output elements.
- Add pre_checkin tests for 'hsgemm' and 'bfloat16s' kernels.
- Add tuning Configs for 'hsgemm' and 'bfloat16s' kernels.

**Some secondary changes:**
	

- Adjust / move some tmpSgpr allocations to squeeze in a few more kernels
-  Addition of a 32 bit offset longjump - used for skip condition past loop over summation sizes. Fixes kernels where the address offset > SIMM16.
-  Fix some malformed asserts
-  Add a 32 bpl load in chooseGlobalRead - (2x bufferload_dwordx4)
-  Cherry pick aazz44ss@2e5c97e to allow new tuning file updates:
 	○ SRVW allowed as -1
 	○ New MaxWorkspaceSize and MinKForGSU parameters

- Tuning sizes adapted from PR #1189
- Fix unit test failure due to new kernel names
- Address issues from code review


Tensile/KernelWriterAssembly.py
```diff
@@ -10347,13 +10396,15 @@ def globalWriteElements(self, kernel, vectorWidths, elements,
...
-          self.vgprPool.checkIn(self.betaVgpr)
-          self.betaVgpr = None
+        if kernel["ProblemType"]["HighPrecisionAccumulate"] and kernel["ProblemType"]["DestDataType"].isHalf():
+          self.betaVgpr = self.vgprPool.checkOut(1, "beta")
+          kStr += inst("v_mov_b32", vgpr(self.betaVgpr), sgpr("Beta"), "sgpr -> vgpr b/c op_sel")
+          kStr += inst("v_cvt_f32_f16", vgpr(self.betaVgpr), vgpr(self.betaVgpr), "convert beta to fp32")
+          if self.betaInSgpr:
+            kStr += inst("v_readfirstlane_b32", sgpr("Beta"), vgpr(self.betaVgpr), "restore beta sgpr")
+            self.vgprPool.checkIn(self.betaVgpr)
+            self.betaVgpr = None
```
solaslin, line 10406: The outer "if" already checked if the kernel is <H/H/H>+HPA and <H/S/H>+HPA, so
1. No need to check kernel["ProblemType"]["HighPrecisionAccumulate"] again here
2. Why does the beta conversion requires kernel["ProblemType"]["DestDataType"].isHalf() ?  I think we still need to do this conversion once ComputeDataType is Half + HPA. So maybe we also don't need this DestDataType.isHalf test as well ?

Tensile/SolutionStructs.py
```diff
@@ -942,14 +942,16 @@ def checkIfSupportedGEMMType(self):
...
+        printExit("DataType=H only allows DestDataType=H/S and ComputeDataType=H/S")
       # if not (outType.isHalf() and computeType.isHalf()):
       #   printExit("DataType=H only allows DestDataType=H and ComputeDataType=H")
 
     if inType.isBFloat16():
-      if not (outType.isBFloat16() and computeType.isSingle()):
-        printExit("DataType=B only allows DestDataType=B and ComputeDataType=S")
+      if not ((outType.isBFloat16() or outType.isSingle()) and  \
+              computeType.isSingle()) :
+        printExit("DataType=B only allows DestDataType=B/S  and ComputeDataType=S")
```
solaslin, line 955: I am wondering how can I benchmark the new type and the validation?  Since I don't see any changes or new typed added in tensile client code.

solaslin: Sorry my bad, I figure that the HSS and BSS APIs are already implemented before. But not HSH, so maybe we have to reject HSH in checkIfSupportedGEMMType() in SolutionStruct.py

solaslin (changes requested):
checkIfSupportedGEMMType needs to reject <H,S,H>, since tensile_host haven't supported this API yet.

Tensile/SolutionStructs.py
cgmillette: #1144 did this already in my prep for this PR. Here I also added the configs so that we might actually test BSS and HSS

Tensile/KernelWriterAssembly.py
cgmillette: Hm I think this is a merge artifact from #1169 and what I needed before.

- Fixed.

Tensile/SolutionStructs.py
```diff
@@ -942,14 +942,16 @@ def checkIfSupportedGEMMType(self):
     if inType.isHalf():
       # TODO- Test and migrate ([H/H/H]+HPA) to ([H/H/S]+HPA)
       # Note that we need to do a little change in rocBLAS and logic yaml
-      if not (outType.isHalf() and (computeType.isHalf() or computeType.isSingle())):
-        printExit("DataType=H only allows DestDataType=H and ComputeDataType=H/S")
+      if not ((outType.isHalf() or outType.isSingle()) and \
+              (computeType.isHalf() or computeType.isSingle())):
+        printExit("DataType=H only allows DestDataType=H/S and ComputeDataType=H/S")
```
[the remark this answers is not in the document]
cgmillette: Good point. Not in scope for this PR. Will disable.

- Fixed

Tensile/SolutionStructs.py
cgmillette: - Fixed.

Tensile/Common.py
```diff
@@ -726,7 +726,7 @@
     # 0:   Disable StoreRemap (default)
     # 1~8: Enable StoreRemap and set the global write vector width
     # Suggest optimum value: fp32 = [2,4], fp16 or bf16 = [4,8] (dwordx2 and dowrdx4)
-    "StoreRemapVectorWidth":      [0,1,2,4,8],
+    "StoreRemapVectorWidth":      [-1,0,1,2,4,8],
```
sdquiring, line 729: Can you explain this new value in the comment above?

solaslin: I believe this is cherry-picked from other commit, maybe @aazz44ss can answer how he did the auto-assignment for the -1 value?

solaslin (approved):

Tensile/SolutionStructs.py
```diff
@@ -1167,9 +1169,15 @@ def __str__(self):
...
       name += "C"
 
-    # precision and other
+    # DataTypes
+    # Format A/B/C/D/Alpha/Beta consistent with rocblas
+    name += "_"
+    name += self["DataType"].toChar()        # Type of A/B
+    name += self["DestDataType"].toChar()    # Type of C/D
+    name += self["ComputeDataType"].toChar() # Type of Alpha/Beta
+
```
solaslin, line 1178: I agree the renaming of the kernel. But will this have some problems to merge the new inc. logics with existing logics in rocBLAS? I believe <H,S,S> and <B,S,S> may not have problems since no existing logics in rocBLAS. But what about our current <B,B,S>? The file names and kernel names are changed so maybe we need to do some tests on this ?

solaslin: For example, 
H,H,H+HPA: old name= _HBH_  <--> new name= _HHH_BH_
B,B,S+HPA: old name= _BBH_  <--> new name= _BBS_BH_ , 
We need to merge two different filenames. 

In addition, if a problem doesn't update the best solution (which is using new name), it might keep the old kernel which is using old name. I'm not sure if this will cause problems, maybe not, but do we need to make it more consistent ?

zaliu: @solaslin Transitioning existing names to the new naming scheme is not absolutely cast to stone at this point.   I'm not quite convinced that existing names must or should be mapped to the new names.  I'm not seeing any major benefit other than "consistency in naming".  All details must be carefully considered and assessed before a final technical decision can be made.

solaslin: @zaliu ok, I got it. but despite of the kernel names inside logic files, the merge script still need to be able to merge files with different filenames, or the new logic won't match the original filenames. I think this can be extended in #1177 @rosenrodt

cgmillette: Or, what we can do is:
If (Ti != To)
   Use new ProblemType_<Ti/To/Tc>_<H/B> format
else
   Use ProblemType_<Ti/H/B> format as before

Tensile/Common.py
cgmillette: @aazz44ss  Can you please offer some insight on the -1 setting above? Thanks!

Tensile/SolutionStructs.py
cgmillette: To avoid the merging issue for now, I've restricted the kernel naming to be applied only on HSS and BSS kernels. This will be fine until (if/when) we decide to apply to all kernels. I am running short on time for the PR so I will only consider the specific cases, @zaliu @solaslin

rosenrodt:
I'd suggest that you do a `git rebase -i` to squash the "fixes" and "cleanups" unless you want to make clear those fixes/cleanups are a direct response to folk's feedbacks

solaslin (changes requested):
@cgmillette 
I just checked-out this PR and did a git-rebase, based on the latest develop branch.
Here is the rebased branch, I've checked no codes are missing.
https://github.com/solaslin/Tensile/tree/rebased_pr1174

So I would suggest you to checkout the above rebased_pr1174 branch to your local branch (e.g. cgmillette:rebased_1174), and then do a "git push -f" to force push to this "cgmillette:add-support-for-fp16-in-fp32-out".

This could make this PR rebased from the latest develop, and hence we won't see the extra merge commits like "Merge branch 'develop'......"

leekillough:
`git rebase` is unnecessarily complicated, and GitHub "squash and merge" is not a true squash -- use `git reset` instead. After merging local source tree with the latest `develop`, use `git reset --soft develop`, and then commit changes.

cgmillette:
Rebased onto develop and cleaned up commits.

solaslin (approved):

ramjana (approved):
Looks good

PR #1182, opened 2020-10-09
restore default PrintLevel to 1

zaliu:
#1166 sets default PrintLevel to -1 in error while the default for PrintLevel in Common.py should be 1.  This causes some important information to fail to show up, in particular the replacement_kernel logging.


cgmillette (approved):
LGTM

PR #1185, opened 2020-10-12
MatrixInstruction for general int8 

solaslin:
Implement MatrixInstruction for general Int8 type instead of Int8x4
* Only support MatrixInstruction now. No legacy VALU and source HIP
* Implemented int8 type in Tensile new client
* Support the existing parameters addressed in the commit msg
* Extend LRVW to 16, GRVW to 16
* SplitLDS not supported yet


Tensile/Common.py
```diff
@@ -158,6 +158,14 @@
 globalParameters["DebugKernel"] = False           # assembly only, kernel gets buffer for debug "printing"; kernel writes data to memory, gets coppied to host and printed
 globalParameters["LibraryPrintDebug"] = False     # solutions will print enqueue info when enqueueing a kernel
 
+# debug for assembly
+globalParameters["EnableAsserts"] = False         # Enable assembly debug assert
+globalParameters["EnableDebugA"] = False          # Enable / Disable CheckValue1A
+globalParameters["EnableDebugB"] = False          # Enable / Disable CheckValue1B
+globalParameters["EnableDebugC"] = False          # Enable / Disable CheckValueC
+globalParameters["ExpectedValueC"] = 16.0         # Expected C Value when CheckValueC, debug for Alpha*A*B
```
cgmillette, line 166: ExpectedValue = arbitrary?  Not entirely sure this should live in globalParams.

solaslin: This is a quick setter for the debug value in KernelAssemblyWriter :: db["ValueCExpectedValue"] which has been existing from very beginning. I added this block (6 parameters) to help us quickly set the debug value for different input, otherwise we need to modify the KernelAssemblyWriter :: db["...."] value each time. 
Do you have any suggestion to refine this ? Or maybe change the name to "DebugExpectedValueC" to clarify this ?

Tensile/KernelWriter.py
```diff
@@ -2614,28 +2680,28 @@ def initKernel(self, kernel, tensorParametersA, tensorParametersB ):
     self.numReadVectorComponentsA = kernel["GlobalLoadVectorWidthA"] \
         if (self.readTileDimComponentsA \
         or self.readUnrollDimComponentsA) else 1
-    self.numWriteVectorComponentsA = kernel["GlobalLoadVectorWidthA"] \
-        if (self.writeTileDimComponentsA \
-        or self.writeUnrollDimComponentsA) else 1
-    self.numReadTileVectorComponentsA = kernel["GlobalLoadVectorWidthA"] \
-        if self.readTileDimComponentsA else 1 # for branches
+    # self.numWriteVectorComponentsA = kernel["GlobalLoadVectorWidthA"] \
```
AlexBrownAMD, line 2683: We can probably just delete all this unused code unless its going to be needed again soon.

AlexBrownAMD (approved):

Tensile/Common.py
```diff
@@ -294,6 +302,18 @@
...
+# This is used in SolutionStruct.py::checkIfSupportedGEMMType()
+validGEMMTypes = [ ('D','D','D'), ('S','S','S'), ('Z','Z','Z'), ('C','C','C'), \
+                  ('H','H','H'), ('H','H','S'), ('H','S','S'), \
+                  ('B','B','S'), ('B','S','S'), \
+                  ('4xi8','I','I'), \
+                  ('I8','I','I')]
+
+# These type are newly supported and we would like to use a better file naming for them: _TiToTc_
+# For the rest of the typed, we keep them with old existing naming.
+typesUsingNewNaming = [ ('H','H','S'), ('H','S','S'), ('B','S','S'),('I8','I','I')]
```
solaslin, line 315: @cgmillette how do you think about this refactor ? Use the look-up list to simplify checkIfSupportedGEMMType() and specify which kinds of kernel will use the new naming _TiToTc_

Tensile/KernelWriterAssembly.py
```diff
@@ -2488,6 +2502,20 @@ def functionSignature(self, kernel ):
...
+    if globalParameters["AsmCaps"][self.version]["HasLshlOr"]:
+      kStr += r"    v_lshl_or_b32 \dst, \src0, \shiftCnt, \src1" + self.endLine
+    else:
+      kStr += r"    v_lshlrev_b32 \dst, \shiftCnt, \src0" + self.endLine
+      kStr += r"    v_or_b32 \dst, \dst, \src1" + self.endLine
+      # if self.AsmBugs["ExplicitCO"]:
+      #   kStr += r"    v_add_co_u32 \dst, vcc, \src0, \src1" + self.endLine
+      # else:
+      #   kStr += r"    v_or_b32 \dst, \dst, \src1" + self.endLine
+    kStr += ".endm" + self.endLine
```
solaslin, line 2517: Not sure about the purpose of ExplicitCO

Tensile/KernelWriterAssembly.py
```diff
@@ -8105,6 +8248,11 @@ def localReadDoVALU(self, kernel, bufferIdx, iui, epsi, tP):
             localReadCode.addInst("s_mov_b32", sgpr(tmpSgpr), hex(0x3f803f80),"CheckValue1: BF16")   # packed 1s
             localReadCode.addCode(self.assert_eq( dbgVgpr, sgpr(tmpSgpr)))
 
+          # TODO - Check if this works
+          if kernel["ProblemType"]["DataType"].isInt8():
+            localReadCode.addInst("s_mov_b32", sgpr(tmpSgpr), hex(0x01010101),"CheckValue1: INT8")   # packed 1s
+            localReadCode.addCode(self.assert_eq( dbgVgpr, sgpr(tmpSgpr)))
+
```
solaslin, line 8255: This could be removed once we are sure Int8 will not be supported in VALU version

Tensile/SolutionStructs.py
```diff
@@ -3292,10 +3316,12 @@ def assignDerivedParameters(state):
         if not (state["PrefetchLocalRead"] >= state["LoopIters"] and state["InnerUnroll"] == 1) and \
           not state["InnerUnroll"] >= state["LocalReadVectorWidth"]//state["ProblemType"]["DataType"].numMIInput():
           reject(state, "wider localRead only support (PrefetchLocalRead %u >= LoopIters %u) or (InnerUnroll %u > LocalReadxN)" % (state["PrefetchLocalRead"],state["LoopIters"],state["InnerUnroll"]))
-      if (state["LoopIters"] - (state["PrefetchLocalRead"]%state["LoopIters"])*state["LocalReadVectorWidth"]//state["ProblemType"]["DataType"].numMIInput()) < 0:
-        reject(state, "LoopIters %u should greater than PrefetchIters %u" % ((state["LoopIters"],(state["PrefetchLocalRead"]%state["LoopIters"])*state["LocalReadVectorWidth"]//state["ProblemType"]["DataType"].numMIInput())))
```
solaslin, line 3296: Remove this test since it is duplicated.

Tensile/Source/lib/include/Tensile/DataTypes_Int8.hpp
```diff
@@ -0,0 +1,51 @@
...
+    {
+    };
+} // namespace Tensile
+
+namespace std
+{
+    inline ostream& operator<<(ostream& stream, const Tensile::Int8 val)
+    {
+        return stream << static_cast<int32_t>(val);
+    }
```
solaslin, line 50: The only purpose of adding this Int8 type is ostream operator<<.
Since when calling logTensorTyped(), we need to print the actual number as int32_t instead of int8_t (it will be printed as characters)

Tensile/KernelWriterAssembly.py
sdquiring: GFX8 has no `v_add_co_u32` instruction. It is implicit in the `v_add_u32`.  GFX9 changed the mnemonic for including carry-out (ExplicitCO).  GFX10 renamed the no-carry case to `v_add_nc_u32` (ExplicitNC).

solaslin: Thanks for the reply, I got it now. So looks like I don't have to care the CO for _v_lshl_or_b32, I'll remove this comment block later.

Tensile/Common.py
cgmillette: Can probably leave as is for now

Tensile/Common.py
cgmillette: I agree with the refactor - Much cleaner except we lose a little bit of verbosity. Not a big deal.

Tensile/Common.py
```diff
@@ -1434,6 +1468,9 @@ def GetAsmCaps(isaVersion):
   else:
     rv["MaxVmcnt"] = 0
 
+  # TODO- Need to query the max cap, just like vmcnt as well?
+  rv["MaxLgkmcnt"] = 15
+
```
cgmillette, line 1473: Good question - @sdquiring do you know of any limitations for MaxLgkmcnt ? 

Also what is the reasoning behind MaxVmcnt checks at 63 and 15? We could maybe add a little note for each of those numbers.

Tensile/KernelWriter.py
```diff
@@ -1273,8 +1293,10 @@ def noLoadLoop( self, kernel, tensorParametersA, tensorParametersB, isOptNLL, is
             self.globalReadIncrements.addCode(Code.Module("globalReadIncrementA"))
             self.globalReadIncrements.addCode(Code.Module("globalReadIncrementB"))
         if not isLastLoop:
-          self.localWriteACode = self.localWriteDo(kernel, tensorParametersA, (uDu+1)%kernel["DepthULdsDivisor"])  # local write in loopcnt N targets data for loopcnt N+1
-          self.localWriteBCode = self.localWriteDo(kernel, tensorParametersB, (uDu+1)%kernel["DepthULdsDivisor"])
+          self.localWriteACode, tmpCheckedOutVgprA = self.localWriteDo(kernel, tensorParametersA, (uDu+1)%kernel["DepthULdsDivisor"])  # local write in loopcnt N targets data for loopcnt N+1
+          self.localWriteBCode, tmpCheckedOutVgprB = self.localWriteDo(kernel, tensorParametersB, (uDu+1)%kernel["DepthULdsDivisor"])
+          self.tmpCheckedOutLWVgprs.append(tmpCheckedOutVgprA)
+          self.tmpCheckedOutLWVgprs.append(tmpCheckedOutVgprB)
```
cgmillette, line 1299: There have been some thoughts about using an RAII approach to temp registers within the allocation pool. This has been loosely done with sgprs  with KernelWriterAssembly.TmpSgpr. (Though would be better to override __enter__ and __exit__ for explicit cleanup).

Idea is that on instantiation registers are checked out and on deletion registers are checked in. This way the registers at least are cleaned up once they go out of scope and we don't manually have to check them in.

This might be enough scope for a separate ticket, but maybe we can work on this in a different ticket. I just wanted to get the idea out there.

Tensile/SolutionStructs.py
```diff
@@ -932,45 +933,12 @@ def checkIfSupportedGEMMType(self):
...
-    # TODO - WIP
-    # if inType.isInt8():
-    #   if not (outType.isInt32() and computeType.isInt32()):
-    #     printExit("DataType=i8 only allows DestDataType=I and ComputeDataType=I")
+    gemmType = ( inType.toChar(), outType.toChar(), computeType.toChar() )
+    if gemmType not in validGEMMTypes:
+      printExit("This typed-GEMM (Ti, To, Tc) = (%s, %s, %s) is not supported yet."%(gemmType[0],gemmType[1],gemmType[2]))
 
+    # TODO- Migrate ([H/H/H]+HPA) to ([H/H/S]+HPA)
+    # Note that we need to do a little change in rocBLAS and logic yaml
```
cgmillette, line 941: Much cleaner!

Tensile/Common.py
cgmillette: Yep, looks good!

cgmillette (approved):
Looks good to me

Tensile/KernelWriter.py
solaslin: Yes, I agree on the tempReg is not really  convenient to use, and tmpGpr could be overwriten when nested scope..Maybe we can have another task to refine the regPool

PR #1186, opened 2020-10-13
Print warning for unexpected rejection during TensileCreateLibrary

solaslin:
* Main purpose for this is to show more msgs when TensileCreaetLibrary failed, espcially due to unexpected solution rejection. 
* Print the rejection reason and the SolutionID/Name **can help us spot which solution and which rejection is the root cause. Can help us to find the error quickly**.
---
* **Support new arguments** for ${​​​​TensileCreateLibrary} cmd​​​​:  --global-parameters=[.....]
  * Example: ${​​​​TensileCreateLibrary}​​​​ **--global-parameters=PrintSolutionRejectionReason=True**
  * This enables the globalParam["PrintSolutionRejectionReason"] =True. 
  * Useful when we are directly calling TensileCreateLibary
---
* If a LibraryLogic contains a solution that is rejected (any rejection in a LibraryLogic is unexpected)
  * It will show the warning messsage, like:
```
 reject: HalfEcc requires GLVWA > 1
 !! Warning: Any rejection of a LibraryLogic is not expected, please check. SolutionIndex:0, (or SolutionName/ProblemType: Cijk_Alik_Bjlk_HBH_MT016x128x08_PGR1_PLR1_TT04_04)
```   
 * So we can find that solution either by Id or name, and check why it was rejected by the reason.

### DIFF
--- a/Tensile/SolutionStructs.py
+++ b/Tensile/SolutionStructs.py
@@ -50,6 +50,17 @@ def reject(state, *args):
     for a in args:
       print(a)
     #traceback.print_stack(None, 2)
+    solutionIndex = state["SolutionIndex"] if (state != None and "SolutionIndex" in state) else -1
+    if solutionIndex != -1:
+      # If we have valid solutionIndex, this means we are during TensileCreateLibrary stage
+      # In this stage, all solutions in the logic should be valid
+      # So if any rejection happens, print the warning for further check
+      # This will be done only when --global-parameters=PrintSolutionRejectionReason=True
+      solutionNameMin = state["SolutionNameMin"] if ("SolutionNameMin" in state) else None
+      # if we don't have SolutionNameMin, we simply use the problemTypeName
+      solutionNameMin = str(state["ProblemType"]) if (solutionNameMin == None) else solutionNameMin
+      print("!! Warning: Any rejection of a LibraryLogic is not expected, please check. \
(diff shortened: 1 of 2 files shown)


Tensile/SolutionStructs.py
```diff
@@ -50,6 +50,16 @@ def reject(state, *args):
     for a in args:
       print(a)
     #traceback.print_stack(None, 2)
+    if globalParameters["_DebugCreateLibrary_Internal"]:
```
aazz44ss, line 53: PrintSolutionRejectionReason is false in TensileCreateLibrary?
if so, TensileCreateLibrary will not enter this condition.

solaslin: Default is false since we don't want to print the bunch of msgs when rocBlas build or doing benchmarks. So you need to add the --global-parameters=PrintSolutionRejectionReason=True when we are debugging purely via TensileCreateLibrary

aazz44ss (approved):

Tensile/SolutionStructs.py
rosenrodt: IMHO we can do away this if-clause by simply printing the message whenever the keys ("SolutionIndex", "SolutionNameMin", ...) are available. So we can avoid adding an extra parameter

solaslin: @rosenrodt Sounds good, I was also wondering how to avoid this "_DebugCreateLibrary_Internal" from being overrideen by Config.

rosenrodt (approved):

PR #1200, opened 2020-10-21
[Hotfix] PK kernel is never picked in rocBLAS, causing perf. drop in dgemm

solaslin:
* PK-predicates function uses bool value **m_eligibleForPK** flag, which is tested in **checkSolution()** during benchmarks to speed up tunning.
* However, in rocBLAS, we never call **checkSolution()** but **findBestSolution()** instead, so the eligibleForPK value is not tested and hence is the dafule value = false. This makes the PK-predicates always fail so any PK-kernel wouldn't be picked (in rocBLAS).
* Makes the eligibleForPK to true as default, **checkSolution()** will turn it off if not eligible during benchmarks. In rocBLAS, we simply don't need to check the eligibility for assume it is true.

### DIFF
--- a/Tensile/Source/lib/include/Tensile/ContractionProblem.hpp
+++ b/Tensile/Source/lib/include/Tensile/ContractionProblem.hpp
@@ -545,7 +545,7 @@ namespace Tensile
         bool           m_transB;
         bool           m_highPrecisionAccumulate = false;
         bool           m_deterministicMode       = false;
-        bool           m_eligibleForPK           = false;
+        bool           m_eligibleForPK           = true;
         ArithmeticUnit m_arithmeticUnit          = ArithmeticUnit::Any;
         KernelLanguage m_kernelLanguage          = KernelLanguage::Any;
 


imcarsonliao (approved):

zaliu (approved):
Please make sure you follow the Tensile Hot fix process [here](http://confluence.amd.com/display/MLSE/Tensile+monthly+release+for+ROCm).  In particular, since ROCm 3.10 branching already took place, please target the Tensile release branch rocm-3.10.x, not master.

Note that we need approval to merge into Tensile rocm-3.10.x; the process requires assigning JIRA ticket to Chanchali, Lingamurty and receiving approval before merge.

Lastly, since this PR requires the update of rocBLAS tensile_tag.txt, please follow rocBLAS team's Hot fix process [here](http://confluence.amd.com/display/MLSE/rocBLAS-internal+and+hipBLAS+feature+complete+merge) for the update of tensile_tag.txt.

PR #1201, opened 2020-10-22
merge rocm-3.10.x into develop after hot-fix #1200

solaslin:
merge rocm-3.10.x into develop after hot-fix #1200


zaliu (approved):
LGTM.  Please remember next time, go from rocm-3.10.x to master, then from master to develop.  Thanks.

PR #1205, opened 2020-10-26
ignore LdcEqualsLdd and let all kernel support ldc != ldd.

TonyYHsieh:
ignore LdcEqualsLdd and let all kernel support ldc != ldd.

This change let Tensile inherently support ldc != ldd.
So caller, ex. rocblas, doesn't have to handle ldc!=ldd.

We have compare performance before and after change.
image


TonyYHsieh:
compare performance between before and after this change.
[ldd.xlsx](https://github.com/ROCmSoftwarePlatform/Tensile/files/5437368/ldd.xlsx)

TonyYHsieh:
open later

PR #1213, opened 2020-11-03
PersistentKernel Refine: Support PKn1, EPS, optimize LW-vmcnt, and sMagicDiv2

solaslin:
* A **PKn1** (PK=-1) will determine a better PK value when launching kernel in host:
  * Automatically choose a "heuristic" value that can possibly get a better gain: (TilesPerWorkgroup = 1~2)
  * Not based on any theory, but on some experiment observation, can be used to reduce the kernels
<br>

* PAP: Support EPS=1 , and fix EPS=1 + PLR
  * Related code modification: **BreakAtEvenIter** or **ExpandPointerSwap**
<br>

* PAP: Optimize some local write waitcnt vmcnt in PreLoop
  * Related code modification: **PreLoopVmcnt**
<br>

* Fix the insufficient precision when using sMagicDiv, add a new sMagicDivAlg2
  * Related code modification: **sMagicDivAlg2, MagicNumber... , MagicShift...**
  * Division is used when we're mapping the PK-WorkGroup to the MacroTile (need quotient and remainder)
  * For big problems, such as 144000 MTs, doing a division like 143993/144000 gives the quotient = 1, should be zero


Tensile/KernelWriterAssembly.py
```diff
@@ -7196,6 +7222,117 @@ def calculateLdsWriteOffset(self, perp, para, sPerp, sPara, kernel, tP, localWri
...
+      LWDoCase3Mod.addText(str(item).replace("__placeholder__","OrdNLL_B0_GW_vmcnt"))
+    LWDoCase3Mod.addInst("s_branch", lwEnd_Label, "finish case 3, jump to end of LW")
+
+    # CASE 4:
+    # duplicate vmcnt("__placeholder__ + Basic_GL_vmcnt - Decrement") to vmcnt("OrdNLL_B1_GW_vmcnt + Basic_GL_vmcnt - Decrement")
+    LWDoCase4Mod = imod.addCode(Code.Module("PreLoopLW_Case4"))
+    LWDoCase4Mod.addText("\n%s:" % ordNLL_B1_lw_Label)
+    LWDoCase4Mod.addComment1("prev-global-store-cnt = OrdNLL_B1_GW_vmcnt, global-load-cnt = Basic_GL_vmcnt")
+    if kernel["StoreRemapVectorWidth"]:
+      LWDoCase4Mod.addText("\ns_barrier // since no need for any vmcnt for case4, so need a barrier to ensure ds_read of SR is done\n")
```
rosenrodt, line 7305: I think all cases, except for first time entering the preloop, needs this barrier when store remapping is enabled. One wave could get here first while other waves are still reading from the LDS.

solaslin: Good catch! Updated and refined

solaslin:
@rosenrodt @jichangjichang @ramjana 
please help to review these refinement :)

rosenrodt (approved):

solaslin:
@cgmillette would you like to help review this if you have some time, thanks !

Tensile/Source/lib/source/ContractionProblem.cpp
```diff
@@ -638,6 +635,19 @@ namespace Tensile
         if(sizeMapping.persistentKernelAlongBatch)
             problemTiles *= numWG.z;
 
+        size_t cuCount      = dynamic_cast<AMDGPU const&>(hardware).computeUnitCount;
```
cgmillette, line 638: If you are not doing any nullptr checking and are fairly certain this will be a AMDGPU object, maybe consider static_cast? Is there any reason hardware obj would be something else?

If dynamic casting is still required, might be a good idea to add nullptr checks.

Tensile/SolutionStructs.py
```diff
@@ -2485,9 +2485,17 @@ def assignDerivedParameters(state):
...
 
-    if state["PrefetchAcrossPersistent"] and (state["PersistentKernel"] == 0):
-      # warn("PrefetchAcrossPersistent requires PersistentKernel != 0, forcing PrefetchAcrossPersistent = False")
-      state["PrefetchAcrossPersistent"] = False
+    if state["PrefetchAcrossPersistent"]:
+      if state["KernelLanguage"] == "Source" or \
+         state["PersistentKernel"] == 0 or \
+         state["PrefetchGlobalRead"] == 0 or \
+         state["SuppressNoLoadLoop"]:
+        #warn("PAP requires Assembly, PK != 0, PGR != 0, SuppressNoLoadLoop = True, forcing PAP = False")
```
AlexBrownAMD, line 2493: warning might be good here since an option is being overridden.

Tensile/SolutionStructs.py
```diff
@@ -2485,9 +2485,17 @@ def assignDerivedParameters(state):
...
+        state["PrefetchAcrossPersistent"] = False
+
+    # TODO- fix this, avoid the bug for now
+    if state["PrefetchAcrossPersistent"] and state["StaggerU"] == 0:
+        state["PrefetchAcrossPersistent"] = False
```
AlexBrownAMD, line 2498: warning here too could help as reminder to implement feature later if needed

Tensile/Source/lib/source/ContractionSolution.cpp
```diff
@@ -308,16 +308,28 @@ namespace Tensile
...
+            {
+                // 1. Get the largest pk value (ex.3)
+                //    which can make the PK.G (ex.3*120=360) <= problemGroups (ex.433)
+                // 2. Scale by 5/8 (can try 0.5~1, to control the tiles-per-workgroup = 1~2)
+                finalPKValue = 5 * (problemGroups / cuCount) / 8;
+                finalPKValue = std::max(finalPKValue, (size_t)1);
+                //std::cout << "final persistent kernel value: " << finalPKValue << std::endl;
+            }
+
+            size_t persistentGroups = cuCount * finalPKValue;
```
AlexBrownAMD, line 330: Repeated in ContractionProblem - these should probably be combined somehow since they involve a heuristic that looks like you might try to adjust in the future.

Tensile/KernelWriter.py
```diff
@@ -1546,7 +1544,9 @@ def kernelBody( self, kernel, tensorParametersA, tensorParametersB ):
         if self.enable["LocalRead"]:
           for plrIdx in range(0, self.numItersPLR):
             pack[plrIdx] = Code.Module()
-            for espi in range(0, (self.prefetchAcrossPersistent and kernel["ExpandPointerSwap"])+1):
+            # no matter EPS or PAP, only prefect local once per plrIdx
+            # for espi in range(0, (self.prefetchAcrossPersistent and kernel["ExpandPointerSwap"])+1):
+            for espi in range(0, 1):
```
AlexBrownAMD, line 1549: Does this alter the behavior of EPS?

Tensile/KernelWriterAssembly.py
```diff
@@ -3325,20 +3389,21 @@ def graWorkGroup(self, kernel, isPap):
         # re-mapping WG2 to WGKSerial -> wg2
         # re-mapping SerialWorkGroupIter to WGIJSerial -> wg0/1
         kStr += self.comment1("compute SerialWorkGroupIter / problemNumGroupTiles0x1 (aka nWG0*nWG1)")
-        kStr += self.sMagicDiv(kernel, stmp, sgpr("SerialWorkGroupIter"), sgpr("MagicNumProblemNumGroupTiles0By1"), 31)
+        kStr += self.sMagicDivAlg2(kernel, stmp, sgpr("SerialWorkGroupIter"), sgpr("MagicNumProblemNumGroupTiles0By1"), sgpr("MagicShiftProblemNumGroupTiles0By1"))
         kStr += inst("s_mov_b32", sgpr("WGKSerial"), sgpr(stmp), "wgKSerial = SerialWorkGroupIter / problemNumGroupTiles0x1")
-        kStr += inst("s_mul_i32", sgpr("WGIJSerial"), sgpr(stmp), sgpr("NumWorkGroups0By1"), "remainder part 1 : quotient * divisor")
+        kStr += inst("s_mul_i32", sgpr("WGIJSerial"), sgpr(stmp), sgpr("NumWorkGroups0"), "remainder part 1 : quotient * divisor (divisor = NumWorkGroups0 x 1)")
+        kStr += inst("s_mul_i32", sgpr("WGIJSerial"), sgpr(stmp), sgpr("NumWorkGroups1"), "remainder part 1 : quotient * divisor (divisor = NumWorkGroups0 x 1)")
```
AlexBrownAMD, line 3395: Typo in comment

Tensile/KernelWriterAssembly.py
```diff
@@ -5432,10 +5497,10 @@ def calculateLoopNumIter(self, kernel, loopIdx, isPap):
       kStr += inst("s_mov_b32", sgpr("OrigLoopCounter"), \
                 loopCounter, \
                 "copy loop counter")
-      if self.prefetchAcrossPersistent0 and kernel["ExpandPointerSwap"] and isPap:
-        kStr += inst("s_and_b32", sgpr("EvenIterStart"), sgpr("OrigLoopCounter"), \
-                  0x1,
-                  "save unroll loop start position - copy1 or copy2")
+      # calculate once and save: will this problem size exit at oddIter or evenIter?
```
AlexBrownAMD, line 5500: a little confused by the new comment (and isPap check switched to "not")

AlexBrownAMD (approved):
Added a few suggestions for code maintenance

Tensile/Source/lib/source/ContractionProblem.cpp
solaslin: Thx! agree on this.

Tensile/SolutionStructs.py
solaslin: Thx! I will use print2

Tensile/SolutionStructs.py
solaslin: Same as above

Tensile/KernelWriter.py
solaslin: Actually not, since prefetchAcrossPersistent + EPS was not supported before, so this "and" was always evaluated as False (0). Thus the range(0,+1) = range(0,1),

Tensile/KernelWriterAssembly.py
solaslin: Thanks for spotting this. 
I also found that the old code is wrong for getting WGIJSerial = stmp * NumWorkGroups0 * NumWorkGroups0.

Tensile/Source/lib/include/Tensile/ContractionProblemPredicates.hpp
```diff
@@ -982,7 +982,8 @@ namespace Tensile
 
                 virtual bool operator()(ContractionProblem const& problem) const override
                 {
-                    return problem.getPersistentKernelEligibility();
+                    // if value = false -> turn off check, always return eligibility as true
+                    return value ? problem.getPersistentKernelEligibility() : true;
```
cgmillette, line 986: This change I would revert. A predicate should only be attached to a solution if necessary.

In the case that value = false, (don't care about PK eligibility) the predicate should not be attached anyway because it always returns true. Additional predictable check that costs time.

Only in the case that value = true (we care about PK eligibility) it should be attached to always check the problem. Not easily predictable but necessary time cost.

In fact, I don't think this predicate should have a value at all since this would indicate the same as attached or not.

Tensile/Source/lib/source/ContractionSolution.cpp
```diff
@@ -308,16 +308,31 @@ namespace Tensile
...
             }
-            rv.numWorkGroups.x = std::min(persistentGroups, problemGroups);
-            rv.numWorkGroups.y = 1;
+
+            if(finalPKValue == -1)
+            {
+                // 1. Get the largest pk value (ex.3)
+                //    which can make the PK.G (ex.3*120=360) <= problemGroups (ex.433)
+                // 2. Scale by 5/8 (can try 0.5~1, to control the tiles-per-workgroup = 1~2)
+                finalPKValue = 5 * (problemGroups / cuCount) / 8;
```
cgmillette, line 328: Small risk, but consider an assertion for cuCount != 0

Tensile/Contractions.py
```diff
@@ -327,6 +327,8 @@ def CompoundPredicates(cls, state, problemType):
 
         if 'PersistentKernel' in state and state['PersistentKernel']:
             rv += [cls("PersistentKernelCheck", value = True)]
+            # debugging: set value = false to allow the problem runnable with PK
```
cgmillette, line 330: See comment on the predicate:
- If wanting to debug here, just don't attach the predicate.
We can probably remove the value.

solaslin: Yes, I totally agree with you. That's also what I did before. (Not to add predict at all) I remember there was some reason why I changed it, but I can not recall...I'll change back to the better one :)

Tensile/Source/lib/source/ContractionSolution.cpp
solaslin: Yes, I've thought about this when I did this part. 
However, the function "checkPersistentKernelEligibility" is not called int the path of rocBLAS side. 
(Also, we want to allow the tuned logic always runable in rocBLAS, since if it was tuned in Tensile, we have no reason to block it from running in rocBLAS). So it turns out I ro-do the PKn1-eval to get the final PKValue in case we wouldn't have done it (checkEligibility) in rocBLAS.
But I agree on make this part more elegant, maybe will re-design this in the next PR.

Tensile/Source/lib/include/Tensile/ContractionProblemPredicates.hpp
solaslin: Reply together below (Contractions.py)

Tensile/Source/lib/source/ContractionSolution.cpp
solaslin: Done !

PR #1214, opened 2020-11-03
Instruction scheduling to reduce LDS by half in TN problems

rosenrodt:
Proposed a scheduling technique that reduces LDS usage by half, while maintaining the same DepthU in TN problems.

## Goal
The initial goal is to increase global read efficiency of TN-layout problems by allowing deeper DepthU (ideally DepthU=128 for half types) kernels, which are often LDS limited.

## Result
When enabled, code gen saves 50% of LDS usage (in addition to savings from enable single LDS buffer 1LDSBuffer).  Actual tuning suggests increasing to occupancy 2 may be a better utilization of overall resources. For example, MT256x128x64 occupancy 2 rather than MT256x128x128 occupancy 1. 

## Implementation
By splitting the in-flight data (from global memory to LDS) held by registers into 2 portions, kernel writes one portion into LDS at any given time. Allowing for deeper DepthU (eg MT128x128x128) or higher occupancy.

To this end, the implementation separates the concept of DepthU: that DepthU determines the global read depth in the same old sense, while DepthULDS determines the local write depth. 

DepthULDS is not directly exposed as tuning parameter (defined internally). Rather it's DepthULdsDivisor such that DepthULds * DepthULdsDivisor = DepthU. Default value is 1, meaning DepthU equals DepthULDS. When value is 2, then DepthULds is 1/2 of DepthU.

Sometimes this feature is referred to as "SplitLDS" informally in some part of the code. But I am not sure whether the name is accurate/appropriate.

## Local test results
| Test item \ Arch | gfx906 | gfx908 |
| -- | -- | -- | 
| pre_checkin* | Passed | Passed |
| extended | Passed | Passed |
| rocblas-test | Passed | Passed |

*Added 2 more test yamls to pre_checkin. Time span increased by ~2min after this patch (27m30s -> 29m37s):

### DIFF
--- a/Tensile/Common.py
+++ b/Tensile/Common.py
@@ -876,6 +876,18 @@ validParameters = {
     # -3 : Only allow min(GLVWA,GLVWB) < VW ?
     "DepthU":                     depthUs,
 
+    # DepthULdsDivisor determines how we pipeline the data from global memory to LDS
+    # Instead of moving all in-flight data from the register buffer (G2L) to the LDS at once, we divide the G2L buffer into N portions and
+    # write each portion of the G2L to LDS, read from LDS and do the actual matrix multiply-accumulate, before moving on to the portion and so on.
+    # This helps cut down LDS usage by the value of the divisor. Helps increase CU occupancy or DepthU if kernel was previously LDS limited.
+    #
+    # The premise of this parameter is to be able to fetch all 256B (equivalent to 128 half's or 64 single's) in a TN laid-out problem size,
+    # maximizing L2 channel efficiency, therefore this parameter only works for TN buffer layout
+    #
+    # Implementation-wise, for now it only supports ScheduleIterAlg=3 and TransposeLDS=1
+    # In addition, it does not work with DirectToLds=1 because it needs the in-flight data to reside in registers
+    "DepthULdsDivisor":           [1, 2], # [4, 8] not tested yet
+
     # integer ammount of padding to put into LDS, in 2016 this didn't seem to help performance, profilers were showing that channel conflicts weren't really hurting
     # performance so this has been deprecated and probably doesn't work
     # -1 means use same padding as the VectorWidth if TLU=0 else 0.  (Padding only helps when transpose is required)
@@ -1038,6 +1050,7 @@ defaultBenchmarkCommonParameters = [
     {"DisableAtomicFail":         [ 0 ] },
     {"DisableKernelPieces":       [ 0 ] },
     {"DepthU":                    [ -1 ] },
+    {"DepthULdsDivisor":          [ 1 ] },
     {"PerformanceSyncLocation":   [ -1 ] },
     {"PerformanceWaitLocation":   [ -1 ] },
     {"PerformanceWaitCount":      [ -1 ] },
--- a/Tensile/KernelWriter.py
+++ b/Tensile/KernelWriter.py
@@ -1271,11 +1271,14 @@ class KernelWriter(metaclass=abc.ABCMeta):
         else:
           self.localWriteACode = Code.Module()
           self.localWriteBCode = Code.Module()
-        if kernel["PrefetchGlobalRead"] and kernel["LoopIters"] == 1 and uDu > 0:
+
+        # TODO schedule waitcnt/barrier in makeSubIterSchedule()
+        if kernel["PrefetchGlobalRead"] and kernel["LoopIters"] in [1, 2] and uDu > 0:
           if self.enable["Wait"]:
             kl.append(self.wait(kernel, tensorParametersA, tensorParametersB, 1, 0, -1, "wait for local write"))
           if self.enable["Sync"]:
             kl.append(self.syncThreads(kernel, "sync for local read after write"))
+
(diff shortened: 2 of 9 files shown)


Tensile/KernelWriter.py
```diff
@@ -222,16 +234,24 @@ def makeSchedule(self, kernel, tensorParametersA, tensorParametersB, localWriteE
...
       for i in range(numEmptyGlobalReadIncCode):
-        itemsGRToSched.append(Code.Module())
+        imod = Code.Module()
+        imod.addCode(Code.Inst("// ", "globalReadIncA scheduler placeholder"))
+        itemsGRToSched.append(imod)
       itemsGRToSched.append(globalReadIncBCode)
       for i in range(numEmptyGlobalReadIncCode):
-        itemsGRToSched.append(Code.Module())
+        imod = Code.Module()
+        imod.addCode(Code.Inst("// ", "globalReadIncB scheduler placeholder"))
```
aazz44ss, line 243: can use addcomment0()

Tensile/KernelWriter.py
```diff
@@ -1224,7 +1295,9 @@ def noLoadLoop( self, kernel, tensorParametersA, tensorParametersB, isOptNLL, is
           isSwapAndResetLwoIter = (u == self.lwEndMfmaIndex//(self.numMfmaPerIter))
 
       extraComment = ""
-      if isNGLL:
+      if isLastLoop:
+        extraComment += " (last unrolled loop)"
+      if not isLastLoop:
```
aazz44ss, line 1300: if isLastLoop:
    ....
else:
    ....

Tensile/KernelWriter.py
rosenrodt: The placeholder needs to be `Code.Inst` instead of `Code.TextBlock` so it can have an effect on scheduling decisions

Tensile/Common.py
```diff
@@ -876,6 +876,18 @@
...
+    # Instead of moving all in-flight data from the register buffer (G2L) to the LDS at once, we divide the G2L buffer into N portions and
+    # write each portion of the G2L to LDS, read from LDS and do the actual matrix multiply-accumulate, before moving on to the portion and so on.
+    # This helps cut down LDS usage by the value of the divisor. Helps increase CU occupancy or DepthU if kernel was previously LDS limited.
+    #
+    # The premise of this parameter is to be able to fetch all 256B (equivalent to 128 half's or 64 single's) in a TN laid-out problem size,
+    # maximizing L2 channel efficiency, therefore this parameter only works for TN buffer layout
+    #
+    # Implementation-wise, for now it only supports ScheduleIterAlg=3 and TransposeLDS=1
+    # In addition, it does not work with DirectToLds=1 because it needs the in-flight data to reside in registers
+    "DepthULdsDivisor":           [1, 2], # [4, 8] not tested yet
```
ramjana, line 889: other usage model for this feature is to fetch larger DepthU or MT than LDS size permits. do we have register limit  + LDS size check implemented that checks total size required < 128 KB.

Tensile/KernelWriter.py
```diff
@@ -314,22 +335,31 @@ def makeSchedule(self, kernel, tensorParametersA, tensorParametersB, localWriteE
...
-        # Scheduler below keeps all writes in the same module in same iteration
-        # so this is better match to what it is trying to do
-        writesToSched = sum(1 for item in itemsLWToSched if item.countType(Code.LocalWriteInst))
-      else:
-        # count the number of writes, this doesn't match how they are
-        # scheduled so pushes writes up too far
-        writesToSched = self.localWriteACode.countType(Code.LocalWriteInst) + \
-                     self.localWriteBCode.countType(Code.LocalWriteInst)
+      # piggyback each global read instruction after each local write instruction to maximize G2L buffer utilization
+      if kernel["PrefetchGlobalRead"] == 2 or kernel["DepthULdsDivisor"]>1:
```
ramjana, line 339: recommend not to interleave longer latency memory instructions with shorter memory latency instructions. you really want to issue longer memory instructions b2b and issue all of them for better memory pipelining efficiency

ramjana (approved):

Tensile/Common.py
rosenrodt: It's broken into same 2 separate checks as it was before: 

* Register limit are checked in `KernelWriter::functionSuffix()` due to the dynamics of temporary use of sgpr/vgprs.
* LDS limit is checked in `Solution` constructor, where the LDS is evaluated based on the internal value `DepthULds`

For example, it is possible to generate MT256x128x128 hgemm kernel w/ `DepthULdsDivisor=2`. The unrolled loop would need only (DepthU/2)*(256+128)*bpe ~= 49K LDS size (excluding LDS padding).

Tensile/KernelWriter.py
rosenrodt: Will try that in the ensuing optimization work

aazz44ss (approved):
LGTM

rosenrodt:
Merging since all CI passed + rocblas-test passed

PR #1225, opened 2020-11-12
add merge=union gitattributes for CHANGELOG.md

zaliu:
- My thank to @cgmb for providing this change
- This update reduces merge conflicts if there are multiple concurrent edits to CHANGELOG.md

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,9 @@
+/CHANGELOG.md merge=union
+*.pdf binary
+*.doc binary
+*.docx binary
+*.ppt binary
+*.pptx binary
+*.xls binary
+*.xlsx binary
+*.xps binary


.gitattributes
```diff
@@ -0,0 +1,9 @@
+/CHANGELOG.md merge=union
+*.pdf binary
+*.doc binary
+*.docx binary
+*.ppt binary
+*.pptx binary
+*.xls binary
+*.xlsx binary
+*.xps binary
```
cgmb, line 9: I would probably just delete lines 2-9. They have no effect on the Tensile repository, as you don't have any of those file types.

That said, it's no big deal either way.

zaliu: > I would probably just delete lines 2-9. They have no effect on the Tensile repository, as you don't have any of those file types.

Who knows where future lies?  Thanks for the comments though.

cgmb (approved):

cgmb:
I included some workflow guidelines for this change in a comment here: https://github.com/ROCmSoftwarePlatform/rocBLAS-internal/pull/429#issuecomment-726310516

The important one is: make sure that your only conflicts within the changelog occur due to adding changelog entries, as this strategy will not correctly handle two people fixing the same typo in old changelog entries.

PR #1227, opened 2020-11-14
Hotfix for PR #1214: count the right instruction type under different scheduling

rosenrodt:
Fixed instruction scheduler regression.  rocblas-test passed when built against latest rocBLAS logic files. 

Regression due to recent refactor to do scheduling depending on the information provided by the instruction type. Since some of the codes emitted were still in form of raw string instead of `Code.Inst`, they were effectively ignored by the scheduler and therefore failed to generate valid codes


aazz44ss (approved):
LGTM

zaliu (approved):
Thanks so much for the quick fix!  Please merge.

PR #1239, opened 2020-11-26
Adding tuning client design doc.

sdquiring:
This is a first draft, feel free to add to it or suggest areas where it could be filled in more or otherwise improved.

### DIFF
--- a/Tensile/Source/client/include/DataInitializationTyped.hpp
+++ b/Tensile/Source/client/include/DataInitializationTyped.hpp
@@ -2,7 +2,7 @@
  *
  * MIT License
  *
- * Copyright 2019-2020 Advanced Micro Devices, Inc.
+ * Copyright 2019-2021 Advanced Micro Devices, Inc.
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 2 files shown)


solaslin (commented):
I have a few suggestion maybe we can add. 

* **Predicates**
What are predicates doing, and how to add a new predicate.

* The **InitMode** of DataInitialization
Problem-independent vs Problem-dependent, e.g what is the use case for Trig__ ? I also updated some comments in PR #1240 for new client (original comments are outdated)

* Difference between **ContractionInputs_X_X_X** (in ContractionProblem.hpp) and **ManaganedContractionInputs_X_X_X** (in DataInitializationTyped.hpp)
ContractionInputs_X_X_X is for kernels, and ManaganedContractionInputs_X_X_X is used for validator (CPU reference)

* **AllSolutionIterator** vs **BestSolutionIterator**
AllSolutionIterator is used by default , and it doesn't call library->findBestSolution()
What is the use case for BestSolutionIterator, and how to enable it when tuning?

cgmillette:
- It might be worth mentioning how ClientParameters.ini files are generated and fit into the whole workflow.
- A little while ago we also discussed the data-initialization of A/B/C inputs was balanced between positive and negative numbers to make sure we don't introduce too much of a bias such that validation accuracy becomes an issue.
- May be too specific but could offer that the data allocations for A/B/C/D are based on max sizes for all problems in the run, and they are allocated only once at the beginning.
- When we do validation, data is copied from device to host, which can drastically affect runtime of overall process.

cgmillette:
>   AllSolutionIterator is used by default , and it doesn't call library->findBestSolution()
>   What is the use case for BestSolutionIterator, and how to enable it when tuning?



The goal of the tuning process is not discriminate between any of the valid kernels such that we can run them all and observe their performance. 'Best solution' inside tuning has no effect because there is no data on 'what is best' yet. We want to run all kernels (possibly several times) to increase our chances of accurately determining what is best.

Only after tuning is complete that we generate the solution library with the best performing kernels given the tuned sizes. Outside of tuning, consumers of Tensile default to the 'best solution' because they will only want the best kernel to solve their problem.

Predicates, however will be enforced inside and out of Tuning. These are hard constraints on whether a kernel can even produce a correct solution given the problem description and available resources. In tuning we may see "DID_NOT_SATISFY_ASSERTS" or something similar for a kernel selected for a problem it cannot solve on that particular hardware.

zaliu (approved):

PR #1240, opened 2020-11-27
Fix the misleading/outdated comments for DataInitType

solaslin:
Comments explaining the **DataInitType** are outdated and could be mis-leading since **new-client** is different from old-client.
Just update some new comments to clarify the meaning.

Adding this PR due to #1239 , this should help
NoCI since this is a pure comment-editing

### DIFF
--- a/Tensile/Common.py
+++ b/Tensile/Common.py
@@ -136,6 +136,13 @@ globalParameters["CSVMergeSameProblemID"] = False
 # trig_float initializes with the sin function to have non-zero values in the mantissa
 # and exponent. It cannot be used for int8 or int32. Need to use tensileAlmostEqual
 # not tensileEqual for checking the result.
+# NOTE- the following comments explaining the DataInitType are for OldClient and should be obsolete.
+#       For NewClient, please read ClientWriter.py, the DataInitName(Enum)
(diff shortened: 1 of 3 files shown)


solaslin:
The CI failed due to the NoCI tag

Tensile/Common.py
```diff
@@ -136,13 +136,18 @@
 # trig_float initializes with the sin function to have non-zero values in the mantissa
 # and exponent. It cannot be used for int8 or int32. Need to use tensileAlmostEqual
 # not tensileEqual for checking the result.
+# NOTE- the following comments explaining the DataInitType are for OldClient and should be obsolete.
```
rosenrodt, line 139: I think maybe the comments for alpha/beta type still holds correct?

solaslin: alpha / beta , the old comments (0~4) holds correct for new client as well.
But in new client, you can also set Alpha/Beta type to 5,6,7,16 (Inf, BadIn, BadOut, rand-narrow) and it works in host
BTW, if you set Alpha/Beta to 8~15, (problem-dept.), client writer doesn't show any warning,
but host-client will send a "Invalid InitMode" exception.

rosenrodt: We may simply make clear the acceptable values for A/B/C/D & Alpha/Beta from end user's perspective

solaslin: Updated

rosenrodt (approved):
Looks good now. Thanks

PR #1243, opened 2020-12-03
Fix configuration: ShortNames

rosenrodt:
Depends on #1242

- Fix global parameter "ShortNames"; brought to you by @aazz44ss
- Remove `xfail` status of TensileCreateLibrary integration test involving the said parameter

### DIFF
--- a/Tensile/KernelWriter.py
+++ b/Tensile/KernelWriter.py
@@ -3483,14 +3483,14 @@ class KernelWriter(metaclass=abc.ABCMeta):
   # get kernel name
   ##############################################################################
   def getKernelFileBase(self, kernel):
-    rv = self.getKernelName(kernel)
-    return self.shortenFileBase(rv)
+    if globalParameters["ShortNames"]:
+      fileBase = Solution.getNameSerial(kernel, self.kernelSerialNaming)
+    else:
+      fileBase = self.shortenFileBase(kernel)
+    return fileBase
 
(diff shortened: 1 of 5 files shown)


aazz44ss (approved):
Old:
"shortNames" will modify file name and kernel name to SerialName.
it will has failure when doing TensileCreateLibrary.
and also hard for developer to search kernel.

Now:
"shortNames" will only modify file name to SerialName.

Tensile/KernelWriter.py
```diff
@@ -3596,7 +3592,10 @@ def getReplacementKernelPath(self, kernel):
     kernelName = self.getKernelName(kernel)
     return ReplacementKernels.Get(kernelName)
 
-  def shortenFileBase(self, base):
+  def shortenFileBase(self, kernel):
```
sdquiring, line 3595: This function's name no longer makes sense.  It should still just shorten the input.  The serial name should be implemented in `getKernelFileBase()` or in a function called from there.

sdquiring (changes requested):

Tensile/KernelWriter.py
aazz44ss: move ShortNames to getKernelFileBase()

sdquiring (approved):

cgmillette (approved):
Looks good to me

PR #1245, opened 2020-12-03
Framework for filtering solutions

benjaminulmer:
Added framework for `AllSolutionsIterator` to filter which solutions to run based on the projected performance. The filtering criteria are implemented as a `std::vector<std::function<bool(ProjectedPerformance const&)>>`, which is provided to the iterator at construction.

Currently there is support for a granularity threshold criteria, which ensures solutions are only run if they meet a minimum total granularity.


sdquiring (changes requested):
Good overall design.  I think we need to be a bit more flexible though. I'd like to be able to filter on other properties of the solution (like macro tile).  I think the function should take the problem, the hardware, and the solution.

sdquiring (approved):

PR #1246, opened 2020-12-08
Tensile cluster tuning tool

cgmillette:
### TensileBenchmarkCluster classes
**Motivation:** Tensile tuning on a single machine can take a very long time. We can split it up between machines, but managing this overhead is tedious and manual. Cluster management systems such as SLURM can take care of the overhead for us, so we need a tool to help us communicate with the cluster.

**Workflow:** 
- Split benchmark config into smaller 'task' sized units. For now 'task size' is defined as a subset of the 'ProblemSizes' section for each benchmark group. (Init)
- Generate binary executables that the cluster will invoke for each task. (Deploy)
- Submit all tasks to cluster. (Run)
- Combine all the resulting logic files from each task into the final logic file. (Results)

TensileBenchmarkCluster script facilitates benchmark splitting and result logic merging (Init and Results), while delegating generation of benchmarking executables and task submission to the cluster backend class (Deploy and Run).

For Tensile we use a SLURM cluster, so the backend class BenchmarkImplSLURM is provided as a default:
- Builds a docker image with the tensile_client executable and run script as the docker target. (Deploy)
- Creates bash scripts that will load the docker image, mount directories for the task and invoke the docker target. (Deploy)
- Drops a series of bash scripts to submit batch job requests to SLURM (currently limiting resources of 1 Node per task) (Deploy)
- Invokes tasks on the cluster (Run)

TensileBenchmarkCluster also uses a new configuration class that will not be using the globalParameters.

### Configuration classes:
**Motivation:** Current Tensile configuration objects are very lenient, and somewhat tedious to maintain. There are different dicts with identical keys to represent current state, default state, or ranges of acceptable values with descriptions. Each time a new parameter is added, each of these dicts need to be updated, which can be error prone if any of them are forgotten. It is also difficult to maintain the dependencies of these parameters where the constraints sometimes difficult to understand and apply. 

Naming can also become an issue because most of the parameters live in the base level of the config. Long and awkward names are sometimes given to illustrate context and make them unique, whereas you could tier them in a more natural and contextual way. Moreover, full config hierarchies are often not necessary for certain parts of the code, so it may be beneficial to share sub-configs to objects who only care about those parameters.

Lastly, dictionary assignment is a bit hazardous due to immediate item creation when the key does not yet exist. This is somewhat error prone when we spell an attribute incorrectly:
`config["MyAttrbute"] = 10 # Oops! I meant config["MyAttribute"] = 10`
Now the config would have an entry MyAttrbute storing my intended value, as MyAttribute now stores an unintended value.

ProjectConfig helps solve some of these problems:

- Configuration objects encapsulate attributes such as a name, value, default value and a description. These have explicit creation methods that help encapsulate configuration parameter attributes, and reduces the chances of introducing bugs due to spelling mistakes.

```
networkConfig = config.createSection("Network")
networkConfig.createValue("IP", value=x.x.x.x, defaultValue=y.y.y.y, description="The IP we want to connect to")
networkConfig.createValue("PORT", value=xxxx, defaultValue=yyyy, description="The IP port we want to connect to")
networkConfig.createValue("MAX_PORT", value=xxxx, defaultValue=yyyy, description="The max port number we wish to use")

config["Netwrk"]["IP"] = 10 # Will fail as Netwrk has not been explicitly created.

print(networkConfig.getDescription("IP")) # Get user description of the parameter
print(networkConfig.getDefault("IP"))        # Get the default value of the parameter
```

- Configurations can have multiple tiers of sub-sections which support attribute style syntax, dict style syntax and flat dict style syntax:

```
# Attribute style
config.Network.IP = X.X.X.X
config.Network.PORT = XXXX
config.Network.MAX_PORT = YYYY

# Dict style
config["Network"]["IP"] = X.X.X.X
config[["Network"]["PORT"] = XXXX
config[["Network"]["MAX_PORT"] = YYYY

# Flat dict style
config["Network.IP"]  = X.X.X.X
config["Network.PORT"]  = XXXX
config["Network.MAX_PORT"]  = YYYY
```

- Configurations also support constraints for implementing configuration assertions or dependencies as expressions:

```
# assertion
config.addConstraint("Network.PORT <= Network.MAX_PORT")

# dependency assignment
config.addConstraint("Network.PORT = min(Network.MAX_PORT, Network.PORT)")
```

### DIFF
--- a/Tensile/TensileBenchmarkCluster.py
+++ b/Tensile/TensileBenchmarkCluster.py
@@ -7,20 +7,21 @@ if __name__ == "__main__":
 import shlex, subprocess
 import sys
 import os
-import re
-import itertools
-import datetime
-import time
-import copy
-import math
 import argparse
-import mgzip
+
 from .BenchmarkSplitter import BenchmarkSplitter
 from .Configuration import ProjectConfig
-from .Common import HR
 from Tensile.Utilities.merge import mergePartialLogics
 
-import yaml
+try:
+    import mgzip
+except ImportError:
+    print("Package mgzip not found: docker zipping will be slow. Install pip3 install mgzip to improve performance.")
+
+    # Fallback package import
+    import gzip
+
+
 
 class ScriptHelper(object):
     """
@@ -412,8 +413,6 @@ class BenchmarkImplSLURM(object):
         with open(os.path.join(logDir, "dockerBuildLog.log"), 'w') as logFile:
 
             # Docker build command
-            dockerFileName = os.path.basename(dockerFilePath)
-            dockerFileDir = os.path.dirname(dockerFilePath)
             buildCmd = str("\
                 docker build \
                 -t {0} \
@@ -439,9 +438,17 @@ class BenchmarkImplSLURM(object):
             # Docker will save .tar binary to stdout as long as it's not attached to console.
             # Pipe stdout into mgzip to get smaller .tar.gz archive
             print("Saving docker image: {0}  to {1} ...".format(tag, archivePath))
-            with mgzip.open(archivePath, 'wb') as zipFile:
-                with subprocess.Popen(shlex.split(saveCmd), stdout=subprocess.PIPE, stderr=logFile) as proc:
-                    zipFile.write(proc.stdout.read())
+
+            try:
+                with mgzip.open(archivePath, 'wb') as zipFile:
+                    with subprocess.Popen(shlex.split(saveCmd), stdout=subprocess.PIPE, stderr=logFile) as proc:
+                        zipFile.write(proc.stdout.read())
+            except NameError:
+                print("mgzip not available: Install with pip3 install mgzip. Falling back to slow single threaded gzip...")
+                with gzip.open(archivePath, 'wb') as zipFile:
+                    with subprocess.Popen(shlex.split(saveCmd), stdout=subprocess.PIPE, stderr=logFile) as proc:
+                        zipFile.write(proc.stdout.read())
+
             print("Done saving docker image!")
 
     @staticmethod
--- a/Tensile/TensileBenchmarkCluster.py
+++ b/Tensile/TensileBenchmarkCluster.py
@@ -3,7 +3,7 @@ if __name__ == "__main__":
     print("This file can no longer be run as a script.  Run 'Tensile/bin/TensileBenchmarkCluster' instead.")
     exit(1)
 
-import yaml
+
 import shlex, subprocess
 import sys
 import os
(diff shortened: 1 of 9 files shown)


Tensile/TensileBenchmarkCluster.py
```diff
@@ -0,0 +1,846 @@
...
+    def genEcho(text):
+        return ScriptHelper.genLine(str("echo \"{0}\";").format(str(text)))
+
+    @staticmethod
+    def makeExecutable(path):
+        mode = os.stat(path).st_mode
+        mode |= (mode & 0o444) >> 2    # copy R bits to X
+        os.chmod(path, mode)
+
+class ScriptWriter(object):
```
sdquiring, line 50: These should be separated either into a separate file or into utility.py.

Tensile/TensileBenchmarkCluster.py
```diff
@@ -0,0 +1,846 @@
...
+
+
+
+class ScriptHelper(object):
+    """
+    Helper class to facilitate formatting when
+    writing bash scripts.
+    """
+
+    @staticmethod
```
sdquiring, line 32: `@classmethod` is preferred to `@staticmethod`.

Tensile/TensileBenchmarkCluster.py
```diff
@@ -0,0 +1,846 @@
...
+
+class BenchmarkImplSLURM(object):
+
+    # baseImage: remote URL to rocm build
+    # dockerFilePath: path to docker file for slurm tensile build
+    # tag: name:TAG of new image
+    # outDir: where to export the new image
+    # tensile<Fork/Branch/Commit>: tensile code to use on github
+    @staticmethod
+    def __createTensileBenchmarkContainer(baseImage, dockerFilePath, tag, outDir, logDir, tensileFork, tensileBranch, tensileCommit):
```
sdquiring, line 405: There's considerable effort here to abstract away Slurm but Docker is implicitly used everywhere.  Getting images from one machine to another and starting containers is often handled implicitly by the scheduler.

sdquiring, line 405: Once a function has this many parameters, we should consider instance variables instead of individual parameters.

Tensile/TensileBenchmarkCluster.py
```diff
@@ -0,0 +1,846 @@
...
+import sys
+import os
+import argparse
+
+from .BenchmarkSplitter import BenchmarkSplitter
+from .Configuration import ProjectConfig
+from Tensile.Utilities.merge import mergePartialLogics
+
+try:
+    import mgzip
```
sdquiring, line 17: `mgzip` seems to have the same interface as `gzip`.  This means we should be able to do:

``` python
        import mgzip as gzip
```

That way the usage below can be common to both libraries.  See how this is implemented with `tqdm` in `Utils.py`.

Tensile/TensileBenchmarkCluster.py
```diff
@@ -0,0 +1,846 @@
+
+if __name__ == "__main__":
```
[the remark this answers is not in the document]
cgmillette: Done

Tensile/TensileBenchmarkCluster.py
cgmillette: Nice tip! Will implement.

sdquiring (approved):

PR #1250, opened 2020-12-11
Make messagepack python dependency optional

JeremyAdamHart:
Allows Tensile to use the yaml library backend without having the python msgpack library. Also includes a small fix to set TENSILE_USE_MSGPACK when running:
```Tensile/bin/Tensile --library-format msgpack```

### DIFF
--- a/Tensile/LibraryIO.py
+++ b/Tensile/LibraryIO.py
@@ -34,7 +34,7 @@ except ImportError:
 try:
   import msgpack
 except ImportError:
-  printExit("You must install MessagePack for Python to use Tensile (to parse config files). See https://github.com/msgpack/msgpack-python for installation instructions.")
+  print("Message pack python library not detected. Must use YAML backend instead.")
 
 ################################################################################
 # Read Benchmark Config from YAML Files
@@ -292,7 +292,11 @@ class YAMLWriter(Writer):
 class MessagePackWriter(Writer):
   def write(self, filename, data):
     with open(filename, 'wb') as f:
+      try:
         msgpack.pack(data, f)
+      except NameError:
+        printExit("You must install MessagePack for Python to use Tensile (to parse config files). See https://github.com/msgpack/msgpack-python for installation instructions.") 
+
 
   def writeLibraryLogicForSchedule(self, filePath, schedulePrefix, architectureName, \
       deviceNames, logicTuple):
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: Tensile/LibraryIO.py [hunk 1]
```diff
@@ -34,7 +34,7 @@
 try:
   import msgpack
 except ImportError:
-  printExit("You must install MessagePack for Python to use Tensile (to parse config files). See https://github.com/msgpack/msgpack-python for installation instructions.")
+  print("Message pack python library not detected. Must use YAML backend instead.")
```
cgmillette, line 37: Instead of "Must use YAML..."
Might be good idea to have "Falling back to YAML..."

Tensile/Source/CMakeLists.txt
```diff
@@ -32,6 +32,9 @@ if(TENSILE_NEW_CLIENT)
 
     set(TENSILE_USE_HIP      ON CACHE BOOL "Use the Hip runtime.")
     set(TENSILE_USE_LLVM     ON CACHE BOOL "Use LLVM for parsing config files.")
+    if(Tensile_LIBRARY_FORMAT MATCHES "msgpack")
+        set(TENSILE_USE_MSGPACK  ON CACHE BOOL "Use message pack for parsing config files.")
+    endif()
     set(TENSILE_USE_OPENMP   ON CACHE BOOL "Use LLVM for parsing config files.")
```
cgmillette, line 38: Can you update the message for TENSILE_USE_OPENMP while we are at it? :)

Tensile/Source/CMakeLists.txt
```diff
@@ -32,6 +32,9 @@ if(TENSILE_NEW_CLIENT)
 
     set(TENSILE_USE_HIP      ON CACHE BOOL "Use the Hip runtime.")
     set(TENSILE_USE_LLVM     ON CACHE BOOL "Use LLVM for parsing config files.")
+    if(Tensile_LIBRARY_FORMAT MATCHES "msgpack")
+        set(TENSILE_USE_MSGPACK  ON CACHE BOOL "Use message pack for parsing config files.")
+    endif()
```
cgmillette, line 37: Is it possible to switch TENSILE_USE_LLVM = OFF in this case and still have things work correctly? As far as exclusivity goes, that would then allow us to use msgpack and not depend at all on llvm.

Tensile/Source/CMakeLists.txt
JeremyAdamHart: Done!

Tensile/Source/CMakeLists.txt
JeremyAdamHart: Good suggestion. I made a couple small changes and that works.

cgmillette (approved):
LGTM!

PR #1254, opened 2021-01-07
Require OpenMP when TENSILE_USE_OPENMP=On

haampie:
If the user asks for `TENSILE_USE_OPENMP`, *require* it in the `find_package(OpenMP ..` bit.

This way not finding OpenMP when requested actually results in a hard configuration error, instead of an annoying linking error.

Also use the "modern" way to make targets use OpenMP, which requires CMake 3.9. This should really not be a problem, because
LLVM requries CMake >=3.13.4 already, and Tensile depends on a build of llvm-amdgpu anyways.


TorreZuk (commented):
This should be tested on all supported platforms which I don't see in the CI.
When I made similar changes in rocBLAS I was asked to revert them as it caused some tensile issues, I am waiting to try again to update to the simpler style.
the compile time flags for OpenMP must also be set and this PR removes them.  The CXX flags should be set for targets to avoid rocm-cmake warnings.   These compile and link flags may be problematic on some platforms even using the "newer" style with 3.9 so again need to be tested on all OS: sles, centos, ubuntu, ...

haampie:
Hi @TorreZuk, thanks for the review! Can I be of any help to that? Or do you have access to the supported systems?

haampie:
Ah, I guess you were talking about this commit? https://github.com/ROCmSoftwarePlatform/rocBLAS/commit/0f45c2827ac59159dfb6a955735b6784f278caa0. Makes me happy to see all those absolute paths to system libs and such go away, keep up the good work! :)

TorreZuk:
HI @haampie yes I agree that is how we would like the OpenMP to be fetched but unfortunately on some OS this would fail until very recently; these failure cases should now be able to find the OpenMP bundled with llvm/hipclang.  Wasn't a cmake issue specifically but a hipclang cmake openmp issue.   I can't help on Tensile PR testing but I am now trying to test again on all OS in hopefully an equivalent form in rocBLAS (new PR but tweak of that commit for cmake 3.9).   If this passes it should also cover your use case.

haampie:
Thank you, that's very helpful :)

TorreZuk (commented):
The current toolchain appears to fail using this linker only dependency target for OpenMP on sles, and centos/redhat 7.   This slightly older style appears to work:
target_compile_options(test_openmp_helloworld PRIVATE ${OpenMP_CXX_FLAGS})
target_link_libraries(test_openmp_helloworld PRIVATE hip::device ${OpenMP_CXX_FLAGS})
Using the OpenMP::OpenMP_CXX linker target only form requires adding runtime linking target information:
  if(OS_ID_rhel OR OS_ID_centos OR OS_ID_sles)
        use any cmake target link advice:    -L${HIP_CLANG_ROOT}/lib -Wl,-rpath=${HIP_CLANG_ROOT}/lib
  endif()
  Additionally we have an internal toolchain (which may still be required) that fails to find_package OpenMP so this may be a constraint for Tensile for the very near future.   I leave it to the Tensile team to assess this further, hopefully we will move to one of these newer forms soon.  Thanks for providing your motivation on this.

zaliu:
Just a quick status of our internal testing: rocBLAS build with this pull request **failed** in an HMM environment on Ubuntu.  We do not yet understand why this is the case yet.  The rocBLAS build log shows:
```
...
Successfully installed Tensile-4.26.0
-- using local Tensile from /home/zaliu/repos/Tensile-develop, copied to
-- Adding /home/zaliu/repos/rocBLAS-develop/build-2021-02-02-1254/release/virtualenv to CMAKE_PREFIX_PATH
-- The C compiler identification is Clang 12.0.0
-- Check for working C compiler: /opt/rocm/bin/hipcc
-- Check for working C compiler: /opt/rocm/bin/hipcc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- ROCclr at /opt/rocm/lib/cmake/rocclr
-- Performing Test HIP_CLANG_SUPPORTS_PARALLEL_JOBS
-- Performing Test HIP_CLANG_SUPPORTS_PARALLEL_JOBS - Success
-- Found OpenMP_C: -fopenmp=libomp
CMake Error at /usr/share/cmake-3.10/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
  Could NOT find OpenMP_CXX (missing: OpenMP_CXX_FLAGS OpenMP_CXX_LIB_NAMES)
Call Stack (most recent call first):
  /usr/share/cmake-3.10/Modules/FindPackageHandleStandardArgs.cmake:378 (_FPHSA_FAILURE_MESSAGE)
  /usr/share/cmake-3.10/Modules/FindOpenMP.cmake:447 (find_package_handle_standard_args)
  build-2021-02-02-1254/release/virtualenv/lib/python3.6/site-packages/Tensile/Source/CMakeLists.txt:68 (find_package)
...
```

HostLibraryTests/CMakeLists.txt
```diff
@@ -78,21 +78,7 @@ if(TENSILE_USE_HIP)
...
-    else()
-        if(EXISTS /etc/redhat-release)
-            set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fopenmp=libgomp")
-        else()
-            set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fopenmp")
-            set(OPENMP_LIBRARY /usr/lib/x86_64-linux-gnu/libomp.so)
-            set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} ${OPENMP_LIBRARY}")
-        endif()
-    endif()
+    find_package(OpenMP REQUIRED)
```
TorreZuk, line 81: For this build environment in rocBLAS we dropped the REQUIRED from this line and it was able to find the header file in order to compile.   Runtime use of OpenMP would thus likely fail so you would only get 1 thread.  Not sure if this HostLibraryTests will suffer much under one thread in this non-standard build environment.

zaliu: I don't think the find_package(OpenMP,...) in HostLibraryTests/CMakeLists.txt was the cause of the rocBLAS build failure; the one in Tensile/Source/CMakeLists.txt was.  In Tensile, we need far more than the header file.  For CPU reference result computation, OpenMP library is needed.  Sure, I can remove the REQUIRED from both places and the rocBLAS build will succeed, but then standalone benchmarking breaks:
```
-- Found OpenMP_C: -fopenmp (found version "4.5")
-- Could NOT find OpenMP_CXX (missing: OpenMP_CXX_FLAGS OpenMP_CXX_LIB_NAMES)
-- Could NOT find OpenMP (missing: OpenMP_CXX_FOUND) (found version "4.5")
LLVMObjectYAML_LIBRARY: /usr/lib/llvm-6.0/lib/libLLVMObjectYAML.a
-- Boost version: 1.65.1
-- Found the following Boost libraries:
--   program_options
--   filesystem
--   system
-- Found ROCmSMI: /opt/rocm/lib/librocm_smi64.so
-- Configuring done
CMake Error at client/CMakeLists.txt:39 (add_executable):
  Target "tensile_client" links to target "OpenMP::OpenMP_CXX" but the target
  was not found.  Perhaps a find_package() call is missing for an IMPORTED
  target, or an ALIAS target is missing?
```

haampie:
@zaliu and @TorreZuk, I just tried this minimal example:

```
$ ls
CMakeLists.txt	main.cpp
harmen-desktop example_openmp $ cat CMakeLists.txt 
cmake_minimum_required(VERSION 3.10.0)

project(TestOpenMP)

find_package(OpenMP REQUIRED)

add_executable(main main.cpp)

target_link_libraries(main OpenMP::OpenMP_CXX)
$ cat main.cpp 
#include <iostream>
#include <sstream>
#include <omp.h>


int main() {
#pragma omp parallel
{
    std::stringstream ss;
    ss << "hello world from " << omp_get_thread_num() << '\n';
    std::cout << ss.str();
}
}
$ docker run --rm -e CC=/opt/rocm/bin/hipcc -e CXX=/opt/rocm/bin/hipcc -w /sources -v $PWD:/sources rocm/dev-ubuntu-18.04 bash -c 'apt-get update -qq && apt-get install --no-install-recommends -qq cmake && mkdir -p build && cd build && cmake .. && make && ./main'
-- The C compiler identification is Clang 12.0.0
-- The CXX compiler identification is Clang 12.0.0
-- Check for working C compiler: /opt/rocm/bin/hipcc
-- Check for working C compiler: /opt/rocm/bin/hipcc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /opt/rocm/bin/hipcc
-- Check for working CXX compiler: /opt/rocm/bin/hipcc -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - failed
-- Found OpenMP_C: -fopenmp=libomp  
CMake Error at /usr/share/cmake-3.10/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
  Could NOT find OpenMP_CXX (missing: OpenMP_CXX_FLAGS OpenMP_CXX_LIB_NAMES)
Call Stack (most recent call first):
  /usr/share/cmake-3.10/Modules/FindPackageHandleStandardArgs.cmake:378 (_FPHSA_FAILURE_MESSAGE)
  /usr/share/cmake-3.10/Modules/FindOpenMP.cmake:447 (find_package_handle_standard_args)
  CMakeLists.txt:5 (find_package)


-- Configuring incomplete, errors occurred!
See also "/sources/build/CMakeFiles/CMakeOutput.log".
See also "/sources/build/CMakeFiles/CMakeError.log".
```

Seems like LLVM is not configured to build OpenMP as part of LLVM. I think it's a better idea to just use the OpenMP implementation that the current compiler provides whenever possible. In spack we enabled OpenMP for llvm-amdgpu by default: https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/llvm-amdgpu/package.py#L59-L60

haampie:
Ok, obviously you need this commit for HIP support: https://github.com/Kitware/CMake/commit/db5a8d78761e59d43db2b463a8f8aaa5e7d2be48

It landed in CMake 3.16.0.

Then there's one last issue to overcome with HIP / clang's OpenMP, because `/opt/rocm/llvm/lib` is not in the search path of the runtime linker by default, so you'd have to set LD_LIBRARY path or do some work yourself to set the RPATH correctly.

See here for an example: https://gist.github.com/haampie/268ee0b46241f421eecd4cafd973672e

Cheers!

TorreZuk:
Thanks @haampie for the detailed analysis.  I leave this to @zaliu to ponder as I am on rocblas team.   In rocBLAS we do a bit of a mix where the target if defined variable is fetched into another list so if blank due to detection failure it doesn't get added to the target link list so it can fall back to no OpenMP support single thread (as long as the header file is found).  I will look at 3.16 cmake as well but pushing that too far may cause trouble for other old platforms.

Current rocblas for reference:
if (TARGET OpenMP::OpenMP_CXX)
  set( COMMON_LINK_LIBS "OpenMP::OpenMP_CXX")
  if(OS_ID_rhel OR OS_ID_centos OR OS_ID_sles)
    list( APPEND COMMON_LINK_LIBS "-L${HIP_CLANG_ROOT}/lib;-Wl,-rpath=${HIP_CLANG_ROOT}/lib")
  endif()
endif()

I've asked that we get more config put into our llvm build tre as we require those for linking so could add PATHS to the find package to try and force the llvm based openmp if we can count on it being there.

haampie:
> I will look at 3.16 cmake as well but pushing that too far may cause trouble for other old platforms.

The other thing you can do is add a copy of cmake 3.16's FindOpenMP.cmake to your `<project>/cmake` folder until you bump the minimum required version.

```
if(OS_ID_rhel OR OS_ID_centos OR OS_ID_sles)
list( APPEND COMMON_LINK_LIBS "-L${HIP_CLANG_ROOT}/lib;-Wl,-rpath=${HIP_CLANG_ROOT}/lib")
endif()
```
Same could be done for Ubuntu right? And should it then be conditional on whether you use hip or not? The important thing is to get rpaths setup to `<llvm prefix>/lib/libomp.so`.

> I've asked that we get more config put into our llvm build tre as we require those for linking so could add PATHS to the find package to try and force the llvm based openmp if we can count on it being there.

Yeah, that'll make life easier, LLVM knows best where its libs are installed :p

I've opened an issue at Kitware/CMake to see if the OpenMP target can be improved a bit: https://gitlab.kitware.com/cmake/cmake/-/issues/21787

haampie:
Hi @TorreZuk and @zaliu,

The plot thickens, I've created an issue upstream about the `cmake` + `hipcc` + `find_package(OpenMP)` problem and got a response: https://gitlab.kitware.com/cmake/cmake/-/issues/21787

> I suspect FindOpenMP predates target_link_options.  The latter could be used to add -fopenmp=libomp as a linker flag.

`target_link_options` landed in CMake 3.13. So it's very likely `find_package(OpenMP REQUIRED)` is going to just work with hip.

To come back to this

>  I will look at 3.16 cmake as well but pushing that too far may cause trouble for other old platforms.

I'd consider that

- CMake is extremely easy to install, there's tutorials online providing one-liners (https://cliutils.gitlab.io/modern-cmake/chapters/intro/installing.html)
- `spack install rocblas` will by default use the latest cmake version 
- Ubuntu / CentOS typically ship ancient versions of CMake requiring the user almost always to install a newer version by hand
-  and CMake 3.17 is required for proper CUDA support, so people who do GPU programming typically use a recent version of CMake

so why not just bump the minimum version number 🤷

pfultz2:
> I suspect FindOpenMP predates target_link_options. The latter could be used to add -fopenmp=libomp as a linker flag.

This allows the compiler to directly link to OpenMP, but it won't set the rpath, right? It seems like downstream users will still need rpath(or `LD_LIBRARY_PATH`) so that openmp is found at runtime. This is because our packages on repo.radeon.com does not set the ldconf. Its only set correctly when building the packages from source. 

This what I do in MIGraphX [here](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/targets/cpu/CMakeLists.txt#L32), which work for older versions of cmake:

```cmake
foreach(LIBRARY ${OpenMP_CXX_LIBRARIES})
    if(LIBRARY MATCHES "libomp")
        get_filename_component(LIBRARY_PATH "${LIBRARY}" PATH)
        target_link_libraries(migraphx_cpu PUBLIC -Wl,-rpath=${LIBRARY_PATH} -Wl,-rpath-link=${LIBRARY_PATH})
    endif()
endforeach()
```

haampie:
`-fopenmp=libomp` will make clang set the rpath and it'll just work. Open Q to me is whether cmake will retain those rpaths when running `make install`, because it sometimes decides to remove rpaths during install

haampie:
So, this is I think a slightly better solution:

CMakeLists.txt:
```cmake
cmake_minimum_required(VERSION 3.13.0)

project(TestOpenMP LANGUAGES CXX)

set(USE_HIP OFF CACHE BOOL "use hip")

# Set up OpenMP
add_library(custom_openmp INTERFACE)

if (USE_HIP)
    target_compile_options(custom_openmp INTERFACE "-fopenmp")
    target_link_options(custom_openmp INTERFACE "-fopenmp")
else ()
    find_package(OpenMP REQUIRED)
    target_link_libraries(custom_openmp INTERFACE OpenMP::OpenMP_CXX)
endif ()

# Add a library
add_executable(main main.cpp)
target_link_libraries(main PUBLIC custom_openmp)
install(TARGETS main RUNTIME DESTINATION bin)
```

main.cpp:
```cpp
#include <iostream>
#include <sstream>
#include <omp.h>


int main() {
#pragma omp parallel
{
    std::stringstream ss;
    ss << "hello world from " << omp_get_thread_num() << '\n';
    std::cout << ss.str();
}
}
```

```bash
$ mkdir build && cd build
$  CXX=/home/harmen/spack/opt/spack/linux-ubuntu20.04-zen2/gcc-9.3.0/llvm-amdgpu-4.0.0-4efvzxjbq2naqneaw6lpc2tggozo7x7s/bin/clang++ cmake -DUSE_HIP=On -DCMAKE_INSTALL_RPATH=/path/to/somewhere -DCMAKE_INSTALL_PREFIX=$PWD/install ..
$ make install
$ ldd install/bin/main | grep omp
	libomp.so => /home/harmen/spack/opt/spack/linux-ubuntu20.04-zen2/gcc-9.3.0/llvm-amdgpu-4.0.0-4efvzxjbq2naqneaw6lpc2tggozo7x7s/bin/../lib/libomp.so (0x00007fe4165ba000)
$ readelf -d install/bin/main | grep PATH
 0x000000000000001d (RUNPATH)            Library runpath: [/path/to/somewhere:/home/harmen/spack/opt/spack/linux-ubuntu20.04-zen2/gcc-9.3.0/llvm-amdgpu-4.0.0-4efvzxjbq2naqneaw6lpc2tggozo7x7s/bin/../lib]
```
it preserves the rpath on install just fine. And without USE_HIP it just falls back to FindOpenMP.cmake:

```
$ cmake -DCMAKE_INSTALL_PREFIX=$PWD/install ..
$ make install
$ ldd install/bin/main | grep omp
	libgomp.so.1 => /lib/x86_64-linux-gnu/libgomp.so.1 (0x00007f6916dd4000)

```

TorreZuk (commented):
I like it if it builds and runs on all platforms.  cmake >= 3.13 probably fine too.

zaliu:
Just a quick note that @haampie's latest code here built rocBLAS just fine and passed rocblas-test sanity checking on all 4 supported platforms: Ubuntu 18.04, CentOS 7.7, CentOS 8.1 and SLES because the 4 DevOps-supplied dockers all had cmake 3.16.8 installed.  Furthermore, standalone Tensile benchmarking mode also worked just fine on Ubuntu.

At this point internal discussions are in progress to see if we want to require a non-native version of cmake (such as 3.16) when building rocBLAS or using Tensile.

TorreZuk:
Changed this to noCI label as still trying to see why it is not using OpenMP for all test cases, and due to permissions just made new PR https://github.com/ROCmSoftwarePlatform/Tensile/pull/1318 for this code.  Found a missing sub-library that was missed out but adding flags for that we still are slow.   Blanket flags could have hit some extra file or the flags are not incorrectly set with this new style.  Will leave this open until determined if we can fix or just fix the new PR and close this one later.

TorreZuk:
Thanks @haampie for your contribution!  Finally merged the modified form of this in PR #1318 so closing this one.

PR #1256, opened 2021-01-14
Fix HSS error due to PR1185

solaslin:
Fix a changing from "DataType" to "DestDataType", hence blocks some HSS logics in rocBLAS from being generated.

Will leave a notification here once the GFX900 tests pass.


zaliu (approved):
Please merge after both Tensile CI (precheckin and extended) and your offline gfx900 rocblas-test passed.  Thanks.

PR #1257, opened 2021-01-14
resolves SWDEV-267920; value of targets to clang-offload-bundler updated

zaliu:
The value of -targets= to clang-offload-bundler has been changed from prefixing with hip to hipv4 for Code Object V4.  During transition, Tensile needs to support both generations of values based on whether clang-offload-bundler supports the new -list option and whether the list returned contains the string "hipv4".

We'll improve the implementation in a future pull request to utilize the list of target values returned by -list.

### DIFF
--- a/Tensile/TensileCreateLibrary.py
+++ b/Tensile/TensileCreateLibrary.py
@@ -190,10 +190,18 @@ def buildSourceCodeObjectFile(CxxCompiler, outputPath, kernelFile):
         print('hipcc:', ' '.join(compileArgs))
       subprocess.check_call(compileArgs)
 
+      infile = os.path.join(buildPath, objectFilename)
+      try:
+        bundlerArgs = [globalParameters["ClangOffloadBundlerPath"], "-type=o", "-inputs=%s" % infile, "-list"]
+        subprocess.check_output(bundlerArgs, stderr=subprocess.STDOUT).decode().split("\n")
+        # TODO - use the list of gfx entries returned as targets directly
+        hipVersion = "hipv4"
+      except subprocess.CalledProcessError:
+        hipVersion = "hip"
+
       for i in range(len(archs)):
-        infile = os.path.join(buildPath, objectFilename)
         outfile = os.path.join(buildPath, "{0}-000-{1}.hsaco".format(soFilename, archs[i]))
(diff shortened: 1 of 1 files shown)


zaliu:
Offline rocblas-test passed on all 3 supported gfx platforms.  One of the 3 runs was using a docker with the hip->hipv4 clang-offload-bundler changes, and it built and tested fine too.

Tensile/TensileCreateLibrary.py
```diff
@@ -190,10 +190,18 @@ def isSupported(arch):
         print('hipcc:', ' '.join(compileArgs))
       subprocess.check_call(compileArgs)
 
+      infile = os.path.join(buildPath, objectFilename)
+      try:
+        bundlerArgs = [globalParameters["ClangOffloadBundlerPath"], "-type=o", "-inputs=%s" % infile, "-list"]
+        subprocess.check_output(bundlerArgs, stderr=subprocess.STDOUT).decode().split("\n")
+        # TODO - use the list of gfx entries returned as targets directly
+        hipVersion = "hipv4"
```
yxsamliu, line 198: since -list support is already in trunk and amd-stg-open branch but hipv4 change is not committed yet. So this change patch will break ePSDB. Even for a quick workaround, you still need to grep hipv4 in the output of 'clang-offloader-bundler -list'. If hipv4 is found, use hipv4, otherwise use hip.

zaliu: Done.

Tensile/TensileCreateLibrary.py
```diff
@@ -190,10 +190,18 @@ def isSupported(arch):
         print('hipcc:', ' '.join(compileArgs))
       subprocess.check_call(compileArgs)
 
+      infile = os.path.join(buildPath, objectFilename)
+      try:
+        bundlerArgs = [globalParameters["ClangOffloadBundlerPath"], "-type=o", "-inputs=%s" % infile, "-list"]
+        buffer = subprocess.check_output(bundlerArgs, stderr=subprocess.STDOUT).decode()
+        # TODO - use the list of gfx entries returned as targets directly
+        hipVersion = "hipv4" if "hipv4" in buffer else "hip"
```
yxsamliu, line 198: I am OK with this quick workaround. I hope you can have a better fix when you get time since the bundle ID may change again in the future.

yxsamliu (approved):

PR #1258, opened 2021-01-14
Tuning automation fixes and improvements

benjaminulmer:
This PR is part of an effort to better automate the Tensile tuning process, and is the beginning of the first step of that process: cleaning up and fixing the current automation scripts and tools.

Improved automation workflow:
- Can now use existing local copies of Tensile and rocBLAS for tuning
  - This is especially important for rocBLAS, as previously, each tuning exercise would build its own copy
- The Tensile directory (whether provisioned or provided) is no longer polluted with files for the tuning exercise
- All scripts use the same working/output directory. This reduces the number of arguments that must be provided to scripts and ensures all outputs and artifacts from Tuning are contained in one directory

General script improvements:
- .sh scripts should now be robust to paths with spaces
- Other safety fixes, such as exiting after failed directory changes
- Removed unused arguments and other dead code
- Removed “stroolean” (strings representing boolean) variables from Python tools

Script interface changes:
- Required arguments now passed in positionally
- Ability to force redoing previously done steps. This allows iterating on a Tuning exercise in the same working directory

Bug fixes:
- Handle different values for a/b_type, c/d_type, and compute type when generating tuning config files
- Record correct function name for gemm call when generating yaml files used with rocblas-bench

Documentation and readability:
- Better help messages for all the tuning scripts
- More output during script execution
- General code readability improvements

It is worth reemphasizing that these changes alter the calling convention for the bash scripts in a **non-backwards compatible** way. If this is determined to be an issue, I can add legacy interface support.

As far as I can tell, the five main tuning scripts (provision_tuning, provision_verification, run_validation, analyze_results, and master_tuning) are working as expected based on my testing.


tuning/scripts/master_tuning_script.sh
```diff
@@ -1,414 +1,330 @@
...
-        if [[ "$(grep -c 'compute_type f16_r' ${LOG})" -gt 0 || "$(grep -c 'compute_type: "f16_r"' ${LOG})" -gt 0 ]]; then
-            printf "compute_type of f16 detected, disabling HighPrecisionAccumulate\n"
-            DISABLE_HPA=true
-        fi
-    else
-        printf "Could not detect data type in log file, assuming sgemm\n"
-        DATA_TYPE=sgemm
-        DVAL=2
-    fi
+if ${DEPENDENCIES}; then
```
cgmillette, line 113: Hi Ben, are you using tabs or spaces for your indentations?
I think the preferred formatting is 4 spaces for indent. Can you please update this in your editor?

I see a few inconsistencies above using 2 spaces, then here looks like tab.

tuning/scripts/analyze_results.sh
```diff
@@ -0,0 +1,125 @@
+#!/bin/sh
```
cgmillette, line 1: Just from an outside perspective, why is this a bash script and not a python script? Argument parsing, file enumeration and calls to other scripts would be drastically simplified. Not to mention portability would now be extended to any platform running Python and would be more consistent.

tuning/scripts/master_tuning_script.sh
cgmillette: Also, did you also run .githooks/install in the Tensile repo root?

tuning/scripts/master_tuning_script.sh
```diff
@@ -1,414 +1,330 @@
...
-    fi
-    if [[ $(ls -A clients/*.deb | wc -c) -eq 0 ]]; then
-        make package_clients
-        cp clients/*.deb ../../../packages/client
-    fi
-}
-
-
+# run the four individual tuning scripts (and Tensile) in order
+SCRIPT_PATH=$(realpath "${0}" | xargs dirname)
```
cgmillette, line 236: I don't think realpath is compatible on OSX or CentOS. (unsure about Windows).
If there are real use cases where we need to support OSX, then we might want to think about alternatives for this.

tuning/scripts/master_tuning_script.sh
```diff
@@ -1,414 +1,330 @@
...
 
-SCRIPT_PATH="$( cd "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
-PROVISION_TUNNING=${SCRIPT_PATH}/provision_tuning.sh
+# provision_tuning.sh
+if [ ! -d "${WORKING_PATH}/library" ] || ${REDO}; then
+  echo "Running ./provision_tuning.sh"
```
cgmillette, line 241: This is misleading about the script to run -- see below as  './provision_tuning.sh' != "${SCRIPT_PATH}/provision_tuning.sh"

I may suggest to save the cmd to a string and then echo it before invoking.

PROVISION_TUNING=${SCRIPT_PATH}/provision_tuning.sh
echo "Running ${PROVISION_TUNING}"
...

tuning/scripts/master_tuning_script.sh
```diff
@@ -1,414 +1,330 @@
...
+pushd "${WORKING_PATH}/make" > /dev/null || exit
+./doit-all.sh
+popd > /dev/null || exit
 
-# run tuning
-./tune_tensile.sh
-popd > /dev/null
+# provision_verification.sh
+if [ ! -d "${WORKING_PATH}/library" ] || ${REDO}; then
+  echo "Running ./provision_verification.sh"
```
cgmillette, line 282: Same as above with script location

tuning/scripts/provision_repo.sh
```diff
@@ -1,56 +1,64 @@
...
-mkdir -p ${WORKING_PATH}
+mkdir -p "${WORKING_PATH}"
 
-pushd ${WORKING_PATH} > /dev/null
+pushd "${WORKING_PATH}" > /dev/null || exit
 
-TENSILE_HOST="https://github.com/${TENSILE_FORK}/Tensile.git"
-ROCBLAS_HOST="https://github.com/${ROCBLAS_FORK}/rocBLAS-internal.git"
+TENSILE_HOST="https://github.com/${FORK}/Tensile.git"
+ROCBLAS_HOST="https://github.com/${FORK}/rocBLAS-internal.git"
```
cgmillette, line 60: Here, I wouldn't force both Tensile and RocBLAS to the same fork.

For example: This would make it pretty annoying to benchmark my custom fork of Tensile with experimental changes against a stable RocmSoftwarePlatform version of rocBLAS.

tuning/scripts/stage_tuning.sh
```diff
@@ -1,86 +1,87 @@
...
+    echo "#!/bin/sh"
+    echo "if [ ! -d 3_LibraryLogic ] || [ -z \"\$(ls -A 3_LibraryLogic)\" ]; then"
+    echo "  touch time.begin"
+    echo "  ${TENSILE}/Tensile/bin/Tensile ${FILE_NAME} ./ > make.out 2>&1"
+    echo "  touch time.end"
+    echo "fi"
+  } > doit.sh
+
+  chmod +x doit.sh
+  popd > /dev/null || exit
```
cgmillette, line 26: "doit.sh" for me doesn't quite "do it" heh. I am hoping to rename this to something more descriptive, say launchTensileBenchmark.sh or something better.

tuning/scripts/stage_tuning.sh
```diff
@@ -1,86 +1,87 @@
...
-    echo "if [ ! -d 3_LibraryLogic ] || [ \$(ls -A 3_LibraryLogic | wc -c) -eq 0 ]; then" >> doit.sh
-    echo "  touch time.begin" >> doit.sh
-    echo "  ../Tensile/bin/Tensile $FILENAME ./ > make.out 2>&1" >> doit.sh
-    echo "  touch time.end" >> doit.sh
-    echo "fi" >> doit.sh
-
-    chmod +x doit.sh
-    popd > /dev/null
+function make_tensile_tuning() {
+
```
cgmillette, line 4: May be out of scope for this review, but this is another example of a script that may be better suited to a Python environment. This generates a bash script that is only useful for that one specific logic file - IMO it makes the workflow pretty rigid.

cgmillette (changes requested):
Overall, I like the improvements with the input arguments handling and the scripts flow.

Requesting minor changes for whitespace formatting and some script naming and verbosity. Also compatibility with other platforms.

More fundamental areas I think worthy of discussion are:
- Cross platform support? I.e. OSX and Windows tuning? There are a couple areas to consider.
- Python / bash crossover. My personal opinion with bash scripts is that they should be used when setting up environments/dependencies and for doing things with minimal overhead. E.g. downloading and installing packages, doing git cmds, etc. Python excels more with the ease of directory and file navigation/enumeration, string manipulation and configuration settings. We might consider some of the scripts below more natural to convert to Python? @wbgilmartin any thoughts?

cgmillette:
An added note with the commit history: - I've learned that as of pull request time, your commit history is a public facing log and many folks really appreciate a tidy and compact history to review. It wouldn't hurt to go back and review your commits with interactive rebase and squash/merge commits together that are relevant, or remove the ones that aren't needed (e.g. a delete and then later restore).

tuning/scripts/provision_repo.sh
```diff
@@ -1,56 +1,64 @@
...
```
wbgilmartin, line 59: Do we want to expose rocBLAS-internal repo as the default to the public? It seems that this should be hidden from view.

tuning/scripts/stage_tuning.sh
wbgilmartin: Yah, this was my half baked naming convention from my original version of the automation. It should be changed to something meaningful.

tuning/scripts/analyze_results.sh
```diff
@@ -0,0 +1,125 @@
...
+  SCLK=1000
+fi
+
+if [ $# != 3 ]; then
+  echo "Exactly three possitional args required"
+  echo "See ${0} --help"
+fi
+
+WORKING_PATH=${1}
+LOG=${2}
```
wbgilmartin, line 65: The current version of this has a log file as a requirement. This was not always the case. This log file is use to extract the call count information. It would be nice if this was optional, in which case the call counts would be set to 1 by default. 

This would enable a user to run the analysis apart from the context of some application use case.

tuning/scripts/master_tuning_script.sh
benjaminulmer: Oops, look like I missed this file when I was cleaning up the indentation. 

I went for 2 spaces for these scripts as that's what they were before, but I can easily update them to 4 spaces if that is preferred.

tuning/scripts/provision_repo.sh
benjaminulmer: This script will provision Tensile _or_ rocBLAS for a given incovation, not both. Consolodating these two variables into one allows the script to be invoked with the same API for Tensile and rocBLAS (with the added -r argument to specify rocBLAS should be provisioned instead of Tensile).

It's definitely not the most intuitive interface, so I can work on something better if we think it's necessary.

tuning/scripts/provision_repo.sh
benjaminulmer: Internal was already in place when I started working on these, so I just left it. However, with the new ability to specify an existing local rocBLAS path, I don't see any reason why the ability to explicitly use the internal repo is needed in these scripts. If the internal (or any other version) is needed, the user can just prep rocBLAS ahead of time themselves.

tuning/scripts/analyze_results.sh
benjaminulmer: At some point the log file was made mandatory, and I left it that way for now as PerformanceAnalysis.py requires it. I agree this should be made optional, but might be outside the scope of this PR.

wbgilmartin (approved):
LGTM

benjaminulmer:
> More fundamental areas I think worthy of discussion are:
> 
> * Cross platform support? I.e. OSX and Windows tuning? There are a couple areas to consider.
> * Python / bash crossover. My personal opinion with bash scripts is that they should be used when setting up environments/dependencies and for doing things with minimal overhead. E.g. downloading and installing packages, doing git cmds, etc. Python excels more with the ease of directory and file navigation/enumeration, string manipulation and configuration settings. We might consider some of the scripts below more natural to convert to Python? @wbgilmartin any thoughts?

I think most of the functionality of these scripts should eventually move to Python, which would also solve most of the cross platform issues. I think this will be even more important as try to automate more aspects of tuning.

tuning/scripts/master_tuning_script.sh
benjaminulmer: After a bit of searching I didn't find a good cross platform way of doing this succinctly. I've left it as is for now, as if we move these scripts into Python (which I want to do eventually), it will not be an issue.

zaliu (changes requested):
You stated: "It is worth reemphasizing that these changes alter the calling convention for the bash scripts in a non-backwards compatible way. If this is determined to be an issue, I can add legacy interface support."  I'm not suggesting backward compatibility is critical, but it is very important to produce a parallel version of @amdkila's [Tensile Tuning Overview](http://confluence.amd.com/display/MLSE/Tensile+tuning+overview) document so that people wanting to do "hands-on" tuning (and auto tuning) can follow your new confluence page.  We do have some internal customers following @amdkila's confluence page to do their own tuning.  Please add a link to @amdkila's confluence page pointing to your new confluence page that matches the calling conventions you changed in this PR.  Thanks so much.

tuning/scripts/provision_repo.sh
cgmillette: Right, but the rocBLAS provision by itself can have a custom tensile branch that it uses to build libraries. I highly recommend keeping both, and forwarding the tensile host to the rocblas builds.

tuning/scripts/master_tuning_script.sh
cgmillette: I think we will be shooting for 4 spaces moving forward. Does a lot for readability.

cgmillette (changes requested):
Almost there, I just think we should still consider different forks / branches for rocBLAS and Tensile.

tuning/scripts/provision_repo.sh
benjaminulmer: I hadn't considered this. I'll take a stab at forwarding the version of Tensile used (either local copy or fork and branch) to the rocBLAS provision.

benjaminulmer:
> You stated: "It is worth reemphasizing that these changes alter the calling convention for the bash scripts in a non-backwards compatible way. If this is determined to be an issue, I can add legacy interface support." I'm not suggesting backward compatibility is critical, but it is very important to produce a parallel version of @amdkila's [Tensile Tuning Overview](http://confluence.amd.com/display/MLSE/Tensile+tuning+overview) document so that people wanting to do "hands-on" tuning (and auto tuning) can follow your new confluence page. We do have some internal customers following @amdkila's confluence page to do their own tuning. Please add a link to @amdkila's confluence page pointing to your new confluence page that matches the calling conventions you changed in this PR. Thanks so much.

I've discussed with @bragadeesh how to best handle the tuning documentation, as the page linked was out of date even prior to the changes made in this PR. For now, we've decided the simplest solution is to have a pre-built pdf of the tuning document available on the page, which I have updated to match the changes made in this PR (see #1272). I'll update the page with the pdf once this PR is live, and continue to keep it updated with any further changes to the document.

tuning/scripts/provision_repo.sh
benjaminulmer: Latest commit addresses this, although it is handled in provision_verification.sh instead of this script. If my understanding is correct, the provisioned rocBLAS will be built with the same local copy of Tensile that was used for tuning.

cgmillette (approved):
LGTM

cgmillette:
@zaliu has your change request been addressed?

zaliu (approved):
As agreed upon offline, a separate confluence page will be created to bring the anaconda portion of Akila's confluence page up to date.  Thanks so much for your willingness to do the extra work, @benjaminulmer!

PR #1263, opened 2021-01-19
CHANGELOG.md update for ROCm 4.1

zaliu:


zaliu:
It looks like for Tensile, the NoCI label doesn't work too well for the "required" StaticAnalysis.

eidenyoshida (approved):
Formatting looks to match the other math libs

PR #1268, opened 2021-01-22
Refactor ConvolutionVsContraction

jichangjichang:
**Refactor ConvolutionVsContraction:**

> The original design of **ConvolutionVsContraction** needs to specify the convolution problem size  exactly into convolution config of yaml. Then the convolution config would be translated into "convolution-identifier" of client parameter. It means that only 1 conv problem size is allowed since all conv problem sizes of this yaml must be exactly identical with conv config.

> This patch refactor the design to store each convolution problem sizes into another client parameter "convolution-problem" along with each problems. It allows user to set different problem sizes into same yaml for ConvolutionVsContraction validation. 

> With this patch, we are able to enable ConvolutionVsContraction for all NCHW conv forward tests.

**support zeropad in CPU convolution validation**

**Fix bug: the channel stride is incorrect when convert conv problem into tensor contraction problem.**

**CI tests: Extend the test scope of conv_vs_contraction CI tests**
  

> With the new design, we are able to test multiple problem sizes in one convolution configs. This commit set the problemLevel = 2 for test_conv_vs_contraction CI tests. It would test 28 problem sizes for each convolution config. This commit also removes all pytest.mark.skip() from test_conv_vs_contraction and add testing for zero padding. 

example: new parameter "convolution-problem" of ClientParameter.ini
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@
 - Support offset for each input/output buffer in Tensile
 - support support ldc != ldd for all GEMM kernel
 
+### Optimizations
+- Refactor ConvolutionVsContraction
+
+### Fixed
+- channel stride is incorrect when converting conv problem into tensor contraction problem
 
 ## [Tensile 4.26.0 for ROCm 4.1.0]
 ### Added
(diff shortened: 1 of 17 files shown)


zaliu:
@imcarsonliao Would you please have someone on your team to do a peer review first?  Thanks so much.

TonyYHsieh (approved):
looks good

jichangjichang:
> @imcarsonliao Would you please have someone on your team to do a peer review first? Thanks so much.
@zaliu we have done the peer review . Please also help assign someone to review this PR.

sdquiring (approved):

PR #1273, opened 2021-01-26
Pass CMAKE_CXX_COMPILER to Python script

AlexBrownAMD:
To work around the build issue, cmake command line for HostLibraryTests must include -DCMAKE_CXX_COMPILER=/opt/rocm/bin/hipcc or have the path defined as an environment variable. Having /opt/rocm/bin pre-pended to PATH may also work.


Tensile/cmake/TensileConfig.cmake
```diff
@@ -178,6 +179,10 @@ function(TensileCreateLibraryFiles
     set(Options ${Options} "--cxx-compiler=${Tensile_COMPILER}")
   endif()
 
+  if(Tensile_COMPILER_PATH)
+    set(Options ${Options} "--cmake-cxx-compiler=${Tensile_COMPILER_PATH}")
+  endif()
+
```
yoichiyoshida, line 185: Why is this not just included in the --cxx-compiler above?

yoichiyoshida (commented):
--cxx-compiler and --cmake-cxx-compiler should be combined?

Tensile/cmake/TensileConfig.cmake
AlexBrownAMD: --cxx-compiler is used to switch between hipcc and hcc and should probably be removed at this point. Mainly added a new path option to avoid problems where --cxx-compiler option does not support having a full path, just a compiler name to switch on

Tensile/Common.py
```diff
@@ -1618,6 +1618,8 @@ def assignGlobalParameters( config ):
     globalParameters["ROCmPath"] = os.environ.get("ROCM_PATH")
   if "TENSILE_ROCM_PATH" in os.environ:
     globalParameters["ROCmPath"] = os.environ.get("TENSILE_ROCM_PATH")
+  if "CMAKE_CXX_COMPILER" in os.environ:
+    globalParameters["CmakeCxxCompiler"] = os.environ.get("CMAKE_CXX_COMPILER")
```
sdquiring, line 1622: We shouldn't have globalParameters that only sometimes exist.  This causes issues with testing.  Using `None` as a default value would work.

AlexBrownAMD: Fixed in latest commit

zaliu (approved):
Thanks so much!

PR #1275, opened 2021-01-28
Tile aware metric integration

wbgilmartin:
Updates to enable runtime switch to run the tile aware metric selection method.

In this version it is enabled using the switch ENABLE_DEBUG_SELECTION=1.

To add the summation sweeps to create the linear summation model used by the TAM selection run the tool

./TensileGenerateSummations existing_logic_dir output_dir

The updated logic will be added to output_dir/final


bragadeesh:
Can we rename the environment variable to TENSILE_TAM_SELECTION_ENABLE

Tensile/Source/lib/source/ContractionSolution.cpp
```diff
@@ -1135,6 +1135,95 @@ namespace Tensile
         return pp;
     }
 
+    ContractionSolution::TAMetricProblemScore
```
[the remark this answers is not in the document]
wbgilmartin: I am not sure what you are getting at here, we do not know the problem dimensions before runtime so how de we compute those granularities.

we can possibly cache the granularities for the exact sizes.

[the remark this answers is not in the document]
sdquiring: I mean that calculating granularity seems to have been implemented in several different places.  I'm suggesting that we should unify the calculation of granularity into a function and then call that from here and from `projectedPerformance`.

Tensile/TensileCreateLibrary.py
```diff
@@ -1326,6 +1336,8 @@ def TensileCreateLibrary():
   print2(HR)
   print2("")
 
+  print ("this is an exit")
```
cgmillette, line 1339: Can remove this

Tensile/TensileCreateLibrary.py
```diff
@@ -1428,6 +1440,7 @@ def splitExtraParameters(par):
   if not os.path.exists(logicPath):
     printExit("LogicPath %s doesn't exist" % logicPath)
 
+
   # Translate GPU targets to filter filenames in Tensile_LOGIC directory
   mapArchitecture = {'all':'_','gfx000':'none', 'gfx803':'r9nano',
         'gfx900':'vega10', 'gfx906:xnack-':'vega20', 'gfx908:xnack-':'arcturus'}
```
cgmillette, line 1446: This conflicts with the new global definition @ line 50. This one includes new xnack flags, which will be needed? 
Can we consolidate these tables and / or put them into common ? 

Also the point of this table seems to be mapping architecture aliases, so would a better name be architectureAliasMap?

Tensile/TensileCreateLibrary.py
```diff
@@ -47,6 +47,16 @@
 import time
 from copy import deepcopy
 
+ # Translate GPU targets to filter filenames in Tensile_LOGIC directory
+mapArchitecture = {'all':'_','gfx000':'none', 'gfx803':'r9nano',
+  'gfx900':'vega10', 'gfx906':'vega20', 'gfx908':'arcturus'}
+
```
cgmillette, line 53: See comment @ line 1446.

Tensile/TensileCreateLibrary.py
```diff
@@ -47,6 +47,16 @@
...
+
+def getArchitectureName(gfxName):
+  if gfxName in mapArchitecture:
+    return mapArchitecture[gfxName]
+  else:
+    return None
```
cgmillette, line 60: This function is not used in this file. May be a better place to put this. Common?

Tensile/GenerateSummations.py
```diff
@@ -0,0 +1,201 @@
...
+import yaml
+import subprocess
+import glob
+
+from shutil import copyfile
+from copy import deepcopy
+
+from . import LibraryIO
+
+from . import TensileCreateLibrary
```
cgmillette, line 36: Probably don't need to import the entire TensileCreateLibrary module - only one function used

Tensile/GenerateSummations.py
```diff
@@ -0,0 +1,201 @@
...
+            deviceNames, logicTuple)
+
+        createLibraryForBenchmark(localLogicPath, libPath, currentPath)
+
+        logic1 = LibraryIO.readLibraryLogicForSchedule(localLogicFilePath)
+
+        (scheduleName1, deviceNames1, problemType1, solutionsForSchedule1, \
+           indexOrder1, exactLogic1, rangeLogic1, _, architectureName1) = logic1
+
+        exactList = []
```
cgmillette, line 130: I am unclear why you would write a modified logic file and read it back in - can we avoid the IO by just using the modified logic tuple?

Tensile/Source/lib/include/Tensile/ContractionSolution.hpp
```diff
@@ -144,6 +181,22 @@ namespace Tensile
                                                       double totalGranularity,
                                                       int    globalSplitU) const;
 
+        TAMetricProblemScore computeProblemScore(Hardware const& hardware,
+                                                 double          M,
+                                                 double          N,
```
cgmillette, line 186: Why is there a disparity between using size_t and double types for M, N and K values? These are integral values, no?

Tensile/GenerateSummations.py
```diff
@@ -0,0 +1,179 @@
...
+    args = [pythonExePath, \
+        "--merge-files", "--no-legacy-components", \
+        "--new-client-only", "--no-short-file-names", "--no-library-print-debug", "--architecture=all", \
+        "--code-object-version=V3", "--cxx-compiler=hipcc", "--library-format=yaml", \
+        logicPath, libraryPath, "HIP"]
+
+    try:
+        subprocess.run(args, check=True, cwd=currentPath)
+    except (subprocess.CalledProcessError, OSError) as e:
+        print("ClientWriter Benchmark Process exited with error: {}".format(e))
```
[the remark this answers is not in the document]
wbgilmartin: This is changed  in the latest version.

Tensile/Source/lib/include/Tensile/PropertyMatching.hpp
```diff
@@ -76,6 +76,10 @@ namespace Tensile
             virtual std::tuple<ReturnValue, double> findBestMatch(Object const& object,
                                                                   Transform transform) const = 0;
 
+            virtual ReturnValue findBestEvaluationSolution(Object const& object,
```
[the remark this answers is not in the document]
wbgilmartin: This is fixed in the latest version.

Tensile/Source/lib/source/ContractionSolution.cpp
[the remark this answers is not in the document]
wbgilmartin: This is fixed in the latest version.

Tensile/Source/lib/include/Tensile/PropertyMatching.hpp
```diff
@@ -76,6 +76,10 @@ namespace Tensile
             virtual std::tuple<ReturnValue, double> findBestMatch(Object const& object,
                                                                   Transform transform) const = 0;
 
+            virtual ReturnValue findBestEvaluationSolution(Object const&   object,
+                                                           Hardware const& hardware,
+                                                           Transform       transform) const = 0;
```
cgmillette, line 81: It almost makes more sense to me to create your own matching table class instead of piggy backing onto DistanceMatchingTable and forcing every other descendant to implement the new findBestEvaluationSolution interface. 

Your new implementation finds the best match based on TAM metrics, and does no 'distance matching' based on sizes. Thus I feel it makes more sense to upgrade the findBestMatch interface to include the hardware reference and implement a separate TAM matching table class that implements findBestMatch instead.

Of course this is my opinion and we can discuss more if you like.

Tensile/TensileCreateLibrary.py
wbgilmartin: fixed.

Tensile/TensileCreateLibrary.py
wbgilmartin: fixed: This was moved to Common.py.

Tensile/TensileCreateLibrary.py
wbgilmartin: Fixed.

Tensile/TensileCreateLibrary.py
wbgilmartin: fixed.

Tensile/GenerateSummations.py
wbgilmartin: fixed

Tensile/GenerateSummations.py
wbgilmartin: This is a work around related to a bug in Tensile.

The first part of this is to generate consistent solution naming for both the bench marks generated by the client and the naming in the logic file.  Unfortunately in Tensile we have inconsistent typing for the DataType member of the ProblemType one version is a string and the other is an integer.  In the logic file this is set as a string and there is no problem reading it. However the process of generating the naming changes the type from string to integer. This is a problem for Tensile further down in the workflow which expects an the DataType field to be a string. To solve this would probably require some refactoring and a deeper investigation which is out of scope for this task.

Tensile/Source/lib/include/Tensile/PropertyMatching.hpp
wbgilmartin: I am not sure how the final version of the TAM is going to be instrumented. 

What you are suggesting would be require the TAM to be recognized as part of de-serialization (for the findBestSolution interface dispatch the selection correctly) and that would require modification to the logic file. 

The requirement here is for a runtime switch and have it work using the existing logic.

At this time the feature is for evaluation. I don't expect this version will survive without revisions and if the TAM is proven the final version may be as you suggest here.

HostLibraryTests/ProjectedPerformance_test.cpp
```diff
@@ -164,8 +164,8 @@ TEST(ContractionPerformance, Problem3)
 
     // std::cout << model << "\n";
 
-    ASSERT_DOUBLE_EQ(model.tilesPerCu, .1875);
-    ASSERT_DOUBLE_EQ(model.totalGranularity, .140625);
+    ASSERT_DOUBLE_EQ(model.granularites.tilesPerCu, .1875);
```
sdquiring, line 167: granularities

Tensile/GenerateSummations.py
```diff
@@ -13,21 +33,29 @@
...
     archid = gfxName(isaName)
-    return TensileCreateLibrary.getArchitectureName(archid)
+    return getArchitectureName(archid)
 
 def isValidArch(archName, currentArch):
     arch = gfxArch(archName)
     return currentArch == arch
 
+
+##############################################################################
```
sdquiring, line 51: Should be a docstring instead of a comment.

Tensile/Source/lib/include/Tensile/ContractionSolution.hpp
```diff
@@ -110,18 +110,39 @@ namespace Tensile
             size_t memGlobalWrites = 0;
         };
 
-        struct ProjectedPerformance
+        struct Granularites
```
sdquiring, line 113: Granularities

sdquiring (changes requested):
This is better.  I'd like to see the spelling fixed and the new comment converted to docstring but I think that should be enough to get this checked in.

Tensile/Source/client/source/BenchmarkTimer.cpp
```diff
@@ -83,14 +83,14 @@ namespace Tensile
                 = solution.projectedPerformance(m_problem, m_hardware);
             m_solution = solution;
 
-            m_reporter->report(ResultKey::Tile0Granularity, pp.tile0Granularity);
-            m_reporter->report(ResultKey::Tile1Granularity, pp.tile1Granularity);
-            m_reporter->report(ResultKey::CuGranularity, pp.cuGranularity);
-            m_reporter->report(ResultKey::WaveGranularity, pp.waveGranularity);
-            m_reporter->report(ResultKey::TotalGranularity, pp.totalGranularity);
+            m_reporter->report(ResultKey::Tile0Granularity, pp.granularites.tile0Granularity);
```
cgmillette, line 86: Granularities

sdquiring (approved):

PR #1277, opened 2021-02-03
merge staging 7f365fa2 into master with GO from CQE 

zaliu:
matches rocBLAS-internal [#538](https://github.com/ROCmSoftwarePlatform/rocBLAS-internal/pull/538)


zaliu:
Note that an offline run of both Tensile precheckin and Tensile extended passed well within the CI configured timeout on gfx908.  I have no idea why time they would time out during CI runs.  Just FYI:
```
precheckin: ================= 110 passed, 1 skipped in 4374.83s (1:12:54) ==================
extended: ================= 510 passed, 51 skipped in 7629.43s (2:07:09) =================
```

amcamd (approved):
LGTM, Tensile tag 7f365fa2 is the one used by rocBLAS master branch, it is one off the tip of this merge, and the commit at the tip is a merge commit.

zaliu:
The tip merge commit was needed because --ff-only didn't work.

PR #1279, opened 2021-02-11
Per CU benchmarking and Tensile library support

benjaminulmer:
This PR is a prototype for adding support in Tensile for kernels with performace reported per CU. There are two main components of this:
1) At the benchmarking phase, reporting either GFlops/CU _or_ raw GFlops as determined by the `BenchmarkPerCU` global parameter and reflecting this in the generated library logic yaml file
2) Adding an additional layer to the Master Solution Library (currently, between the Problem Map and Predicate layers) which decides between picking kernels with per CU or overall performance data

Below are some notes on the implementation of these features that I would like some feedback on:
- Each library logic file will have either overall or per CU tuned kernels, not both
  - This is indicated by a boolean at [10] in the library logic file: true for per CU and false for overall
  - For backwards compatibility, if [10] in the logic file doesn't exist, the file is treated as having overall results
- The `PerCUorOverall` layer of the Master Solution Library is an `ExactSolutionLibrary` and is placed above the normal Predicate layer
  - Allows quickly filtering out all per CU kernels if overall speed is desired and vice versa
  - Allows falling back to overall speed kernels (all existing ones in rocBLAS) if a per CU one is requested but cannot be provided
  - Currently its own class, but could be implemented with the existing `ProblemSelectionLibrary` at the cost of not having a unique name for the layer.  
- Currently, the predicate for determining if per CU kernels should be selected is hardcoded to return true
  - For the short term, some metric for choosing per CU or overall is needed. This could either be determined by the problem sizes or the total flop count of the problem
  - Eventually, user control can be added so that per CU or overall speed can be requested specifically

Other things yet to do/add:
- New tests?
- Buckets in tuning scripts for tuning per CU
- Update changelog once ready for review

Also, this PR fixes a bug which resulted in the Master Solution Library having multiple rows for the same hardware (changes in Hardware.py).

### DIFF
--- a/Tensile/LibraryIO.py
+++ b/Tensile/LibraryIO.py
@@ -130,9 +130,9 @@ def readLibraryLogicForSchedule( filename ):
   data = yaml.load(stream, yaml.SafeLoader)
   stream.close()
 
-  # verify
-  if len(data) < 6:
-    printExit("len(%s) %u < 6" % (filename, len(data)))
+  # verify data has all the fields we need
+  if len(data) < 9:
+    printExit("Library logic file %s is missing required fields (len = %u < 9)" % (filename, len(data)))
 
   # parse out objects
   versionString     = data[0]["MinimumRequiredVersion"]
--- a/Tensile/LibraryIO.py
+++ b/Tensile/LibraryIO.py
@@ -1,5 +1,5 @@
 ################################################################################
-# Copyright 2016-2020 Advanced Micro Devices, Inc. All rights reserved.
+# Copyright 2016-2021 Advanced Micro Devices, Inc. All rights reserved.
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
@@ -132,7 +132,7 @@ def readLibraryLogicForSchedule( filename ):
 
   # verify
   if len(data) < 6:
-    printExit("len(%s) %u < 7" % (filename, len(data)))
+    printExit("len(%s) %u < 6" % (filename, len(data)))
 
   # parse out objects
   versionString     = data[0]["MinimumRequiredVersion"]
@@ -264,7 +264,10 @@ class Writer:
       tileSelectionIndices = logicTuple[6]
       tileSelectionLogic["TileSelectionIndices"] = tileSelectionIndices
       data.append(tileSelectionLogic)
+    else:
+      data.append(None)
 
+    data.append(logicTuple[7])
     return data
 
 class YAMLWriter(Writer):
@@ -295,7 +298,7 @@ class MessagePackWriter(Writer):
       try:
         msgpack.pack(data, f)
       except NameError:
-        printExit("You must install MessagePack for Python to use Tensile (to parse config files). See https://github.com/msgpack/msgpack-python for installation instructions.") 
+        printExit("You must install MessagePack for Python to use Tensile (to parse config files). See https://github.com/msgpack/msgpack-python for installation instructions.")
 
 
   def writeLibraryLogicForSchedule(self, filePath, schedulePrefix, architectureName, \
@@ -315,4 +318,4 @@ class MessagePackWriter(Writer):
     except IOError:
       printExit("Cannot open file: %s" % filename)
     except NameError:
-      printExit("You must install MessagePack for Python to use Tensile (to parse config files). See https://github.com/msgpack/msgpack-python for installation instructions.") 
+      printExit("You must install MessagePack for Python to use Tensile (to parse config files). See https://github.com/msgpack/msgpack-python for installation instructions.")
--- a/Tensile/Source/client/include/ClientProblemFactory.hpp
+++ b/Tensile/Source/client/include/ClientProblemFactory.hpp
@@ -2,7 +2,7 @@
  *
  * MIT License
  *
- * Copyright 2019-2020 Advanced Micro Devices, Inc.
+ * Copyright 2019-2021 Advanced Micro Devices, Inc.
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
@@ -70,17 +70,18 @@ namespace Tensile
             ContractionProblem::BatchIndices m_batchIndices;
             ContractionProblem::BoundIndices m_boundIndices;
 
-            DataType       m_aType;
-            DataType       m_bType;
-            DataType       m_cType;
-            DataType       m_dType;
-            DataType       m_alphaType;
-            DataType       m_betaType;
-            bool           m_stridedBatched;
-            bool           m_highPrecisionAccumulate;
-            bool           m_deterministicMode;
-            ArithmeticUnit m_arithmeticUnit;
-            KernelLanguage m_kernelLanguage;
+            DataType        m_aType;
+            DataType        m_bType;
+            DataType        m_cType;
+            DataType        m_dType;
+            DataType        m_alphaType;
+            DataType        m_betaType;
+            bool            m_stridedBatched;
+            bool            m_highPrecisionAccumulate;
+            bool            m_deterministicMode;
+            ArithmeticUnit  m_arithmeticUnit;
+            BenchmarkMetric m_benchmarkMetric;
+            KernelLanguage  m_kernelLanguage;
 
             std::vector<std::vector<size_t>> m_problemSizes;
             std::vector<std::vector<size_t>> m_aStrides;
--- a/Tensile/Source/client/include/ClientProblemFactory.hpp
+++ b/Tensile/Source/client/include/ClientProblemFactory.hpp
@@ -70,18 +70,18 @@ namespace Tensile
             ContractionProblem::BatchIndices m_batchIndices;
             ContractionProblem::BoundIndices m_boundIndices;
 
-            DataType        m_aType;
-            DataType        m_bType;
-            DataType        m_cType;
-            DataType        m_dType;
-            DataType        m_alphaType;
-            DataType        m_betaType;
-            bool            m_stridedBatched;
-            bool            m_highPrecisionAccumulate;
-            bool            m_deterministicMode;
-            ArithmeticUnit  m_arithmeticUnit;
-            BenchmarkMetric m_benchmarkMetric;
-            KernelLanguage  m_kernelLanguage;
+            DataType          m_aType;
+            DataType          m_bType;
+            DataType          m_cType;
+            DataType          m_dType;
+            DataType          m_alphaType;
+            DataType          m_betaType;
+            bool              m_stridedBatched;
+            bool              m_highPrecisionAccumulate;
+            bool              m_deterministicMode;
+            ArithmeticUnit    m_arithmeticUnit;
+            KernelLanguage    m_kernelLanguage;
+            PerformanceMetric m_performanceMetric;
 
             std::vector<std::vector<size_t>> m_problemSizes;
             std::vector<std::vector<size_t>> m_aStrides;
--- a/Tensile/Source/client/source/BenchmarkTimer.cpp
+++ b/Tensile/Source/client/source/BenchmarkTimer.cpp
@@ -104,7 +104,7 @@ namespace Tensile
                 = m_solution.projectedPerformance(m_problem, m_hardware);
 
             double gflops      = m_problem.flopCount() / (timePerEnqueue_us) / 1000.0;
-            double tiles       = pp.tilesPerCu * perf.CUs;
+            int    tiles       = pp.tilesPerCu * perf.CUs;
             int    usedCus     = std::min(tiles, perf.CUs);
             double gflopsPerCu = gflops / usedCus;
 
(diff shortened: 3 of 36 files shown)


JeremyAdamHart (commented):
Aside from the comment I left, lgtm.

Tensile/Source/lib/include/Tensile/ContractionProblemPredicates.hpp
```diff
@@ -1055,6 +1055,29 @@ namespace Tensile
...
+                PerCU() = default;
+
+                static std::string Type()
+                {
+                    return "PerCU";
+                }
+
+                virtual bool operator()(ContractionProblem const& problem) const override
+                {
+                    return true; // TODO - logic for determining to use per CU or not
```
sdquiring, line 1077: Ultimately some kind of user-accessible switch in the ContractionProblem should be involved here. Something like `deterministicMode` or `arithmeticUnit`.  I think naming it something along the lines of selecting the most efficient solution vs. selecting the fastest solution will be more intuitive.  I wonder if naming this feature in that way ('efficient' or 'efficiency') would be better than 'per-cu'?

sdquiring (commented):
I wonder if

Tensile/Source/lib/include/Tensile/ContractionProblemPredicates.hpp
benjaminulmer: What about 'CUEfficiency' for the predicate name? Still conveys the idea of selecting fastest vs most efficient, but retains the fact that this is calculated relative to CUs.

Tensile/Source/lib/include/Tensile/ExactLogicLibrary.hpp
```diff
@@ -213,6 +213,33 @@ namespace Tensile
         }
     };
 
+    template <typename MyProblem, typename MySolution>
+    struct PerCUorOverallLibrary
```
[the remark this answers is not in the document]
benjaminulmer: Yup, Scott is right. I originally made it distinct to future-proof for changes that would need to be made to support the API for selecting perCU or overall, but now I know this can/will be handled by a switch in ContractionProblem.

Tensile/Source/lib/include/Tensile/ContractionProblemPredicates.hpp
cgmillette: How about "Global efficiency" versus "Local efficiency" ? Kind of alludes to 'overall' versus 'per-cu'.

cgmillette:
Can you easily separate the bug fix and submit a different PR ? I think that issue on it's own will have a sufficient importance to be tracked separately. Hoping that it will also generate a unit test to protect against regression.

Tensile/LibraryIO.py
```diff
@@ -132,7 +132,7 @@ def readLibraryLogicForSchedule( filename ):
 
   # verify
   if len(data) < 6:
-    printExit("len(%s) %u < 7" % (filename, len(data)))
+    printExit("len(%s) %u < 6" % (filename, len(data)))
```
cgmillette, line 135: What does '6' or '7' mean? Don't have enough data?
Can we improve this with a more verbose message?

Tensile/Source/client/include/ClientProblemFactory.hpp
```diff
@@ -70,17 +70,18 @@ namespace Tensile
...
+            DataType        m_bType;
+            DataType        m_cType;
+            DataType        m_dType;
+            DataType        m_alphaType;
+            DataType        m_betaType;
+            bool            m_stridedBatched;
+            bool            m_highPrecisionAccumulate;
+            bool            m_deterministicMode;
+            ArithmeticUnit  m_arithmeticUnit;
+            BenchmarkMetric m_benchmarkMetric;
```
cgmillette, line 83: Should we rename BenchmarkMetric to PerformanceMetric?

Tensile/Source/lib/include/Tensile/BenchmarkMetricTypes.hpp
```diff
@@ -0,0 +1,150 @@
...
+    };
+
+    /**
+ * \ingroup BenchmarkMetrics
+ * \brief Compile-time accessible benchmark metric type metadata.
+ */
+    template <BenchmarkMetric T_Enum>
+    struct BenchmarkMetricInfo
+    {
+    };
```
cgmillette, line 99: Should be able to leave this just as a declaration:

template <BenchmarkMetric T_Enum>
struct BenchmarkMetricInfo;

Tensile/Source/lib/include/Tensile/ContractionProblem.hpp
```diff
@@ -725,14 +736,15 @@ namespace Tensile
...
-        KernelLanguage m_kernelLanguage          = KernelLanguage::Any;
+        bool            m_transA;
+        bool            m_transB;
+        bool            m_stridedBatched          = true;
+        bool            m_highPrecisionAccumulate = false;
+        bool            m_deterministicMode       = false;
+        bool            m_eligibleForPK           = true;
+        ArithmeticUnit  m_arithmeticUnit          = ArithmeticUnit::Any;
+        BenchmarkMetric m_benchmarkMetric         = BenchmarkMetric::Overall;
+        KernelLanguage  m_kernelLanguage          = KernelLanguage::Any;
```
cgmillette, line 747: Same as above - ContractionProblem is used in many places outside of benchmarking, might be more intuitive to call it 'performance metric'

Tensile/Source/lib/include/Tensile/ContractionProblemPredicates.hpp
```diff
@@ -1055,6 +1055,41 @@ namespace Tensile
...
+
+            struct CUEfficiency : public Predicate_CRTP<CUEfficiency, ContractionProblem>
+            {
+                enum
+                {
+                    HasIndex = false,
+                    HasValue = false
+                };
+                bool value;
+
```
cgmillette, line 1067: HasValue = false, we can remove the value :)

tuning/automation/ConvertToEfficiency.py
```diff
@@ -0,0 +1,116 @@
...
+        fullPath = os.path.join(startDir, filename)
+        if os.path.isdir(fullPath):
+            files = files + allFiles(fullPath)
+        else:
+            files.append(fullPath)
+    return files
+
+def peakGFlops(sclk, alu, numCUs):
+    return (sclk / 1000) * alu * numCUs
+
```
cgmillette, line 65: Might be a good idea just for numbers sake to indicate the units of sclk (=MHz in this case), alu (= wave size?)

Tensile/LibraryIO.py
```diff
@@ -264,7 +264,10 @@ def _getLibraryLogicForSchedule(self, schedulePrefix, architectureName, deviceNa
       tileSelectionIndices = logicTuple[6]
       tileSelectionLogic["TileSelectionIndices"] = tileSelectionIndices
       data.append(tileSelectionLogic)
+    else:
+      data.append(None)
```
cgmillette, line 269: Are we prepared to handle "None" objects in parsing?

Tensile/LibraryIO.py
benjaminulmer: Initially I just changed the message to match the condition without looking any further into it. Upon further inspection, this condition hasn't been updated in ~4 years; I'll update it to be current and improve the message.

Tensile/Source/lib/include/Tensile/ContractionProblemPredicates.hpp
benjaminulmer: Haha good catch

tuning/automation/ConvertToEfficiency.py
benjaminulmer: `alu` is Flops/cycle/CU here. I'll add a comment documenting the units.

Tensile/LibraryIO.py
benjaminulmer: As far as I can tell, yes. It ends up coming out as `null` in the yaml file. Having the else branch here ensures that `logicTuple[7]` (the Benchmark/PerformanceMetric) ends up at a consistent index in the library logic file.

All the local testing I did follows the else branch and works. What I haven't tested is having tile selection indices _and_ a BenchmarkMetric, although I can't think of any reasons why this wouldn't work.

benjaminulmer:
> Can you easily separate the bug fix and submit a different PR ? I think that issue on it's own will have a sufficient importance to be tracked separately. Hoping that it will also generate a unit test to protect against regression.

The multiple rows for the same hardware bug? It should be easy enough to isolate, and I don't think anything else in this PR depends on it.

benjaminulmer:
> Can you easily separate the bug fix and submit a different PR ? I think that issue on it's own will have a sufficient importance to be tracked separately. Hoping that it will also generate a unit test to protect against regression.

The multiple rows for the same hardware bug? It should be easy enough to isolate, and I don't think anything else in this PR depends on it.

Tensile/Source/lib/include/Tensile/BenchmarkMetricTypes.hpp
benjaminulmer: I'll be honest, I didn't take the time to fully understand this code and just copied and modified from `ArithmeticUnitTypes.hpp/cpp` and `KernelLanguageTypes.hpp/cpp`. You seem to be correct though, so I'll update this and the other two files as well.

Tensile/Source/lib/include/Tensile/ContractionProblem.hpp
benjaminulmer: I don't have a strong preference here. I'll make the change to PerformanceMetric unless someone else has strong feelings on what it should be called.

Tensile/Source/client/source/BenchmarkTimer.cpp
```diff
@@ -100,13 +100,19 @@ namespace Tensile
...
 
-            double   gflops     = m_problem.flopCount() / (timePerEnqueue_us) / 1000.0;
-            uint64_t gflopsUint = static_cast<uint64_t>(round(gflops));
-
             ContractionSolution::ProjectedPerformance pp
                 = m_solution.projectedPerformance(m_problem, m_hardware);
 
+            double gflops      = m_problem.flopCount() / (timePerEnqueue_us) / 1000.0;
+            double tiles       = pp.tilesPerCu * perf.CUs;
+            int    usedCus     = (tiles > perf.CUs) ? perf.CUs : tiles;
```
cgmillette, line 108: could use min (tiles, perf.CUs)

Tensile/Source/client/source/ResultFileReporter.cpp
```diff
@@ -35,19 +35,26 @@ namespace Tensile
...
+                                               PerformanceMetric  performanceMetric)
             : m_output(filename)
             , m_extraCol(exportExtraCols)
             , m_mergeSameProblems(mergeSameProblems)
+            , m_performanceMetric(performanceMetric)
         {
-            m_output.setHeaderForKey(ResultKey::ProblemIndex, "GFlops");
+            if(m_performanceMetric == PerformanceMetric::CUEfficiency)
+                m_output.setHeaderForKey(ResultKey::ProblemIndex, "GFlopsPerCU");
+            else // (m_performanceMetric == PerformanceMetric::Overall)
```
cgmillette, line 56: Just checking if this is still valid for 'Best' or 'Count' enum values

tuning/automation/ConvertToEfficiency.py
```diff
@@ -0,0 +1,118 @@
...
+         "specs"   : ".yaml file containing hardware specifications", \
+         "per-cu"  : "If tuning was done per CU", \
+         "name"    : "Name substring to filter which files are modified", \
+         "mfma"    : "If MFMA instructions were used for tuning", \
+         "mi50"    : "For vega20, if tuning was done on mi50"
+    }
+
+    argParser.add_argument("inDir", metavar="input-dir", type=str, help=h["inDir"])
+    argParser.add_argument("outDir", metavar="output-dir", type=str, help=h["outDir"])
+    argParser.add_argument("sclk", type=int, help=h["sclk"])
```
cgmillette, line 43: Can you quickly update the help message with the units as well ? (MHz)

cgmillette (approved):
Overall looks good to me, just a couple tiny details. Thanks Ben!

Tensile/Source/client/source/BenchmarkTimer.cpp
benjaminulmer: Fixed

Tensile/Source/client/source/ResultFileReporter.cpp
benjaminulmer: Only CUEfficiency and Overall make sense at the benchmarking stage. I updated the comment (and added one in LogReporter.hpp) clarifying this.

The other option is to add an error/warning message if something other than one of these two values is passed in.

zaliu:
@benjaminulmer please investigate CI failures and resolve them soon.  Thanks.

benjaminulmer:
> @benjaminulmer please investigate CI failures and resolve them soon. Thanks.

Silly mistake on my part; should be fixed now. I'll watch to make sure CI passes.

PR #1280, opened 2021-02-18
MasterSolutionLibrary duplicated hardware rows fix

benjaminulmer:
Fixed a bug where the master solution library would have multiple rows for the same hardware. This was caused by the less than operator of HardwarePredicate modifying its own `__class__`, which caused later equality checks with other HardwarePredicates to return false negatives due to differing `__class__` values.

Originally fixed in a draft of #1279, this change has been isolated for idependent tracking.


sdquiring (approved):

cgmillette (approved):
LGTM

Issue #1172, opened 2020-09-28
Stop calling dpkg


haampie:
Could you please stop calling `dpkg` as part of the build?

I'm referring to: https://github.com/ROCmSoftwarePlatform/Tensile/blob/rocm-3.8.0/Tensile/Common.py#L1516-L1536

This suddenly deals with a system-wide installation, and as outlined in the comment, _will not_ work with different package managers, in particular package managers you don't support (AUR / Spack, snap? etc etc). Also, it will not work when _multiple_ versions of rocm packages are installed next to one another; this is what some package managers allow you to do. Also, installations might not be system-wide at all.

Why wouldn't you just call `hipcc --version` / `hcc --version` instead?

haampie:
This is still the case in 4.0.0 unfortunately. Has this issue been noted? (Pinging @TorreZuk :sweat_smile:)

https://github.com/ROCmSoftwarePlatform/Tensile/blob/rocm-4.0.0/Tensile/Common.py#L1575-L1592

TorreZuk:
Sorry @haampie it looks like this will still be true for 4.1, I would hope someone can fix it for 4.2.   Agreed just using hipcc --version or the clang++ --version is better.   I'll also push to see if this can be assigned for the current sprint.

zaliu:
Thanks @haampie for reporting it.  We'll get it fixed for 4.2.  I'll keep you posted.

JeremyAdamHart:
@zaliu, Looks like this issue can be closed? I see that the fix is in the most recent release.

zaliu:
Yes, but normally the person reporting the issue gets t close the ticket.  @haampie?

haampie:
Looks like it! Thanks!

PR #1283, opened 2021-02-19
Fix calling dpkg

powderluv:
Remove dpkg dependency so you can point to AOMP etc
Remove hcc as it is deprecated.

Fixes:https://github.com/ROCmSoftwarePlatform/Tensile/issues/1172

TEST: Build with latest hipcc from AOMP. Tensile will report:
Found  hipcc version 4.0.20496-4f163c68


zaliu (approved):
Thanks for fixing this one!

AlexBrownAMD (approved):

PR #1284, opened 2021-02-19
Remove hcc support

AlexBrownAMD:
Since hcc has been deprecated for a while now, this change removes support in tensile. This change currently leaves options and global params like --cxx-compiler and CxxCompiler in place.


zaliu (approved):
The shortage of gfx908 CI executors is really getting to be a problem, @saadrahim @eidenyoshida .  The weekend planned shutdown didn't help.

saadrahim:
CI is not running on this due to the branch conflict. Not sure if gfx908 shortages are impacting this.

zaliu:
There wasn't any merge conflict when the timeouts took place earlier this morning and yesterday.  I just merged in a guest PR that caused this merge conflict within the past hour or so.

cgmillette (approved):
LGTM. 
Only thing I would ponder is whether to keep the tensile option for cxx-compiler in case we needed to need to differentiate with a future release again with another supported compiler. Otherwise, nice to clean a bunch of stuff up :)

AlexBrownAMD:
> LGTM.
> Only thing I would ponder is whether to keep the tensile option for cxx-compiler in case we needed to need to differentiate with a future release again with another supported compiler. Otherwise, nice to clean a bunch of stuff up :)

Thanks for reviewing! Yea, I left the cxx-compiler option in for now, we can review if it should also be removed later.

PR #1289, opened 2021-02-24
Use LLVM bundled with rocm for Tensile

JeremyAdamHart:
ROCm comes packaged with llvm-12.0, however Tensile requires llvm-6.0 for the HostLibraryTests. The library however can already build with just llvm-12.0 installed. To eliminate the need to have multiple versions of llvm installed for the tests, they now require llvm 12.0 (Which should always be present), and I propagated the same change over to the library to prevent conflicts when the library finds llvm-6.0 first.

### DIFF
--- a/HostLibraryTests/CMakeLists.txt
+++ b/HostLibraryTests/CMakeLists.txt
@@ -127,7 +127,7 @@ set(test_sources ${test_sources}
 )
 
 if(TENSILE_USE_LLVM)
-    find_package(LLVM 6.0 REQUIRED CONFIG)
+    find_package(LLVM 12.0 REQUIRED CONFIG)
 
     set(test_sources ${test_sources}
         ContractionLibraryLoading_test.cpp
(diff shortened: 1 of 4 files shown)


TorreZuk (commented):
You might want to follow the old pattern and put QUIET CONFIG 12.0 and as we are now on 13 and add the if (NOT... fallback REQUIRED CONFIG 13.0 for both cases.   REQUIRED probably doesn't need QUIET does it?

sdquiring (commented):
Great! It's probably worth creating a rocBLAS PR with these changes to verify it on all the platforms run in rocBLAS CI.

zaliu (changes requested):
llvm 13.0 support must be added.  Thanks.

JeremyAdamHart:
I added support for llvm-13 and added back the fallback pattern from before. The find_package in the HostLibraryTests wouldn't work with the fallback strategy because LLVM_DIR becomes unset after the first failed attempt, and /opt/rocm would no longer be in the search path, so I added /opt/rocm into the prefix. 

I also updated the docker files to remove llvm-6.0.

TorreZuk (approved):
Looks like it could be a robust strategy.  I guess the last case may be your future proofing and legacy support.

zaliu (approved):
Please merge only after @amcamd announces the all-clear to do so.  Please use squash-merge.  Thanks.
p.s. In the mean time, it'd be great if you can confirm that rocBLAS build succeeds with your PR on CentOS 7.7/8.1 and SLES if you can.

PR #1290, opened 2021-02-25
tuning config & logic files for rocBLAS-internal #560

zaliu:
Winners data 2_BenchmarkData/ are compressed due to its size.
This PR is purely for archival purposes; no source code impact to speak of.


AlexBrownAMD (approved):

AlexBrownAMD:
github says 2_BenchmarkData.tar.gz is an empty file?

benjaminulmer:
> github says 2_BenchmarkData.tar.gz is an empty file?

Says the same for me, but when I download the raw it looks like everything is in place. I guess it's an issue with GitHub?

benjaminulmer (approved):

zaliu:
Thanks for noticing this, guys!  Short of downloading the compressed files, one indication that they really aren't empty would be when we click on View File, the size of the file shown isn't zero.

PR #1296, opened 2021-03-04
Option to print tensor data in hex format

AlexBrownAMD:


benjaminulmer (approved):
The template specializations for float and double are annoying, but I can't think of a better way of doing it. In other words, looks good.

Tensile/Common.py
```diff
@@ -178,6 +178,7 @@
 globalParameters["PrintTensorC"] = 0          # Print TensorC.  0x1=after init; 0x2=after copy-back; 0x3=both
 globalParameters["PrintTensorD"] = 0          # Print TensorD.  0x1=after init; 0x2=after copy-back; 0x3=both
 globalParameters["PrintTensorRef"] = 0          # Print reference tensor.  0x1=after init; 0x2=after copy-back; 0x3=both
+globalParameters["PrintTensorHex"] = 0        # Print tensor data in hex format
 globalParameters["PrintIndexAssignments"] = 0      # Print the tensor index assignment info
```
cgmillette, line 182: Just curious about not using boolean values here. Below, booleans are used.

Tensile/Source/lib/include/Tensile/TensorDescriptor.hpp
```diff
@@ -342,11 +342,47 @@ namespace Tensile
 
     std::ostream& operator<<(std::ostream& stream, const TensorDescriptor& t);
 
+    template <typename T>
+    inline void WriteTensor1DHex(std::ostream& stream, T* data, TensorDescriptor const& desc)
+    {
```
cgmillette, line 347: Might be able to write a smaller function to do the stream treatments:
template<typename T>
inline void WriteElement(std::ostream& stream, T* data, int index){  stream << data[index]; }

template<>
inline void WriteElement<float>(std::ostream& stream, float* data, int index){  stream << *reinterpret_cast<const uint32_t*>(&data[i]); }

...

And then plug this into the WriteTensor1DHex function to avoid specializing this bigger function and copying loops.

Tensile/Source/lib/include/Tensile/TensorDescriptor.hpp
```diff
@@ -355,16 +391,187 @@ namespace Tensile
...
-            stream << " " << data[i];
+            for(size_t i = 1; i < desc.sizes()[0]; i++)
+                stream << " " << data[i];
+        }
 
         if(decorated)
             stream << "]" << std::endl;
     }
 
+    template <typename T>
```
cgmillette, line 409: Same as above. Should be able to solve a lot of the repetition.

babakpst (dismissed):
Blocked until after FC merge.

babakpst (changes requested):
Blocked until FC merges are complete.

babakpst:
Closing this PR, as this is not a priority now.

PR #587, opened 2019-07-04
Add new feature to support vpgr limit at training.

PeterJkPeng:

PR #1299, opened 2021-03-09
Rename PerformanceMetric enum values

benjaminulmer:
Renamed `PerformanceMetric` enums: Best -> Auto and Overall -> DeviceEfficiency.
This was done to keep the names consistent between Tensile and rocBLAS (see [rocBLAS-internal #587](https://github.com/ROCmSoftwarePlatform/rocBLAS-internal/pull/578)).

Since no CUEfficiency solution files are in rocBLAS yet, this change should have no effect.


cgmillette (approved):
LGTM, as long as CI is good!

JeremyAdamHart:
LGTM too. I think auto is more clear than best as well.

PR #1300, opened 2021-03-11
Initial implementation support for aldebaran

yoichiyoshida:
Add initial support for aldebaran (gfx90a)

- adds initial support for double precision MFMA
- adds alignment restrictions for 64 bit wide register instructions for gfx90a compatibility
- adds MFMA bf16_1k instruction support


zaliu (changes requested):
In the interest of not blocking gfx10 merging for too long, **flat addressing tests in extended CI can be disabled for now.**  We need to remember to fix this issue though.  At one point in the past, we relied on flat addressing for very large matrices; not anymore.  We might need flat addressing in the future for different reasons.  Thanks.

Tensile/KernelWriterAssembly.py
```diff
@@ -1620,6 +1628,8 @@ def initKernel(self, kernel, tPA, tPB ):
     PLR = kernel["PrefetchLocalRead"] if kernel["PrefetchLocalRead"] < kernel["LoopIters"] else kernel["LoopIters"] - 1
     valuBlocks = (1+PLR) * kernel["InnerUnroll"]
 
+    # TODO: alignment hack, figure out a better solution
+    vgprIdx = ((vgprIdx+1)//2)*2
```
AlexBrownAMD, line 1632: Leftover todo stuff

Tensile/Source/client/main.cpp
```diff
@@ -491,7 +489,7 @@ int main(int argc, const char* argv[])
     {
         listeners.addListener(std::make_shared<ReferenceValidator>(args, dataInit));
         listeners.addListener(std::make_shared<BenchmarkTimer>(args, *hardware));
-        listeners.addListener(std::make_shared<HardwareMonitorListener>(args));
+        //listeners.addListener(std::make_shared<HardwareMonitorListener>(args));
```
AlexBrownAMD, line 492: Is it intended to always disable this?

AlexBrownAMD: (Also I think the default global param was changed)

sdquiring (approved):
Looks good to me.

AlexBrownAMD (approved):

zaliu (approved):
Thanks for the good work @yoichiyoshida and @AlexBrownAMD!

PR #1302, opened 2021-03-12
Custom kernel alternative to replacement kernels

JeremyAdamHart:
I've left the Replacement Kernel system intact and introduced a new alternative which should allow for an easier introduction of user specified kernels (Hopefully allowing for replacement kernels to be deprecated at some point). Part of the reason for this is due to the fact that in this system the custom kernels aren't "replacing" anything, and the name replacement kernel is misleading.

### Benchmarking
Custom kernels are benchmarked alongside conventionally generated kernels within its BenchmarkProblemSizeGroup. To include a custom kernel, list its name (Without file extension) under the "CustomKernels" category. During the benchmarking process, it will find the assembly kernel with this name (within the CustomKernels directory by default), and use it directly -- bypassing kernel generation. Beyond this, it is treated exactly like a normal kernel.

Example config file:
```BenchmarkProblems:
  - # sgemm NN
    - # ProblemType
      OperationType: GEMM
      DataType: s
      TransposeA: False
      TransposeB: False
      UseBeta: True
      Batched: True

    - # BenchmarkProblemSizeGroup
      CustomKernels: ["TestKernel"]
      InitialSolutionParameters:
      BenchmarkCommonParameters:
        - KernelLanguage: ["Assembly"]
        - WorkGroupMapping: [1]
      ForkParameters:
        - ThreadTile:
          - [ 8, 8 ]
        - WorkGroup:
          - [  8, 16,  1 ]
        - VectorWidth: [4, 1]
        - DepthU: [ 8, 4 ]
      JoinParameters:
      BenchmarkJoinParameters:
      BenchmarkFinalParameters:
        - ProblemSizes:
          - Exact: [ 256, 256, 1, 256 ]
          - Exact: [ 512, 256, 1, 256 ]
```

### Kernel specification
To specify a custom kernel, in addition to writing the assembly kernel, the creator must include a custom.config section within the yaml block, outlining the parameters that the kernel uses. Tensile will infer parameters in the same manner it would during kernel generation, so if the creator started from a Tensile generated kernel, the configuration parameters used to generate that kernel would be sufficient. To avoid Tensile incorrectly inferring parameters,  ```AssignedDerivedParameters: True``` can be included in the list. Asserts can also be included here to specify under which conditions the kernel is valid. For example, the following kernel would only be selected with m and n < 257 and k < 1025
```
<start of kernel>
.amdgpu_metadata
---
custom.config:
  ProblemType:
    OperationType: GEMM
    DataType: s
    TransposeA: False
    TransposeB: False
    UseBeta: True
    Batched: True
  LoopDoWhile: False
  WorkGroupMapping:  1
  ThreadTile: [ 8, 8 ]
  WorkGroup: [  8, 16,  1 ]
  DepthU: 8
  VectorWidth: 4
  AssertSizeLessThan: {0: 257, 3: 1025 }
  AssertSizeMultiple: {1: 512}
  AssertCEqualsD: True
  AssertAlphaValue: -1
  AssertBetaValue: 1 
<remainder of kernel>
...
``` 
Additionally, if this kernel has branched off of a Tensile generated assembly kernel, all references to the original kernel name must be changed to the new kernel's name (without the file extension).

### Additonal changes

To allow custom kernels to be able to be used in all the same use cases (to my knowledge) as replacement kernels, I've added SizeGreaterThan and SizeLessThan predicates to specify the range in which these kernels should be used.

### Limitations
- If the user wants the replacement kernel to always be selected when it's valid, this system does not currently enable that. This will likely be added in a future extension.
- If the replacement kernel is slower than a generated kernel, it may not not be included in the library at all. An option to force it to be selected as winner for the sizes its benchmarked for could address this, although maybe a solution to the first issue may be a better way to address this as well.

### New Changes (April 21st)

Made a number of changes to the system to skip more unnecessary parts of the Tensile benchmarking process for custom kernels. I also added a NoReject kernel option which can be used to prevent a CustomKernel from being rejected during parameter derivation due to what Tensile detects as invalid parameters. Note that it is possible that this still may cause Tensile to abort assignDerivedParameters before having derived every parameter that is needed, so it may be necessary to define extra parameters in the custom.config section when this is used. As before, AssignedDerivedParameters: False can be used to skip this entire process, but requires more input for the custom.config section.

### New predicates (April 21st)

In addition to the predicates added earlier, I've also added AssertSizeMultiple, AssertAlphaValue, AssertBetaValue and AssertCEqualsD parameters, each of which assign a predicate to the kernel. When using the AssertAlphaValue and AssertBetaValue parameters during benchmarking, you must set globalParameters["DataInitTypeAlpha"] and globalParameters["DataInitTypeBeta"] to a matching value. The same is true with AssertCEqualsD, globalParameters["CEqualD"] must be used for this assert to pass.


zaliu (commented):
This is a great first step.  However, how do we make sure rocBLAS build can and will pick up this replacement kernel?  Please explain.  Thanks.

benjaminulmer:
> This is a great first step. However, how do we make sure rocBLAS build can and will pick up this replacement kernel? Please explain. Thanks.

Jeremy can correct me if I'm wrong, but it looks like it uses the same mechanism as the current replacement kernel system through an update to `KernelWriter.getReplacementKernelPath`. Essentially, if a solution in a library logic file has a non-empty CustomKernelName, this signals that it's a custom kernel and `TensileCreateLibrary` handles it appropriately.

benjaminulmer (commented):
Played around with it a bit—seems to be working well. Code changes lgtm so far.

JeremyAdamHart:
> 
> 
> Jeremy can correct me if I'm wrong, but it looks like it uses the same mechanism as the current replacement kernel system through an update to `KernelWriter.getReplacementKernelPath`. Essentially, if a solution in a library logic file has a non-empty CustomKernelName, this signals that it's a custom kernel and `TensileCreateLibrary` handles it appropriately.

@zaliu This is correct, I don't believe there should be any issues in this regard, but I was trying to look into it a little bit just to make sure. I'm going to swap in the custom kernel into the logic files of a rocblas build just to be 100% sure it's finding it correctly. I'm assuming that the rocblas replacement kernels are stored in Tensile/ReplacementKernels[-cov3] directory and there isn't anything special done with them? If so, this should work the same, just instead be found in the Tensile/CustomKernels (Or wherever the user sets globalParameters["CustomKernelDirectory"] to)

JeremyAdamHart:
Just to update, I can confirm that everything works after adding the kernel to the rocBLAS logic file.

JeremyAdamHart:
Currently I have the TN 64x128x32 MDA2 kernel translated into a custom kernel using the parameters in the kernel name (With the multiple and greater than assertions added). This probably doesn't match the parameters that the kernel _should_ have (and I think that it would be a good idea to make sure it has the correct parameters before merging), but as it stands currently, I believe its been given the parameters the replacement kernel version would have had. I have a script for this now, and if it's useful I can do the same with the other kernels to make sure they have the same configurations the kernel they were replacing would have had.

Tensile/Tests/unit/test_CustomKernels.py
```diff
@@ -0,0 +1,78 @@
+################################################################################
+# Copyright 2020 Advanced Micro Devices, Inc. All rights reserved.
```
yoichiyoshida, line 2: fix year

Tensile/CustomKernels.py
```diff
@@ -0,0 +1,59 @@
...
+        else              : assembly += line + "\n"; config += "\n"  #Second statement to keep line numbers consistent for yaml errors
+
+    return (config, assembly)  
+
+def getCustomKernelConfig(name, directory=globalParameters["CustomKernelDirectory"]):
+    rawConfig, _ = getCustomKernelConfigAndAssembly(name, directory)
+    try:
+        return yaml.safe_load(rawConfig)["custom.config"]
+    except yaml.scanner.ScannerError as e:
+        raise RuntimeError("Failed to read configuration for custom kernel: {0}\nDetails:\n{1}".format(name, e))
```
yoichiyoshida, line 59: add newline at the end of the file

Tensile/CustomKernels.py
```diff
@@ -0,0 +1,59 @@
+################################################################################
+# Copyright 2020-2021 Advanced Micro Devices, Inc. All rights reserved.
```
yoichiyoshida, line 2: new file year should only include current year

Tensile/Source/lib/source/ScalarValueTypes.cpp
```diff
@@ -0,0 +1,139 @@
+/*******************************************************************************
+ *
+ * MIT License
+ *
+ * Copyright 2019-2020 Advanced Micro Devices, Inc.
```
yoichiyoshida, line 5: Fix year

Tensile/Source/lib/source/ScalarValueTypes.cpp
```diff
@@ -0,0 +1,139 @@
...
+    std::istream& operator>>(std::istream& stream, ScalarValue& t)
+    {
+        std::string strValue;
+        stream >> strValue;
+
+        t = ScalarValueTypeInfo::Get(strValue).m_value;
+
+        return stream;
+    }
+} // namespace Tensile
```
yoichiyoshida, line 139: add newline at the end of the file

Tensile/Source/lib/include/Tensile/ScalarValueTypes.hpp
```diff
@@ -0,0 +1,163 @@
...
+{
+    template <>
+    struct hash<Tensile::ScalarValue>
+    {
+        inline size_t operator()(Tensile::ScalarValue const& val) const
+        {
+            return hash<int>()(static_cast<int>(val));
+        }
+    };
+} // namespace std
```
yoichiyoshida, line 163: add newline

Tensile/Source/lib/include/Tensile/ScalarValueTypes.hpp
```diff
@@ -0,0 +1,163 @@
+/*******************************************************************************
+ *
+ * MIT License
+ *
+ * Copyright 2019-2021 Advanced Micro Devices, Inc.
```
yoichiyoshida, line 5: fix year

yoichiyoshida (approved):
Code LGTM, some minor formatting issues and minor spacing weirdness I didn't point out specifically. I have tested the code works for my own sample test kernel.

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # Change Log for Tensile
 
+
```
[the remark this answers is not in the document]
benjaminulmer: Now that this is ready for review though, the changelog should probably be updated with the new features added by this PR

zaliu (approved):

PR #1306, opened 2021-03-18
New Python unit tests for predicates and predicate libraries

benjaminulmer:
`test_hardware_library_merge_dups` motivated by #1280.

Other tests added to help ensure fallback to Device Efficiency kernels when a CU efficiency kernel cannot be found happens correctly (see #1279).

### DIFF
--- a/Tensile/Tests/unit/test_HardwarePredicates.py
+++ b/Tensile/Tests/unit/test_HardwarePredicates.py
@@ -101,3 +101,12 @@ def test_hardware_library_merge_dups(libraries):
         lib.merge(lib2)
 
     assert len(lib.rows) == 4
+
+    def getPred(row):
+        return row['predicate']
+    rowPreds = map(getPred, lib.rows)
+
+    assert HardwarePredicate.FromISA((9,0,0)) in rowPreds
+    assert HardwarePredicate.FromISA((9,0,6)) in rowPreds
+    assert HardwarePredicate.FromISA((9,0,8)) in rowPreds
+    assert HardwarePredicate('TruePred') in rowPreds
--- a/Tensile/Tests/unit/test_HardwarePredicates.py
+++ b/Tensile/Tests/unit/test_HardwarePredicates.py
@@ -1,5 +1,5 @@
 ################################################################################
-# Copyright 2020 Advanced Micro Devices, Inc. All rights reserved.
+# Copyright 2020-2021 Advanced Micro Devices, Inc. All rights reserved.
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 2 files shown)


Tensile/Tests/unit/test_HardwarePredicates.py
```diff
@@ -82,3 +83,21 @@ def test_predicate_library_merge(libraries):
...
+
+    return [copy.deepcopy(libs) for libs in itertools.permutations(objs)]
+
+@pytest.mark.parametrize("libraries", hardware_library_objects_dups())
+def test_hardware_library_merge_dups(libraries):
+    lib = libraries[0]
+    for lib2 in libraries[1:]:
+        lib.merge(lib2)
+
+    assert len(lib.rows) == 4
```
JeremyAdamHart, line 103: This will probably be good enough, but it might be more reliable to test against the expected result, since it's possible for the merge  function to remove the wrong library and still pass.

Tensile/Tests/unit/test_PerfMetricPredicates.py
```diff
@@ -0,0 +1,67 @@
...
+
+    return [copy.deepcopy(libs) for libs in itertools.permutations(objs)]
+
+@pytest.mark.parametrize("libraries", perf_metric_library_objects_dups())
+def test_perf_metric_library_merge_dups(libraries):
+    lib = libraries[0]
+    for lib2 in libraries[1:]:
+        lib.merge(lib2)
+
+    assert len(lib.rows) == 2
```
JeremyAdamHart, line 67: Same comment here

JeremyAdamHart (commented):
Aside from the two comments, lgtm!

Tensile/Tests/unit/test_HardwarePredicates.py
benjaminulmer: Sure. It's a pretty unlikely edge case, but testing to ensure against it should be simple enough.

cgmillette (changes requested):
While it's good that we're testing the general expectations in terms of resulting library row count etc., the verification should include the specific expected outputs as well. Like Jeremy mentioned, we should check the specific expected output because we have an explicit input case. Gotta test it like a 'black box' and not trust it will do anything correctly :)

benjaminulmer:
> While it's good that we're testing the general expectations in terms of resulting library row count etc., the verification should include the specific expected outputs as well. Like Jeremy mentioned, we should check the specific expected output because we have an explicit input case. Gotta test it like a 'black box' and not trust it will do anything correctly :)

Agreed. I just updated this in the latest commit.

cgmillette (approved):
LGTM

JeremyAdamHart (approved):
Looks all good to me too!

PR #1312, opened 2021-03-22
LibraryIO.py refactor and tests

benjaminulmer:
Refactor of LibraryIO.py:
- Removed duplicated code
- Separated file reading and data parsing into own functions
- Renamed functions to be more descriptive and consistent
- Better error messages
- Oneline docstrings describing use of each function

Basic tests for LibraryIO:
- Currently only test parsing of solution and library logic file data
- Tests only check that functions execute without fail: no validation of output is currently done
- Main purpose for now is to ensure all library logic file types which can be expected to be in rocBLAS---legacy, DeviceEfficiency, and CUEfficiency---are parsable

Potential future expansions:
- More in depth testing (ideas on how to best test these functions welcome)
- Better parameter and return types for functions
  - e.g. Custom objects instead of long lists and tuples
  - Alternatively, simply document expected values of parameters better
- Better error reporting mechanism that doesn't requiring passing data/file source into parsing functions

### DIFF
--- a/Tensile/LibraryIO.py
+++ b/Tensile/LibraryIO.py
@@ -24,7 +24,6 @@ from .SolutionStructs import Solution, ProblemSizes, ProblemType
 from . import __version__
 from . import Common
 from . import SolutionLibrary
-import os
 
 try:
     import yaml
--- a/Tensile/LibraryIO.py
+++ b/Tensile/LibraryIO.py
@@ -193,8 +193,8 @@ def rawLibraryLogic(data):
 
     dataLength = len(data)
     if dataLength > 9:
-            for idx in range(9, dataLength):
-                    otherFields.append(data[idx])
+        for idx in range(9, dataLength):
+            otherFields.append(data[idx])
 
     return (versionString, scheduleName, architectureName, deviceNames,\
             problemTypeState, solutionStates, indexOrder, exactLogic, rangeLogic, otherFields)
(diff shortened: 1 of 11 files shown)


Tensile/LibraryIO.py
```diff
@@ -19,212 +19,192 @@
...
+    solutionStates    = data[5]
+    indexOrder        = data[6]
+    exactLogic        = data[7]
+    rangeLogic        = data[8]
+    otherFields       = []
+
+    dataLength = len(data)
+    if dataLength > 9:
+            for idx in range(9, dataLength):
+                    otherFields.append(data[idx])
```
AlexBrownAMD, line 197: Double indents

Tensile/Tests/unit/test_LibraryIO.py
```diff
@@ -0,0 +1,119 @@
...
+    # paths to test data
+    scriptDir = os.path.dirname(os.path.realpath(__file__))
+    dataDir = os.path.realpath(os.path.join(scriptDir, "..", "test_data", "unit"))
+    solutionsPath = os.path.realpath(os.path.join( \
+            dataDir, "solutions", "solutions_nn_3.yaml"))
+
+    solutions = LibraryIO.readYAML(solutionsPath)
+
+    LibraryIO.parseSolutionsData(solutions, "test_parseSolutionsData")
+    assert True
```
AlexBrownAMD, line 104: All tests assert true, is this anticipating a crash in LibraryIO on failure to parse?

benjaminulmer: Precisely. I'm leaving expanding these tests with more checks as future TODO unless others deem necessary for this PR to get merged.

AlexBrownAMD (approved):

PR #1313, opened 2021-03-24
Swap MFMA source A and B to optimize store code

AlexBrownAMD:
Current TODOs include fixing dgemm edge code with this change, and adding support for other data types like sgemm.


Tensile/KernelWriterAssembly.py
```diff
@@ -10325,30 +10336,45 @@ def chooseGlobalRead(self, useBuffer, bpl, destVgpr, \
       extraFields += " glc, slc, dlc"
 
     if useBuffer:
+      rv = Code.Module("Global Read")
       tailFields = "offen offset:%u"%offset
+      if offset > 4095:
```
yoichiyoshida, line 10341: Maybe change that to >= 4096 and a comment for 4k limit so it makes sense to others?

Tensile/KernelWriterAssembly.py
```diff
@@ -10391,28 +10417,34 @@ def chooseGlobalWrite(self, useBuffer, bps, srcVgpr, rpv, \
     kStr = ""
 
     if useBuffer:
+      tmpSgpr = 0
+      if offset > 4095:
```
yoichiyoshida, line 10421: same here

yoichiyoshida (approved):

cgmillette (approved):

PR #1315, opened 2021-03-26
merge staging 18875ad6 into master with GO from CQE

zaliu:


amcamd (approved):
LGTM, the following two compares indicate to me that this commit to master has the same content as 18875ad6. Both compares show zero changed files. 
- https://github.com/zaliu/Tensile/compare/master...zaliu:18875ad6
- https://github.com/ROCmSoftwarePlatform/Tensile/compare/18875ad6...zaliu:master

PR #1316, opened 2021-03-26
Update to merge script allowing mfma and non-mfma solution for sizes

JeremyAdamHart:
This change allows for a MFMA and non-MFMA solution to both be present in a logic file for a given problem size. When the option ```--add_mfma_tag``` is used, a SolutionTag will be prepended to problem key, resulting in a different key for the MFMA and non-MFMA versions, regardless of whether the sizes match. This tag is only present during the merging process, and is removed before the new logic file is written.

This feature can be used with ContractionProblem::setArithmeticUnit to select which type of kernel is used.

Sample output:
```
Base logic file: Orig/Test.yaml | Incremental: Inc/Test.yaml | Merge policy: Winner | Trim size: True | Add MFMA tag: True 
6 sizes and 2 kernels in base logic file 
6 sizes and 2 kernels in incremental logic file 
[O] [MFMA, 128, 128, 1, 128] already exists and has improved in performance. Efficiency: 10.0 -> 11.0  
[-] [MFMA, 128, 256, 1, 128] has been added to solution table, Efficiency: N/A -> 12.0 
[X] [MFMA, 128, 257, 1, 128] already exists but does not improve in performance. Efficiency: 10.0 -> 8.0 
[-] [VALU, 128, 128, 1, 128] has been added to solution table, Efficiency: N/A -> 14.0 
0 unused kernels removed from base logic file 
0 unused kernels removed from incremental logic file 
4 size(s) and 2 kernel(s) added, 0 kernel(s) removed 
File written to Out/Test.yaml
```
### Limitations
- This only allows for a MFMA and non-MFMA kernel to be assigned the same size if they are present in separate logic files. If both are tuned together, only the faster one will be present. Easiest usage of this would be to tune the MFMA and non-MFMA kernels separately, and then merge together with the new option enabled.
- This will continue to work with repeated merges *provided* the ```--add_mfma_tag``` option is always used while merging. If it is not, the script will either risk comparing the wrong solution's efficiency while merging, or remove the duplicates if trimming is enabled.


Tensile/Utilities/merge.py
```diff
@@ -147,8 +149,44 @@ def debug(*args, **kwargs):
...
+    else:
+        return SolutionTag.VALU
+
+def findSolutionWithIndex(solutionData, solIndex):
+    # Check solution at the index corresponding to solIndex first
+    solution = solutionData[solIndex]
+    if solution["SolutionIndex"] == solIndex:
+        return solution
+    else:
+        print("Searching for index...")
```
rosenrodt, line 176: Suggest using either `verbose()`/`debug()` for verbosity control

Tensile/Utilities/merge.py
```diff
@@ -166,15 +209,13 @@ def mergeLogic(origData, incData, forceMerge, trimSize=True):
 
     origData, numOrigRemoved = removeUnusedKernels(origData, "Base logic file: ")
     incData, numIncRemoved = removeUnusedKernels(incData, "Inc logic file: ")
-
+
```
rosenrodt, line 212: Additional whitespace

rosenrodt (commented):
See the comments inline.

It is recommended you add a few test cases in `Tensile/Tests/unit/test_mergeLogic.py` to make clear what the expected input/output is.

cgmillette (commented):
I agree that a merge test case should be added to show that both MFMA and VALU kernels are merged for the same size. In the case that both are present, we also should try to show the faster one is chosen if ANY alu is chosen.

Do we need a warning to be shown in the event that addMfmaTag is 'False' and mfma kernels are detected? Are we expecting all tuning runs to keep distinct mfma and valu kernels?

Tensile/Utilities/merge.py
```diff
@@ -147,8 +149,44 @@ def debug(*args, **kwargs):
...
+    VALU = 0
+    MFMA = 1
+
+    def __str__(self):
+        return ["VALU", "MFMA"][self]
+    def __repr__(self):
+        return str(self)
+
+def getSolutionTag(solution):
+    if solution["EnableMatrixInstruction"] or solution["MatrixInstruction"]:
```
rosenrodt, line 165: CI reports failure on this line because you are querying a nonexistent key. You should check that if key exists in the first place, or catch `KeyError` exception.

JeremyAdamHart: Odd, that was working on my branch. I'll take a look into this.

JeremyAdamHart: Looks like I made a change after running the tests that broke them, should be good now!

rosenrodt (approved):

PR #1318, opened 2021-03-29
Haampie/develop

TorreZuk:
* add testLib which I expect was missing the compile flags


TorreZuk:
Okay it looks quite comparable now to other extended tests I can see so think the missing bits are covered now.  This issue was when removing the blanket flags you need to ensure all libraries and targets are set as you need both compile and link flags.

zaliu (approved):
Can we please add a Changed line to CHANGELOG.md about the new cmake minimum version requirement?  Thanks!

TorreZuk:
Sure @zaliu  I'll add that if you can run a stand-alone test and see if it meets your expectations as well, the test time variance is quite high but I think this is now within the bounds.  Of course if you spot a stage slower than expected please advise.  Can see how the next extended tests take as well.

zaliu:
I did a quick standalone benchmark run with full validation enabled, the overall time was comparable with a run without this PR.  I did a full rocBLAS build with this PR on Ubuntu, and everything worked fine.

PR #1322, opened 2021-04-01
TensileRetuneLibrary

benjaminulmer:
### Motivation
Through updates to Tensile and rocBLAS, it is possible that the size mappings in rocBLAS become out of date and do not necessarily map each size to the best performing kernel in the library. We have evidence this is happening for F32 on MI100, and it's likely that other data types and hardware have a similar issue. 

This PR adds the TensileRetuneLibrary script, which updates the size mappings in an existing library logic file, thus "retuning" the library.

### Usage
TensileRetuneLibrary takes a single library logic file and a directory to output the results. After executing successfully, the output directory will contain the updated library logic file. This file is identical to the input one except for the size mapping portion, which will have been updated so that each size maps to the _true_ best performing solution.

### Future work and limitations
The current usage of TensileRetuneLibrary is pretty rigid to accommodate a narrow intended use case. A more general type of library based tuning---which allows tuning arbitrary sizes with the solutions present in a library logic file---is planned, but should be added as an expansion of the main Tensile tuning client, not TensileRetuneLibrary.

### DIFF
--- a/Tensile/LibraryIO.py
+++ b/Tensile/LibraryIO.py
@@ -48,18 +48,18 @@ def write(filename_noExt, data, format="yaml"):
     else:
         printExit("Unrecognized format {}".format(format))
 
-def writeYAML(filename, data, **kwags):
+def writeYAML(filename, data, **kwargs):
     """Writes data to file in YAML format."""
     # set default kwags for yaml dump
-    if "explicit_start" not in kwags:
-        kwags["explicit_start"] = True
-    if "explicit_end" not in kwags:
-        kwags["explicit_end"] = True
-    if "default_flow_style" not in kwags:
-        kwags["default_flow_style"] = None
+    if "explicit_start" not in kwargs:
+        kwargs["explicit_start"] = True
+    if "explicit_end" not in kwargs:
+        kwargs["explicit_end"] = True
+    if "default_flow_style" not in kwargs:
+        kwargs["default_flow_style"] = None
 
     with open(filename, "w") as f:
-        yaml.dump(data, f, **kwags)
+        yaml.dump(data, f, **kwargs)
 
 def writeMsgPack(filename, data):
     """Writes data to file in Message Pack format."""
--- a/Tensile/LibraryIO.py
+++ b/Tensile/LibraryIO.py
@@ -19,7 +19,6 @@
 # CTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 ################################################################################
 
-import copy
 from .Common import printExit, printWarning, versionIsCompatible
 from .SolutionStructs import Solution, ProblemSizes, ProblemType
 from . import __version__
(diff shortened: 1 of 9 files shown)


Tensile/LibraryIO.py
```diff
@@ -48,10 +48,18 @@ def write(filename_noExt, data, format="yaml"):
     else:
         printExit("Unrecognized format {}".format(format))
 
-def writeYAML(filename, data):
+def writeYAML(filename, data, **kwags):
```
sdquiring, line 51:
```suggestion
def writeYAML(filename, data, **kwargs):
```

Tensile/TensileRetuneLibrary.py
```diff
@@ -0,0 +1,178 @@
...
+
+    shortName = "benchmark"
+    benchmarkDir = os.path.join(outPath, shortName)
+    sourceDir = os.path.join(benchmarkDir, "source")
+    ensurePath(sourceDir)
+
+    pushWorkingPath(shortName)
+    pushWorkingPath("source")
+
+    filesToCopy = [
```
sdquiring, line 75: I think we can omit the old client stuff from here.

sdquiring (approved):

Tensile/TensileRetuneLibrary.py
benjaminulmer: I agree in theory, but there appears to be a bug with "Kernels.h" and "Kernels.cpp" being generated and then not compiling due to missing includes if you skip these steps. I think I noticed a similar thing a while ago with the normal Tensile client trying to run it with OldClientSourceTmp = False. 

I'll take a look at fixing the underlying issue on Monday. If it ends up being easy to fix, I'll update this. Otherwise, I'll leave this hacky fix for now so we can start retuning and then reevaluate after that's done.

Tensile/TensileRetuneLibrary.py
```diff
@@ -0,0 +1,173 @@
...
+    print1("#  Library Logic: {}".format(libPath))
+    print1("#")
+    print1(HR)
+    print1("")
+
+    # setup global parameters
+    outPath = ensurePath(os.path.abspath(args.output_path))
+    restoreDefaultGlobalParameters()
+    assignGlobalParameters({"LibraryFormat": "msgpack",
+                            "OutputPath": outPath,
```
cgmillette, line 149: I am assuming this narrow case only considers msgpack output? Or would we also need to consider yaml?

benjaminulmer: This just controls the format for the master solution library that will be used by the Tensile client, which is just an intermediary for producing the final result. The overall input and output for this tool is a library logic file, which is always yaml. 

Using the yaml library format should work just fine, but I simply chose to hardcode for msgpack due to the narrow intended use case of this tool.

cgmillette (approved):
LGTM

PR #1324, opened 2021-04-06
Re-enable HardwareMonitor for gfx90a

AlexBrownAMD:


nakajee (approved):
Looks good.
I am OK with this change as long as HardwareMonitor works.

AlexBrownAMD:
Yea, verified locally, can see monitor data in the benchmark results

benjaminulmer (approved):
Same, as long as it all works this lgtm

PR #1331, opened 2021-04-20
Add compiler launcher environment variables

cgmb:
This change adds support for two environment variables: `Tensile_ASM_COMPILER_LAUNCHER` and `Tensile_CXX_COMPILER_LAUNCHER`, which serve a similar purpose as the [`CMAKE_<LANG>_COMPILER_LAUNCHER`](https://cmake.org/cmake/help/latest/envvar/CMAKE_LANG_COMPILER_LAUNCHER.html) variable in CMake.

With this change and `export Tensile_ASM_COMPILER_LAUNCHER=/usr/local/bin/ccache`, the cached assembly kernels are retreived almost instantly when building rocBLAS. The source kernels remain a major cause of slow builds, however.

With `export Tensile_CXX_COMPILER_LAUNCHER=/usr/local/bin/ccache`, Tensile will invoke ccache when building `/src/rocBLAS/build/release/Tensile/Kernels.cpp`, though the result always seems to be a cache miss. That said, I don't think the miss has anything to do with my implementation of this feature. The Kernels.cpp file is 46 MB and preprocessing with -E results in 514 MB of output, so it's a bit difficult to debug. In any case, perhaps another compiler launcher (like sccache) will be able to handle that beast.

### DIFF
--- a/Tensile/KernelWriterAssembly.py
+++ b/Tensile/KernelWriterAssembly.py
@@ -30,6 +30,8 @@ from .AsmUtils import inst, vgpr, sgpr, log2, vectorStaticDivideAndRemainder, ve
 from math import ceil, trunc, modf
 from copy import deepcopy
 import collections
+import os
+import shlex
 import traceback
 from enum import Enum
 
@@ -598,7 +600,9 @@ class KernelWriterAssembly(KernelWriter):
 
     archHasV3 = globalParameters["AsmCaps"][isa]["HasCodeObjectV3"]
 
-    rv = [globalParameters['AssemblerPath'],
+    launcher = shlex.split(os.environ.get('Tensile_ASM_COMPILER_LAUNCHER', ''))
+
+    rv = launcher + [globalParameters['AssemblerPath'],
           '-x', 'assembler',
           '-target', 'amdgcn-amd-amdhsa']
 
--- a/Tensile/TensileCreateLibrary.py
+++ b/Tensile/TensileCreateLibrary.py
@@ -42,6 +42,7 @@ import collections
 import itertools
 import os
 import re
+import shlex
 import shutil
 import subprocess
 import sys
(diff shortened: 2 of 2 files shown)


cgmb:
@JeremyAdamHart answered many of my questions about Tensile while I was making this change, though I will admit my understanding is still quite limited.

AlexBrownAMD (approved):
lgtm

zaliu (changes requested):
Pending @bragadeesh's explicit approval.  Thanks.

cgmb:
This isn't an urgent change. It's a feature for developers using Tensile and it should have no impact on the kernels, so there's little difference in merging this week or next. Feel free to ignore this until Monday.

pfultz2:
This should be a flag that can be passed from cmake(in the `TensileCreateLibraryFiles` function), so the user only needs to set `CMAKE_CXX_COMPILER_LAUNCHER` for it to work with tensile.

cgmb:
My understanding of Tensile is limited, so I tried to define this feature in a way that it would have no impact on anybody unless they specifically enable it. My change purposefully avoids changing behaviour for any existing users, but it may very well be that existing users would benefit from different behaviour. I just opted for the smallest, safest change I could make. (That said, nothing here precludes adding support for `CMAKE_CXX_COMPILER_LAUNCHER` later, either.)

In terms of broader changes, I would love to dig into the Tensile / rocBLAS build and try to clean it up a bit. I think we could cut the rocBLAS compile times on developer workstations and the CI by nearly order of magnitude if we took a step back and reworked the way the Tensile part of the build is defined. However, I'm neither a rocBLAS nor a Tensile developer, so my time to work on that is in short supply.

cgmb:
The source kernels never get a cache hit with ccache due to non-deterministic output from hipcc -E. I've opened SWDEV-283320 to track the issue.

cgmb:
The CI now supports ccache. If you'd like to try out these changes, you can add the `ci:ccache` label to this PR.

PR #1332, opened 2021-04-20
propagate hip error code from Tensile to rocBlas

jichangjichang:
SWDEV-279840 mentioned that rocblas/tensile don't propagate hip return code properly but just swallow the actual error by trigger exception.
This PR add the hip return code into some APIs of SolutionAdapter:: called by rocblas.
rocblas part can be added as well after this PR merged accordingly.

### DIFF
--- a/Tensile/Source/client/main.cpp
+++ b/Tensile/Source/client/main.cpp
@@ -301,12 +301,26 @@ namespace Tensile
             }
             else
             {
+                //only trigger exception when failed to load all code objects.
+                bool       loaded   = false;
+                hipError_t retError = hipSuccess;
+
                 for(auto const& filename : filenames)
                 {
+                    hipError_t ret;
+
                     if(logLevel >= LogLevel::Verbose)
                         std::cout << "Loading " << filename << std::endl;
-                    HIP_CHECK_EXC(adapter.loadCodeObjectFile(filename));
+                    ret = adapter.loadCodeObjectFile(filename);
+
+                    if(ret == hipSuccess)
+                        loaded = true;
+                    else
+                        retError = ret;
                 }
+
+                if(!loaded)
+                    HIP_CHECK_EXC(retError);
             }
         }
 
--- a/Tensile/Source/client/main.cpp
+++ b/Tensile/Source/client/main.cpp
@@ -305,7 +305,7 @@ namespace Tensile
                 {
                     if(logLevel >= LogLevel::Verbose)
                         std::cout << "Loading " << filename << std::endl;
-                    adapter.loadCodeObjectFile(filename);
+                    HIP_CHECK_EXC(adapter.loadCodeObjectFile(filename));
                 }
             }
         }
@@ -587,10 +587,13 @@ int main(int argc, const char* argv[])
                             {
                                 listeners.preWarmup();
                                 if(gpuTimer)
-                                    adapter.launchKernels(
-                                        kernels, stream, warmupStartEvents[i], warmupStopEvents[i]);
+                                    HIP_CHECK_EXC(adapter.launchKernels(kernels,
+                                                                        stream,
+                                                                        warmupStartEvents[i],
+                                                                        warmupStopEvents[i]));
                                 else
-                                    adapter.launchKernels(kernels, stream, nullptr, nullptr);
+                                    HIP_CHECK_EXC(
+                                        adapter.launchKernels(kernels, stream, nullptr, nullptr));
                                 listeners.postWarmup();
                             }
 
@@ -611,10 +614,11 @@ int main(int argc, const char* argv[])
                                 for(int j = 0; j < enq; j++)
                                 {
                                     if(gpuTimer)
-                                        adapter.launchKernels(
-                                            kernels, stream, startEvents[j], stopEvents[j]);
+                                        HIP_CHECK_EXC(adapter.launchKernels(
+                                            kernels, stream, startEvents[j], stopEvents[j]));
                                     else
-                                        adapter.launchKernels(kernels, stream, nullptr, nullptr);
+                                        HIP_CHECK_EXC(adapter.launchKernels(
+                                            kernels, stream, nullptr, nullptr));
                                 }
 
                                 listeners.postEnqueues(startEvents, stopEvents);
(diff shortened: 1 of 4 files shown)


Tensile/Source/client/main.cpp
```diff
@@ -305,7 +305,7 @@ namespace Tensile
                 {
                     if(logLevel >= LogLevel::Verbose)
                         std::cout << "Loading " << filename << std::endl;
-                    adapter.loadCodeObjectFile(filename);
+                    HIP_CHECK_EXC(adapter.loadCodeObjectFile(filename));
```
AlexBrownAMD, line 308: It looks like this is causing problems for some tests because of loading code objects for other architectures. We probably need to either ignore loading invalid code objects, or maybe filter them somehow.

jichangjichang: @AlexBrownAMD 
The precheck_in test of aldebaran failed due to that "xnack+" is enabled but original tensile didn't treat hipErrorSharedObjectInitFailed as a error.
I suggest that we should not ignore the loading error.
I file another commit to temporarily disable building xnack+ for aldebaran and test can pass.

@zaliu @yoichiyoshida 
The xnack+ commit is  https://github.com/ROCmSoftwarePlatform/Tensile/commit/92944f09209fe3677882a91164aa6b671b87c94a
Could you help to check why xnack+ cannot be loaded?
Before the root cause is identified, I suggest to disable it temporarily. https://github.com/ROCmSoftwarePlatform/Tensile/pull/1332/commits/e3718acb821fcd4cb62b23bbc25138c2341704f4 is in this PR.
You can enable AMD_LOG_LEVEL=3 and run pre_checkin test on aldebaran to see the errors.
test commands:

> AMD_LOG_LEVEL=3 tox -v -e py36 -- Tensile/Tests -m pre_checkin

Logs:
> Loading /workspace/Tensile/.tox/py36/client/1_BenchmarkProblems/Cij_Aik_Bjk_SB_00/00_Final/source/library/Kernels.so-000-gfx90a-xnack+.hsaco
> :3:hip_module.cpp           :58  : 519815182272 us: 54343: [7fb6627db180] hipModuleLoad ( 0x7fffa7719a30, /workspace/Tensile/.tox/py36/client/1_BenchmarkProblems/Cij_Aik_Bjk_SB_00/00_Final/source/library/Kernels.so-000-gfx90a-xnack+.hsaco )
> :3:devprogram.cpp           :2394: 519815182540 us: Using Code Object V4.
> :1:devprogram.cpp           :1698: 519815182647 us: Error: AMD HSA Code Object loading failed: HSA_STATUS_ERROR_INCOMPATIBLE_ARGUMENTS: The arguments passed to a functions are not compatible.
> 
> :3:hip_module.cpp           :60  : 519815182700 us: 54343: [7fb6627db180] hipModuleLoad: Returned hipErrorSharedObjectInitFailed :

zaliu: The system can be in either xnack+ or xnack- mode at any moment in time, never both.  As I mentioned in private, our test systems boot in xnack- mode; as such, the xnack+ code object loading at runtime will fail due to it being incompatible with the current xnack mode the system is in.

Tensile/TensileCreateLibrary.py
```diff
@@ -141,9 +141,9 @@ def isSupported(arch):
       for arch in globalParameters['SupportedISA']:
         if isSupported(arch):
           if (arch == (9,0,6) or arch == (9,0,8) or arch == (9,0,10)):
-            if (arch == (9,0,10)):
-              archs += [gfxName(arch) + '-xnack+']
-              cmdlineArchs += [gfxName(arch) + ':xnack+']
+            #if (arch == (9,0,10)):
+            #  archs += [gfxName(arch) + '-xnack+']
+            #  cmdlineArchs += [gfxName(arch) + ':xnack+']
```
zaliu, line 146: This is not what we want to do.  xnack+ support must  be in.  We need to figure out a proper way to not load xnack+ code objects when the system is in xnack- mode.  @t-tye @kzhuravl please advise.  Thanks.

Tensile/Source/client/main.cpp
jichangjichang: Thanks your replies, I will remove the second commit :"disable xnack+" from this PR once the runtime can dynamically check/load correct xnack mode. 
I will set this PR as non-merge first.

Tensile/Source/lib/include/Tensile/hip/HipUtils.hpp
```diff
@@ -69,6 +69,14 @@
         }                                                                                         \
     } while(0)
 
+#define HIP_REPORT_EXC(expr)   \
```
sdquiring, line 72: `EXC` in this context stands for 'exception'.  A better name for this macro would be `HIP_CHECK_RET` or `HIP_CHECK_RETURN` since it does not throw an exception.

sdquiring (commented):
Great to finally fix this oversight.

Tensile/Source/lib/include/Tensile/hip/HipUtils.hpp
jichangjichang: @sdquiring Thanks for your suggestion. Done.

zaliu (approved):
A very reasonable way of handling the loading of code objects.  LGTM.

zaliu:
Note that since hipclang-4.1 replaced compute-rocm-dkms-no-npi-hipclang, all compute-rocm-dkms-no-npi-hipclang related CI failures may be ignored. All hipclang-4.1 CI pipelines passed. OK to merge. Thanks.

jichangjichang:
> A very reasonable way of handling the loading of code objects. LGTM.

@zaliu For Tensile side,

jichangjichang:
> A very reasonable way of handling the loading of code objects. LGTM.

@zaliu 
Tensile client can use this way to handle error. But rocblas side would still encounter hipModuleLoad error if system is gfx90a. The implementation of rocblas for hip error code propagation must be postponed until all hipModuleLoad for tensile library return success in normal case.
@AlexBrownAMD May I know the status for that Tensile only use bundled code objects for source kernel?
Thanks.

PR #1340, opened 2021-04-28
Win10 support

TorreZuk:
* testing minimal modifications that work with rocBLAS
* hipcc.bat and perl used while rocBLAS only directly uses clang


TorreZuk:
This is just a heads up (so draft) as not sure who is taking over windows so put a bunch of you on it.  Unless you will be taking over you can ignore this until it is not Draft.

Tensile/Source/SolutionMapper.h
```diff
@@ -22,8 +22,16 @@
 
 #pragma once
 
+#if defined(max)
```
cgmillette, line 25: If we include <cstdlib> here, is std::getenv() not portable to Windows?

cgmillette:
Seems like we are hitting some CI timeouts - I'll pull your changes and hopefully see where the extra time is going.

Tensile/Source/SolutionMapper.h
TorreZuk: We can retest this again, it was based on Herve's changes.   I think env changes during test execution were not affecting the cloned environment in a reliable way.   If based on our tests it works we can try again just using std::  there are some envp variations on windows mainin which you get some "frozen" environment variables.   And this disclaimer: "getenv operates only on the data structures accessible to the run-time library and not on the environment "segment" created for the process by the operating system"

Tensile/Source/SolutionMapper.h
```diff
@@ -22,8 +22,16 @@
...
+#undef max
+#endif
 #include <cstddef>
 #include <limits>
+
+#ifdef _WIN32
+#include <windows.h>
```
cgmillette, line 32: If we can get away with it, can we try adding the #define WIN32_LEAN_AND_MEAN here ? 

Alternatively, I might tuck this env variable code into Utils.h/.cpp and drag the windows headers into the compilation unit only.

sdquiring (commented):
I started reviewing this some time ago and didn't submit these comments.

Tensile/Common.py
```diff
@@ -1573,7 +1573,7 @@ def tryAssembler(isaVersion, asmString, debug=False, *options):
 
 def gfxArch(name):
     import re
-    match = re.search(r'gfx(\S{3,})', name)
+    match = re.search(r'gfx([0-9a-fA-F]+).*', name)
```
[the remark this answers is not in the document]
TorreZuk: That extra was for hipinfo.exe as it prints different information at the end, e.g. :sramecc+:xnack-  that may change.

Tensile/Common.py
```diff
@@ -1690,25 +1694,40 @@ def assignGlobalParameters( config ):
...
 
   if "ROCmAgentEnumeratorPath" in config:
     globalParameters["ROCmAgentEnumeratorPath"] = config["ROCmAgentEnumeratorPath"]
 
   # read current gfx version
-  if os.name != "nt" and globalParameters["CurrentISA"] == (0,0,0) and globalParameters["ROCmAgentEnumeratorPath"]:
-    command = [globalParameters["ROCmAgentEnumeratorPath"]]#, "-t", "GPU"]
-    result = subprocess.run(command, stdout=subprocess.PIPE)
-    for line in result.stdout.decode().split("\n"):
+  if globalParameters["CurrentISA"] == (0,0,0) and globalParameters["ROCmAgentEnumeratorPath"]:
```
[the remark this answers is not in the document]
TorreZuk: Sure can move it to a function.

cgmillette:
Last integration failures look like to be an issue with that node. Eiden has scheduled that node for a reboot, and in the meantime I ran the job again on a different node. Everything seems to pass now.

TorreZuk:
Any other comments or reviews?  Hoping to get this in this week so windows CI can be pushed to develop branches for rocBLAS.  Work will continue on this of course so further refining and cleanup can occur. @yoichiyoshida the '////;' system failed on windows with different options so switched to '_' separators for multiple arch lists in the create call.    Working in CI build here:  http://mlsejenkinsvm.amd.com/blue/organizations/jenkins/rocBLAS-psdb/detail/rocBLAS-psdb/143/pipeline/5  tensile test run in later build http://mlsejenkinsvm.amd.com/blue/organizations/jenkins/rocBLAS-psdb/detail/rocBLAS-psdb/146/pipeline/37

Tensile/Common.py
```diff
@@ -1752,45 +1795,54 @@ def assignGlobalParameters( config ):
   globalParameters["ROCmPath"] = "/opt/rocm"
   if "ROCM_PATH" in os.environ:
     globalParameters["ROCmPath"] = os.environ.get("ROCM_PATH")
-  if "TENSILE_ROCM_PATH" in os.environ:
```
cgmillette, line 1755: Is anyone still using the TENSILE_ROCM_PATH for anything?

There appear to be other uses of TENSILE_ROCM_*  symbols below.

Tensile/Parallel.py
```diff
@@ -29,7 +29,10 @@ def CPUThreadCount(enable=True):
   if not enable or globalParameters["CpuThreads"] == 0:
     return 0
   else:
-    cpu_count = len(os.sched_getaffinity(0))
+    if os.name == "nt":
+      cpu_count = 1 # os.cpu_count()
```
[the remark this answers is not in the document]
TorreZuk: Yes there are and cpu_count is fine too, the problem is that the python parallelization code used in tensile does not work on windows.  It requires a rewrite.

Tensile/Source/ReferenceCPU.h
```diff
@@ -22,6 +22,9 @@
 
 #ifndef REFERENCE_CPU_H
 #define REFERENCE_CPU_H
+#if defined(max)
+#undef max
+#endif
```
cgmillette, line 27: This is interesting - where is the macro for max coming from ?

Tensile/Common.py
TorreZuk: Good catch.  Sorry about that, it shouldn't have been removed I'll add it back. 
Not sure why you want to potentially mix two ROCMs but that is fine by windows build.

Tensile/Source/SolutionMapper.h
```diff
@@ -194,6 +203,16 @@ class SolutionMapper : public SolutionMapperBase<ProblemDimsType>
...
+#ifdef _WIN32
+        const DWORD nSize = _MAX_PATH;
+        static char db[nSize];
+        GetEnvironmentVariableA("TENSILE_DB", db, nSize);
+        _db = strtol(db,nullptr,0);
+
+        static char alg[nSize];
+        GetEnvironmentVariableA("TENSILE_FIND_ALG", alg, nSize); // If <0 see Algo enumeration, or >=0 specified specific solution index
+        _findAlg = strtol(alg,nullptr,0);
+#else
```
cgmillette, line 215: Can you move this to a cross-platform function in Tools.h / .cpp that returns a const char*, and only pull in the windows header in the .cpp ?

Tensile/Source/lib/include/Tensile/KernelArguments.hpp
```diff
@@ -121,10 +121,9 @@ namespace Tensile
         bool m_log;
     };
 
-    TENSILE_API std::ostream& operator<<(std::ostream& stream, const KernelArguments& t);
     TENSILE_API KernelArguments::const_iterator begin(KernelArguments const&);
```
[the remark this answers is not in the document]
TorreZuk: It is already defined as a friend in the class so this is a duplicate definition.   I would think there would be a compile or link error if it wasn't duplicate no, or does the tensile build not test everything?    It definitely was reported as a duplicate export when I compile.   Note you can start a "review" if you have a lot more left to review to reduce the messaging for others but I don't mind.

yoichiyoshida:
> 
> 
> Any other comments or reviews? Hoping to get this in this week so windows CI can be pushed to develop branches for rocBLAS. Work will continue on this of course so further refining and cleanup can occur. @yoichiyoshida the '////;' system failed on windows with different options so switched to '_' separators for multiple arch lists in the create call. Working in CI build here: http://mlsejenkinsvm.amd.com/blue/organizations/jenkins/rocBLAS-psdb/detail/rocBLAS-psdb/143/pipeline/5 tensile test run in later build http://mlsejenkinsvm.amd.com/blue/organizations/jenkins/rocBLAS-psdb/detail/rocBLAS-psdb/146/pipeline/37

Doesn't that break convention? If other users wishes to call create library they have to use a non-standard separator?

TorreZuk:
> Doesn't that break convention? If other users wishes to call create library they have to use a non-standard separator? 

';' can also bash command separator along with conflict with cmake list specifier.
Did you document the explicit multi-arch argument convention as it wasn't supported before?  
As to confusion with cmake list specifiers the workarounds all seem bad as reports on the web offer no good solution.  Can try to keep it but there will be more ugly cmake code to partition the command string.

Tensile/Source/SolutionMapper.h
TorreZuk: I guess but it will have to be a pointer to thread local data which shouldn't be held on to.   Should be fine for the current uses.

Tensile/Source/ReferenceCPU.h
TorreZuk: I'll delete and see if current windows includes don't complate, was likely an old battle with NOMINMAX.

Tensile/Source/SolutionMapper.h
TorreZuk: Does anyone build tensile on windows ?   All of your windows support code I assumed was out of date.  I didn't really want to get into updating the old code I wasn't going to be testing locally as that is another job for someone else.   If I add to tools.cpp I guess I'll be forced to update it.   @bragadeesh do you know the history of the windows side?  Can move that discussion to an email/channel.

cgmillette (approved):
LGTM

TorreZuk:
Thanks Chris.  
So this is now just using std::getenv until we spot an issue with it.     The windows helper for read_env is only for potential issue down the road in .cpp, nothing in the header file.  To include tools.h in other files that had getenv required changes to python file replication system (as tools.h is only in LegacyComponents now in TensileCreateLibrary.py) so thought it best to leave the usage of std::getenv alone for now.   Wide char formats of paths may be a reason to introduce back later. The tools.h already had windows includes in the header so switched to chrono to push the windows include into only the .cpp and replaced the two win/linux versions of timing with one variant using chrono. 

Deleted some likely outdated #ifdef WIN32 and thread_local should be supported in windows builds.  Again if anyone knows of actual builders of tensiles on windows with old toolchains speak up as this will assume clang/visual studio 2019 for now.  Now using _WIN32 consistently which should be compiler set for all our cases, I still use WIN32 form in blas but can potentially replace now with _WIN32 and drop our define.  I prefer the simpler form but won't force others to follow that convention without the _ character.
 
Wondering if your random testing should be seeded deterministically for the CI builds and only randomly for separate stress testing as it is best if daily/routine CI is running identical tests on all PRs.   This random variation confounds any comparisons to develop and other PRs.

I still have to see if the '_' arch list delimiter I can restore back to ';' for Yoichi's concern.  Maybe if I just replace back right before the command is issued the '_' can just be used only internal to the one cmake file.

Tensile/Source/Tools.cpp
```diff
@@ -19,18 +19,40 @@
...
+#ifdef _WIN32
+    const DWORD nSize = _MAX_PATH;
+    static thread_local char lpBuffer[nSize];
+    DWORD len = GetEnvironmentVariableA(env_var, lpBuffer, nSize);
+    if (len && len < nSize)
+        return lpBuffer;
+    else 
+        return nullptr; // variable not found or longer than nSize
+#else
+    return std::getenv(env_var);
```
AlexBrownAMD, line 48: Doesn't std::getenv work on windows?

AlexBrownAMD (approved):

Tensile/Source/Tools.cpp
TorreZuk: Alex did you read the earlier discussion?   It works but there may be issues with replication and modification during tests, hopefully none are applicable to tensile.    If you have wide char paths in the environment variable that would also pose issues as there is a separate API for reading those.

AlexBrownAMD: Oops, just saw the discussion, that makes sense

cgmillette:
> Thanks Chris.
> So this is now just using std::getenv until we spot an issue with it. The windows helper for read_env is only for potential issue down the road in .cpp, nothing in the header file. To include tools.h in other files that had getenv required changes to python file replication system (as tools.h is only in LegacyComponents now in TensileCreateLibrary.py) so thought it best to leave the usage of std::getenv alone for now. Wide char formats of paths may be a reason to introduce back later. The tools.h already had windows includes in the header so switched to chrono to push the windows include into only the .cpp and replaced the two win/linux versions of timing with one variant using chrono.
> 
> Deleted some likely outdated #ifdef WIN32 and thread_local should be supported in windows builds. Again if anyone knows of actual builders of tensiles on windows with old toolchains speak up as this will assume clang/visual studio 2019 for now. Now using _WIN32 consistently which should be compiler set for all our cases, I still use WIN32 form in blas but can potentially replace now with _WIN32 and drop our define. I prefer the simpler form but won't force others to follow that convention without the _ character.
> 
> Wondering if your random testing should be seeded deterministically for the CI builds and only randomly for separate stress testing as it is best if daily/routine CI is running identical tests on all PRs. This random variation confounds any comparisons to develop and other PRs.
> 
> I still have to see if the '_' arch list delimiter I can restore back to ';' for Yoichi's concern. Maybe if I just replace back right before the command is issued the '_' can just be used only internal to the one cmake file.

Thanks Torre for the updates. The seed we use for the 'random GEMMs' is a static variable so will be same per launch of test executable. That way will be more deterministic. 

We don't have many (if any) Windows devs working on Tensile atm to my knowledge. I assume this will change very soon. Seems clang / visual studio 2019 is reasonable, but we can always adjust later if there are other needs.

TorreZuk:
@yoichiyoshida put back the semi-colon list for you and now keep it as a separate list so it doesn't need the escape magic.

sdquiring (approved):

PR #1342, opened 2021-04-29
Report GFlops with 6 digits precision

benjaminulmer:
These changes also make all other double values report with 6 digits of precision; however, I don't think this will cause any issues.

### DIFF
--- a/Tensile/Source/client/include/LibraryUpdateReporter.hpp
+++ b/Tensile/Source/client/include/LibraryUpdateReporter.hpp
@@ -2,7 +2,7 @@
  *
  * MIT License
  *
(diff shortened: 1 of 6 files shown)


zaliu:
Note that since hipclang-4.1 replaced compute-rocm-dkms-no-npi-hipclang, all compute-rocm-dkms-no-npi-hipclang related CI failures may be ignored. All hipclang-4.1 CI pipelines passed at least once.  Reviewers please consider this PR passed Tensile CI fully.  Thanks.

zaliu (approved):

cgmillette (approved):
LGTM, as long as CI is happy

PR #1343, opened 2021-05-03
TensileRetuneLibrary updates

benjaminulmer:
In addition to a bit of cleaning up, this PR adds two new features to the retuning client:
1) Option to fully remake the resulting library logic file instead of---or in addition to---using the previous update method
2) Option to skip replacement kernels and sizes that map to replacement kernels. This option only works when fully remaking the output logic

### DIFF
--- a/Tensile/TensileRetuneLibrary.py
+++ b/Tensile/TensileRetuneLibrary.py
@@ -50,8 +50,17 @@ def parseCurrentLibrary(libPath, skipRK):
 
     # process exactLogic into ProblemSizes
     sizes = []
-    for (size, _) in exactLogic:
+    for (size, mapping) in exactLogic:
+        if skipRK:
+            sol = solutions[mapping[0]]
+            if sol["ReplacementKernel"]:
+                continue
+
         sizes.append({"Exact": size})
+
+    if skipRK:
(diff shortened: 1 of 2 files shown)


Tensile/TensileRetuneLibrary.py
```diff
@@ -115,11 +133,21 @@ def runBenchmarking(solutions, problemSizes, outPath):
     forBenchmark = True
     # TODO make this work with TileAware selection
     returncode = ClientWriter.runClient(libraryLogicPath, forBenchmark, False)
-
     if returncode:
-        printExit("BenchmarkProblems: Benchmark Process exited with code %u" % returncode)
+        printWarning("BenchmarkProblems: Benchmark Process exited with code {}".format(returncode))
```
wbgilmartin, line 137: Language? With this change you are not really exiting here.

benjaminulmer: I've updated this statement to more accurately describe the state of the program

zaliu: @benjaminulmer Please don't forget to re-request a review by @wbgilmartin, thanks.

sdquiring (approved):

PR #1344, opened 2021-05-03
Prevent repeated GitHub API calls in integration tests

eidenyoshida:


zaliu (approved):

rosenrodt:
@eidenyoshida does the downloaded file persist in disk for a long time? If so, the zipball link saved in the disk may not keep up with latest rocm release. Unless CI nodes regularly clear the working cache.

rosenrodt:
@eidenyoshida got it. thanks for the clarification!

PR #1345, opened 2021-05-04
Speed up dgemm using atomic add when c=d and beta=1

AlexBrownAMD:


yoichiyoshida (approved):

TonyYHsieh:
Hi @AlexBrownAMD,

All tensile kernels need to support C buffer != D buffer.
We can't expect rocblas to pass C buffer = D buffer.
CEqualsD is just how Tensile train client to allocate C, D buffer.
CEqualsD can't limit how user to pass C/D buffer to kernel.

TonyYHsieh (changes requested):
We can't limit rocblas/user to pass C=D buffer.

yoichiyoshida:
> 
> 
> We can't limit rocblas/user to pass C=D buffer.

No, we already re-added support for allowing rocblas to specify C=D for performance uplift. Please see PR #1302 and rocBLAS PR 665.

imcarsonliao:
We have some concern regarding to this solution which is specified to beta=1 and C=D. 
case 1:  When the same size of M, N, K but beta=0 is tuning,  beta=1 kernel will be replaced by beta=0 kernel in candidate pool.
case2 : When the same size of M, N, K but not satisfies "beta=1 && C=D" is calling, the will be no candidate kernel to be able to choose. 

FYI - Tony has another PR coming up that could global write 2 coalesced elements per inst and which performs equally to have atomic add. The advantage of it is more general without specifying C=D and beta.  Just for your information.

yoichiyoshida:
Both cases are already covered, the merge script in PR #1302 already has a flag that allows for the beta=0 kernel to be appended and not replace the previous tuned kernel for the general case. Case 2 is not a thing as if a predicate is not satisfied it should pick the next closest size, as per the regular search behavior.

Also, we have custom replacement kernels that outperform this atomic kernel with the same restrictions, that are being tested for specific use cases.

PR #1352, opened 2021-05-18
Support multiple accumulated buffer GSU in all type

TonyYHsieh:
Support multiple accumulated buffer GSU in all type
- default GSU algorithm will change to multi-buffer
- previous trained logic yaml will remain in single buffer.

### DIFF
--- a/Tensile/Common.py
+++ b/Tensile/Common.py
@@ -1189,7 +1189,7 @@ defaultBenchmarkCommonParameters = [
     {"StaggerUMapping":           [ 0 ] },    # recommend [0,1]
     {"MagicDivAlg":               [ 2 ] },
     {"GlobalSplitU":              [ 1 ] },
-    {"GlobalSplitUAlgorithm":     [ "MultipleBuffer" ] },
+    {"GlobalSplitUAlgorithm":     [ "SingleBuffer" ] },
     {"GlobalSplitUSummationAssignmentRoundRobin": [ True ] },
     {"GlobalSplitUWorkGroupMappingRoundRobin":    [ False ] },
     {"MacroTileShapeMin":         [ 1 ] },
--- a/Tensile/KernelWriterAssembly.py
+++ b/Tensile/KernelWriterAssembly.py
@@ -8408,53 +8408,14 @@ class KernelWriterAssembly(KernelWriter):
     if component:
       return component(self, kernel, tP)
 
+
   ##############################################################################
   # Complex Declare Tmp Registers - SKIP
   ##############################################################################
   def complexDeclareTmpRegisters(self, kernel):
     kStr = ""
     return kStr
-    #if kernel["ProblemType"]["DataType"].value == DataType.complexSingle:
-    #  kStr += "  float type_mac_tmp" + self.endLine
-    #if kernel["ProblemType"]["DataType"].value == DataType.complexDouble:
-    #  kStr += "  double type_mac_tmp" + self.endLine
-    #return kStr
 
-  ##############################################################################
-  # isLds = true if querying about LDS operations (which can use dword operations)
-  #     isLds=False if we want element step for the VALU add operations
-  ##############################################################################
-  def getLocalSplitUElementStep(self, kernel, isLds):
-
-    if isLds and \
-       kernel["VectorWidth"]*self.bpeCinternal >= 8 and \
-       kernel["GlobalWriteVectorWidth"]*self.bpeCinternal >= 8:
-      useDwordX2 = 1
-    else:
-      useDwordX2 = 0
-
-    #useDwordX2 = 0
-
-    if kernel["ProblemType"]["DataType"].isHalf() and not kernel["ProblemType"]["HighPrecisionAccumulate"]:
-      assert(kernel["VectorWidth"]%2 == 0)
-      elementStep = 2*(useDwordX2+1)
-    # TODO: LocalSplitU - INT8
-    elif kernel["ProblemType"]["DataType"].isInt8x4() or \
-         kernel["ProblemType"]["DataType"].isBFloat16() or \
-         kernel["ProblemType"]["DataType"].isHalf() or \
-         kernel["ProblemType"]["DataType"].isSingle():
-      elementStep = 1*(useDwordX2+1)
-    elif kernel["ProblemType"]["DataType"].isDouble() or \
-         kernel["ProblemType"]["DataType"].isSingleComplex():
-      if isLds:
-        assert (useDwordX2==1)
-      elementStep = 1
-    elif kernel["ProblemType"]["DataType"].isDoubleComplex():
-      if isLds:
-        assert (useDwordX2==1)
-      elementStep = 1
-
-    return (elementStep, useDwordX2)
 
   ##############################################################################
   # LocalSplitU: Local Write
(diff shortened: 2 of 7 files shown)


Tensile/SolutionStructs.py
```diff
@@ -2401,22 +2401,26 @@ def assignDerivedParameters(state):
...
+    if ("_GlobalAccumulation" not in state) or ("_WorkspaceSizePerElemC" not in state):
+      state["_GlobalAccumulation"] = None
       state["_WorkspaceSizePerElemC"] = 0
+      if state["GlobalSplitU"] > 1:
+        computeName  = state["ProblemType"]["ComputeDataType"].toName()
+        computeBytes = state["ProblemType"]["ComputeDataType"].numBytes()
+        if state["ProblemType"]["DataType"].isHalf() and state["ProblemType"]["HighPrecisionAccumulate"]:
+          computeName  = 'single'
+          computeBytes = 4
+
```
imcarsonliao, line 2413: No need to consider bf16 and in8 here and everywhere else?

TonyYHsieh: they have correct compute type
Only half is wrong on compute type.

We should correct compute type of hh to s, but all code will get influence.
So we don't correct.

imcarsonliao (approved):

Tensile/Common.py
```diff
@@ -1189,7 +1189,7 @@ def getArchitectureName(gfxName):
     {"StaggerUMapping":           [ 0 ] },    # recommend [0,1]
     {"MagicDivAlg":               [ 2 ] },
     {"GlobalSplitU":              [ 1 ] },
-    {"GlobalSplitUAlgorithm":     [ "MultipleBuffer" ] },
+    {"GlobalSplitUAlgorithm":     [ "SingleBuffer" ] },
     {"GlobalSplitUSummationAssignmentRoundRobin": [ True ] },
```
cgmillette, line 1193: How is this 

> default GSU algorithm will change to multi-buffer

?

Tensile/KernelWriterAssembly.py
```diff
@@ -8580,85 +8543,86 @@ def localSplitULocalWrite(self, kernel):
...
-                vgpr("ValuC+%u" % regIdx), vgpr("ValuC+%u"%cIdx), "c[%u] += c[%u]"%(cIdx, regIdx) )
+    for r in range(1, kernel["LocalSplitU"]):
+      for i in range(0, kernel["NumGlobalWriteVectorsPerThread"]):
+        for s in range(0, kernel["GlobalWriteVectorWidth"], elementStep):
+          cIdx = int((s + i * kernel["GlobalWriteVectorWidth"]) * regsPerElem)
+          regIdx = int((s + i * kernel["GlobalWriteVectorWidth"] + r * kernel["GlobalWriteVectorWidth"] * kernel["NumGlobalWriteVectorsPerThread"]) * regsPerElem)
 
-          # TODO: LocalSplitU - INT8
+          if kernel["ProblemType"]["DataType"].isHalf() and (not kernel["ProblemType"]["HighPrecisionAccumulate"]):
+            kStr += inst("v_pk_add_f16", vgpr("ValuC+%u"%cIdx), vgpr("ValuC+%u" % regIdx), vgpr("ValuC+%u"%cIdx), \
```
cgmillette, line 8602: Can use is_pack16 for test

Tensile/KernelWriterBetaOnly.py
```diff
@@ -58,7 +58,10 @@ def functionSignature(self):
 
     # pointers
     ptrStr = self.state["ProblemType"]["DestDataType"].toDevice(self.language)
-    ptrStr = "float" if self.state["_GlobalAccumulation"] else ptrStr
+    if self.state["_GlobalAccumulation"]:
+      ptrStr = self.state["ProblemType"]["ComputeDataType"].toDevice(self.language)
+      if self.state["ProblemType"]["DataType"].isHalf() and self.state["ProblemType"]["HighPrecisionAccumulate"]:
+        ptrStr = "float"
```
cgmillette, line 64: Does this include bf16?

cgmillette (commented):
Some checks I am unsure why the half type is only checked with HPA. Should we not also check bf16 types as well? Or does this not apply to bf16?

Also default GSU algorithm is switched to SingleBuffer, which contradicts the description.

Tensile/Common.py
TonyYHsieh: SingleBuffer means putting each GSU output to same one buffer, so we have to use atomic.
MultipleBuffer means putting each GSU output to different buffer, so we can use buffer_store.
In MultipleBuffer, we use another Kernel to summation result from different buffer.
This parameter only take effect on HPA BF16/FP16 type before.
We enhance to let each type to use this multipleBuffer.
Default behavior of other type is using atomic, we change this parameter back to single buffer.

Tensile/KernelWriterBetaOnly.py
TonyYHsieh: Definition of HPA FP16 is wrong.
Compute type of HPA FP16 should be float.
That is why we only modify it back to float.
Compute type of BF16 is float.

TonyYHsieh:
> tc

We have two type of GSU algorithm.
Multiple buffer only take effect on HPA FP16/BF16 before.
We extend to let all type to use multi-buffer algorithm.
But default of other type behavior is single, so we change parameter back to single buffer.

Tensile/KernelWriterAssembly.py
TonyYHsieh: done

Tensile/KernelWriterSource.py
```diff
@@ -833,8 +839,11 @@ def functionSignature(self, kernel ):
     if self.language == "HIP":
       #s += "  hipLaunchParm lp," + self.endLine
       globalStr = ""
-    ptrStr = kernel["ProblemType"]["DestDataType"].toDevice(self.language) \
-        if not kernel["_GlobalAccumulation"] else "float"
+    ptrStr = kernel["ProblemType"]["DestDataType"].toDevice(self.language)
+    if kernel["_GlobalAccumulation"]:
+      ptrStr = kernel["ProblemType"]["ComputeDataType"].toDevice(self.language)
+      if kernel["ProblemType"]["DataType"].isHalf() and kernel["ProblemType"]["HighPrecisionAccumulate"]:
+        ptrStr = "float"
```
cgmillette, line 846: having this string here seems a little fragile - is there a way to instantiate a float datatype and then use the toDevice(self.language) on it?

Tensile/SolutionStructs.py
```diff
@@ -2401,22 +2401,26 @@ def assignDerivedParameters(state):
...
-    else:
+    if ("_GlobalAccumulation" not in state) or ("_WorkspaceSizePerElemC" not in state):
+      state["_GlobalAccumulation"] = None
       state["_WorkspaceSizePerElemC"] = 0
+      if state["GlobalSplitU"] > 1:
+        computeName  = state["ProblemType"]["ComputeDataType"].toName()
+        computeBytes = state["ProblemType"]["ComputeDataType"].numBytes()
+        if state["ProblemType"]["DataType"].isHalf() and state["ProblemType"]["HighPrecisionAccumulate"]:
+          computeName  = 'single'
+          computeBytes = 4
```
cgmillette, line 2412: Here the string is now 'single' whereas above it was 'float'. Again instead of manually assigning strings and byte counts, can we instantiate a proper datatype object and retrieve the toName and numBytes values?

cgmillette (commented):
Looks good to me, minus the coupling to explicit strings. I would prefer to see name strings and byte counts come from DataType objects instead of manually pasted.

zaliu:
> Looks good to me, minus the coupling to explicit strings. I would prefer to see name strings and byte counts come from DataType objects instead of manually pasted.

@TonyYHsieh can you please address this comment?  Thans.

Tensile/SolutionStructs.py
TonyYHsieh: modify to use DataType.

Tensile/KernelWriterSource.py
TonyYHsieh: modify to use DataType.

TonyYHsieh:
> > Looks good to me, minus the coupling to explicit strings. I would prefer to see name strings and byte counts come from DataType objects instead of manually pasted.
> 
> @TonyYHsieh can you please address this comment? Thanks.

updated.

cgmillette (approved):
LGTM

PR #1354, opened 2021-05-25
fix large K correctness issue

TonyYHsieh:
1. non-power-of-2 GSU
  non-power-2 of GSU need divide operation.
  LargeK will make salu divide overflow
  use valu to handle non-power-2 of GSU case.

2. set buffer limit to ffffffff


zaliu:
@TonyYHsieh Can we please add a simple test to e.g. precheckin to cover this issue?  Thanks.

TonyYHsieh:
> @TonyYHsieh Can we please add a simple test to e.g. precheckin to cover this issue? Thanks.

Add test case in extend.

yoichiyoshida (approved):

PR #1359, opened 2021-06-01
don't allocate vgpr in globalWriteBatch

TonyYHsieh:
globalWriteBatch will allocate most of vgpr in order to do batch write.
So allocate vgpr in globalWriteBatch is easily to met vpgr overflow.
We should allocate all needed vgpr outside of globalWriteBatch.

### DIFF
--- a/Tensile/KernelWriterAssembly.py
+++ b/Tensile/KernelWriterAssembly.py
@@ -10274,6 +10274,9 @@ class KernelWriterAssembly(KernelWriter):
       numTmpVgpr = 2 + 3 # GLOBAL_OFFSET_C needs 3, plus 2 tmps?
     tmpVgpr = self.vgprPool.checkOutAligned(numTmpVgpr, 2, "store tmps")
 
+    isHpaBF16 = kernel["ProblemType"]["DestDataType"].isBFloat16() and kernel["ProblemType"]["HighPrecisionAccumulate"]
+    bf16CVTVgpr = self.vgprPool.checkOut(4) if isHpaBF16 else None
+
     ########################################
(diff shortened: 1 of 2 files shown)


rosenrodt:
Can we have a check to disallow checkouts of vgprs during store, in case someone in the future doesn't know what kind of trap he'll walk into?

TonyYHsieh:
> Can we have a check to disallow checkouts of vgprs during store, in case someone in the future doesn't know what kind of trap he'll walk i

TonyYHsieh:
> Can we have a check to disallow checkouts of vgprs during store, in case someone in the future doesn't know what kind of trap he'll walk into?

not in this PR.

Tensile/KernelWriterAssembly.py
```diff
@@ -10623,6 +10626,8 @@ def globalWriteElements(self, kernel, vectorWidths, elements,
     # End label
     kStr += "label_%s:%s"%(endLabel, self.endLine)
     self.vgprPool.checkIn(tmpVgpr)
+    if bf16CVTVgpr != None:
```
cgmillette, line 10629: Should probably use 
if bf16CVTVgpr is not None

cgmillette (approved):
I suggest using "is not None" vs "!= None".

Otherwise, LGTM, as long as CI is happy.

Tensile/KernelWriterAssembly.py
TonyYHsieh: update

zaliu:
@TonyYHsieh please resolve the source conflict.  Thanks.

PR #1360, opened 2021-06-01
let mfma half and i8 use vectorwidth 1

TonyYHsieh:
let fp16/BF16/i* run with vectorwidth 1, so we can use 32x32 MT for MFMA SourceSwap.
MFMA non-SourceSwap doesn's use vectorwidth.
MFMA SourceSwap use vectorwidth only in M dimension.


aazz44ss (approved):
LGTM

cgmillette (approved):
LGTM, as long as CI is happy

PR #1365, opened 2021-06-07
Fix multiple Github api calls for integration tests

benjaminulmer:
Previously, each worker in the integration tests would make an api.github.com call which would occasionally exceed the api rate limit. This PR adds checks to ensure the api call is only made once per test session. This should help mitigate the issue of ever exceeding the rate limit. 

Tested locally both with and without multiple test workers (`-n 4` and `-n 0` when calling pytest).

### DIFF
--- a/Tensile/Tests/integration/test_integration.py
+++ b/Tensile/Tests/integration/test_integration.py
@@ -1,8 +1,7 @@
 import os, subprocess, shlex, shutil, random, pytest
+from filelock import FileLock
 from Tensile import ClientWriter, LibraryIO, Common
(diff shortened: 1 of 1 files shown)


rosenrodt:
Awesome. The current free-tier Github API rate limit is 60 queries/hr. Now we can practically stop worrying about exceeding that limit.

yoichiyoshida (approved):

eidenyoshida (approved):

zaliu (approved):
Thanks for the fix.  Please merge ASAP.

PR #1371, opened 2021-06-23
Refactor vgpr occupancy calculation for Aldebaran's new hardware capability Unified Register Files 

jichangjichang:
For UnifiedRegisterFiles, arch vgpr and acc vgpr share same pool of 512 registers.
This PR refactor the getOccupany calculation of KernelWriter and extend getOccupancy CI tests.

### DIFF
--- a/Tensile/KernelWriterAssembly.py
+++ b/Tensile/KernelWriterAssembly.py
@@ -400,31 +400,6 @@ class ZeroPadReg:
   def isMatch(self, perp, sPerp, para, sPara):
     return self.perp==perp and self.sPerp==sPerp and self.para==para and self.sPara==sPara
 
-class VgprOccupancyCurve:
-  def __init__(self):
-    self._vgprOccupancy = [0]*(256+1)
-    for i in range(0,   24+1): self._vgprOccupancy[i] = 10
-    for i in range(25,  28+1): self._vgprOccupancy[i] = 9
-    for i in range(29,  32+1): self._vgprOccupancy[i] = 8
-    for i in range(33,  36+1): self._vgprOccupancy[i] = 7
-    for i in range(37,  40+1): self._vgprOccupancy[i] = 6
-    for i in range(41,  48+1): self._vgprOccupancy[i] = 5
-    for i in range(49,  64+1): self._vgprOccupancy[i] = 4
-    for i in range(65,  84+1): self._vgprOccupancy[i] = 3
-    for i in range(85, 128+1): self._vgprOccupancy[i] = 2
-    for i in range(129,256+1): self._vgprOccupancy[i] = 1
-  def __call__(self, index):
-    if index < 0 or index > 256:
-      return 0
-    return self._vgprOccupancy[index]
-  def __getitem__(self, item):
-    return self(item)
-  def __len__(self):
-    return len(self._vgprOccupancy)
-
-vgprOccupancy = VgprOccupancyCurve()
-
-
 class PreLoopVmcntCase(Enum):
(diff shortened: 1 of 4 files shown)


Tensile/KernelWriterAssembly.py
```diff
@@ -861,6 +841,34 @@ def selectMemoryInstruction(self,
     else:
       raise RuntimeError("Could not find valid memory instruction for operation=%s, width=%f, kernel=%s" %(operation, width, self.kernelName))
 
+  class VgprOccupancyCurve:
```
rosenrodt, line 844: Any particular motivation to move the class into `KernelWriterAssembly`?

rosenrodt (changes requested):
Please check the failing tests in Tensile/Tests/unit/test_KernelWriterAssembly.py

Some of the relevant functions are made either a static method of `KernelWriterAssembly` or independent from `KernelWriterAssembly` is because of ease of test. So you may want to revisit the your changes that moved those function into `KernelWriterAssembly`.

jichangjichang (commented):
> Please check the failing tests in Tensile/Tests/unit/test_KernelWriterAssembly.py
> 
> Some of the relevant functions are made either a static method of `KernelWriterAssembly` or independent from `KernelWriterAssembly` is because of ease of test. So you may want to revisit the your changes that moved those function into `KernelWriterAssembly`.

This test should be removed.
Occupancy table is different between different GCN version.
Test should not treat it as a static table.

rosenrodt:
> This test should be removed.
> Occupancy table is different between different GCN version.
> Test should not treat it as a static table.

The updated occupancy calculator still needs to be proven correct one way or the other. Removing the tests won't help with that goal.

You can use `pytest.mark.parametrize` decorator for arch-dependent tests. There is also a predefined pytest fixture `useGlobalParameters` in conftest.py to allow you to access `globalParameters` from within the test. That should be enough

jichangjichang:
> > This test should be removed.
> > Occupancy table is different between different GCN version.
> > Test should not treat it as a static table.
> 
> The updated occupancy calculator still needs to be proven correct one way or the other. Removing the tests won't help with that goal.
> 
> You can use `pytest.mark.parametrize` decorator for arch-dependent tests. There is also a predefined pytest fixture `useGlobalParameters` in conftest.py to allow you to access `globalParameters` from within the test. That should be enough

Designed has been changed. Occupancy is calculated when runtime but not refer to a table.
The CI test scope also extend for UnifiedRegisterFiles.

jichangjichang:
> > This test should be removed.
> > Occupancy table is different between different GCN version.
> > Test should not treat it as a static table.
> 
> The updated occupancy calculator still needs to be proven correct one way or the other. Removing the tests won't help with that goal.
> 
> You can use `pytest.mark.parametrize` decorator for arch-dependent tests. There is also a predefined pytest fixture `useGlobalParameters` in conftest.py to allow you to access `globalParameters` from within the test. That should be enough

jichangjichang:
> Please check the failing tests in Tensile/Tests/unit/test_KernelWriterAssembly.py
> 
> Some of the relevant functions are made either a static method of `KernelWriterAssembly` or independent from `KernelWriterAssembly` is because of ease of test. So you may want to revisit the your changes that moved those function into `KernelWriterAssembly`.

@rosenrodt need your review to unblock the change requested

rosenrodt (approved):
Nice. Looks good

jichangjichang:
rocblas-test pass on gfx908 and gfx90a

zaliu (approved):

PR #1372, opened 2021-06-23
Bug fix: codeMulAlpha object is not None when MIArchVgpr is disabled

jichangjichang:


aazz44ss (approved):
LGTM

zaliu (approved):
@aazz44ss & @jichangjichang Thanks for the quick fixes.

PR #1377, opened 2021-07-08
Add PrefetchAcrossPersistentMode 

AlexBrownAMD:
Add PrefetchAcrossPersistentMode to disable static tile setup code and merge prefetch with ord. NoLoadLoop for sizes where K is a multiple of DepthU

### DIFF
--- a/Tensile/KernelWriter.py
+++ b/Tensile/KernelWriter.py
@@ -1307,7 +1307,9 @@ class KernelWriter(metaclass=abc.ABCMeta):
       #    but since there are some vgpr value is used in the later lwaFirstOffset (when not OptNLL, such as "lwoT")
       #    so we still do this part when "isPap & not OptNLL"
       # 2. if tile edge, then we still need to add all these codes even isPap
-      self.dontAppendCode = isPap and isOptNLL and (not needShift)
+
+      # self.dontAppendCode = isPap and isOptNLL and (not needShift)
+      self.dontAppendCode = isPap and kernel["PrefetchAcrossPersistentMode"] == 1 and (not needShift)
       # tile assignments
       kl.append(self.comment("global read addresses: tile offset assignment a"))
       kl.append(self.graTileAssignment(kernel, tensorParametersA))
@@ -1405,7 +1407,9 @@ class KernelWriter(metaclass=abc.ABCMeta):
       kl.append(self.lwaUnrollAssignment(kernel, tensorParametersB))
 
       # if PAP, no need to reset LWA, but if not OptNLL, we still do this (due to TailLoop)
-      self.dontAppendCode = isPap and isOptNLL
+
+      # self.dontAppendCode = isPap and isOptNLL
+      self.dontAppendCode = isPap and kernel["PrefetchAcrossPersistentMode"] == 1
       # first offsets
       kl.append(self.comment("local write addresses: first offset a"))
       kl.append(self.lwaFirstOffset(kernel, tensorParametersA))
@@ -1542,6 +1546,15 @@ class KernelWriter(metaclass=abc.ABCMeta):
 
     # the scheduled GlobalRead,Inc code of PAP is inside openSumAtLeastUnroll (if PAP=on)
     kl.append(self.openSumAtLeastUnroll(kernel, prefetch=False, isOptNLL=isOptNLL, isPap=isPap))
+
+    if self.prefetchAcrossPersistent and kernel["PrefetchAcrossPersistentMode"] == 1 and isPap and not isOptNLL:
+      kStr = ""
+      kStr += str(self.openPrefetchAcrossPersistent(kernel, isOptNLL=False, useBufferOOB=True))
+      newTileCodes = self.setupNewTile(kernel, self.tPA, self.tPB, isPap=True, isOptNLL=True)
+      codes = '\n'.join([str(x) for x in newTileCodes])
+      kStr += codes
+      kStr += str(self.closePrefetchAcrossPersistent(kernel, isOptNLL=False, useBufferOOB=True))
+      kl.append(kStr)
 
     if not self.numItersPLR:
       if self.enable["Wait"]:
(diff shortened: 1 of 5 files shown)


Tensile/KernelWriterAssembly.py
```diff
@@ -11748,17 +11748,32 @@ def openPrefetchAcrossPersistent(self, kernel, isOptNLL):
     if kernel["PersistentKernelAlongBatch"]:
       imod.addInst("s_mul_i32", sgpr(stmp), sgpr(stmp), sgpr("NumWorkGroups2"), "Total WG-0 x 1 x 2")
     imod.addInst("s_cmp_ge_u32", sgpr("SerialWorkGroupIter"), sgpr(stmp), "outside legal WG?")
-    imod.addInst("s_cbranch_scc1", self.getNamedLabel(label), "skip pf if OOB - last tile no PAP, go to pure OptNLL")
+    if useBufferOOB and kernel["PrefetchAcrossPersistentMode"] == 1:
```
nakajee, line 11751: Do we need to check openPrefetchAcrossPersistent here?
It is already checked when setting useBufferOOB = True.

Tensile/KernelWriter.py
```diff
@@ -1307,7 +1307,9 @@ def setupNewTile(self, kernel, tensorParametersA, tensorParametersB, isPap, isOp
       #    but since there are some vgpr value is used in the later lwaFirstOffset (when not OptNLL, such as "lwoT")
       #    so we still do this part when "isPap & not OptNLL"
       # 2. if tile edge, then we still need to add all these codes even isPap
-      self.dontAppendCode = isPap and isOptNLL and (not needShift)
+
+      # self.dontAppendCode = isPap and isOptNLL and (not needShift)
```
nakajee, line 1311: Do we need to keep original code here?
Don't we need to update above comment regarding OptNLL since OptNLL is no longer used here?

Tensile/KernelWriter.py
```diff
@@ -1543,6 +1547,15 @@ def noLoadLoop( self, kernel, tensorParametersA, tensorParametersB, isOptNLL, is
     # the scheduled GlobalRead,Inc code of PAP is inside openSumAtLeastUnroll (if PAP=on)
     kl.append(self.openSumAtLeastUnroll(kernel, prefetch=False, isOptNLL=isOptNLL, isPap=isPap))
 
+    if self.prefetchAcrossPersistent and kernel["PrefetchAcrossPersistentMode"] == 1 and isPap and not isOptNLL:
+      kStr = ""
+      kStr += str(self.openPrefetchAcrossPersistent(kernel, isOptNLL=False, useBufferOOB=True))
+      newTileCodes = self.setupNewTile(kernel, self.tPA, self.tPB, isPap=True, isOptNLL=True)
```
nakajee, line 1553: Is isOptNLL=True correct here?
If it is intended, it might be good to have some comment here.

Tensile/KernelWriterAssembly.py
```diff
@@ -11748,17 +11748,32 @@ def openPrefetchAcrossPersistent(self, kernel, isOptNLL):
...
 
   ##############################################################################
   ##############################################################################
-  def closePrefetchAcrossPersistent(self, kernel, isOptNLL):
+  def closePrefetchAcrossPersistent(self, kernel, isOptNLL, useBufferOOB=False):
     label = "SkipPrefetchAcrossPersistent_OptNLL" if isOptNLL else "SkipPrefetchAcrossPersistent"
     imod = Code.Module()
     # imod.addCode(Code.WaitCnt(self.version, 0,0, "bozo, conservative wait"))
-    imod.addCode("%s: //%s"%(self.getNamedLabel(label), "SkipPrefetchAcrossPersistent"))
+    addLabel = kernel["PrefetchAcrossPersistentMode"] != 1 or not useBufferOOB
```
nakajee, line 11774: Same here. Do we need to check PrefetchAcrossPersistentMode != 1 here?

Tensile/KernelWriter.py
```diff
@@ -1543,6 +1547,15 @@ def noLoadLoop( self, kernel, tensorParametersA, tensorParametersB, isOptNLL, is
     # the scheduled GlobalRead,Inc code of PAP is inside openSumAtLeastUnroll (if PAP=on)
     kl.append(self.openSumAtLeastUnroll(kernel, prefetch=False, isOptNLL=isOptNLL, isPap=isPap))
 
+    if self.prefetchAcrossPersistent and kernel["PrefetchAcrossPersistentMode"] == 1 and isPap and not isOptNLL:
```
nakajee, line 1550: Can we use prefetchAcrossPersistent ==0 and PrefetchAcrossPersistentMode==1?

Tensile/KernelWriter.py
```diff
@@ -2512,11 +2526,11 @@ def kernelBody( self, kernel, tensorParametersA, tensorParametersB ):
       kl.append(self.globalReadIncrementAB(kernel, i, 0))
       kl.append(self.closeLoop(kernel, i, True))
 
-    if self.prefetchAcrossPersistent:
+    if self.prefetchAcrossPersistent and kernel["PrefetchAcrossPersistentMode"] != 1:
       kl.append(str(self.openPrefetchAcrossPersistent(kernel, isOptNLL=False)))
       kl += self.setupNewTile(kernel, self.tPA, self.tPB, isPap=True, isOptNLL=False)
       kl.append(str(self.closePrefetchAcrossPersistent(kernel, isOptNLL=False)))
-
+
```
nakajee, line 2533: Do we need to add space here?

Tensile/KernelWriter.py
AlexBrownAMD: No, PAPMode only matters if PAP is on (> 0)

nakajee: Do we need to care invalid combination (print rejected or force PrefetchAcrossPersistentMode to 0 or something else)?

nakajee: What I thought is that some of your code change does not have check for prefetchAcrossPersistent and it might cause some unexpected behavior.

AlexBrownAMD: I think all the changes are inside PK sections, so it shouldn't cause any problems - but, yea, probably best to add a reject for PAPMode 1 when PAP not enabled to be sure. I'll add that change.

nakajee:
Thanks. I do not have any further comment.

aazz44ss (approved):
LGTM

PR #1387, opened 2021-07-29
Add new capability for selecting a specific kernel index

afanfa:
Hi all,
this patch is meant to let the user choose a specific kernel identified by its index in the Master Solution Library.
Comments and suggestions are more than welcome.

Alessandro

### DIFF
--- a/Tensile/Source/lib/include/Tensile/MasterSolutionLibrary.hpp
+++ b/Tensile/Source/lib/include/Tensile/MasterSolutionLibrary.hpp
@@ -29,8 +29,8 @@
 #include <map>
 #include <memory>
 
+#include <Tensile/Debug.hpp>
 #include <Tensile/SolutionLibrary.hpp>
-
 #include <Tensile/Tensile.hpp>
 
 namespace Tensile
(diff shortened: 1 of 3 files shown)


sdquiring:
First, we should be consistent in calling this a 'solution' index, not a 'kernel' index, since a solution can involve multiple kernels being invoked (for GSU).

Second, if this environment variable is set, then Tensile will only support one particular transpose, and one particular data type.  It may also have other restrictions in terms of problem/stride size since some solutions may have such limitations.  @zaliu I'm not sure if there's a need to communicate this better?

Finally, this would be best implemented in `MasterSolutionLibrary`: https://github.com/ROCmSoftwarePlatform/Tensile/blob/develop/Tensile/Source/lib/include/Tensile/MasterSolutionLibrary.hpp#L87 as the the `solutions` member is a map from SolutionIndex to Solution.  We just need to check the predicates and then either return the solution or `nullptr` if the predicates fail.

zaliu:
> @zaliu I'm not sure if there's a need to communicate this better?
Your comments are clear enough.  I'm sure @afanfa will communicate directly with us if clarifications are needed.  Thanks!

sdquiring:
Sorry, by communicate I mean by means of a comment in the code or a printed warning when this environment variable is active.

afanfa:
Thanks for the quick review. I will implement this functionality in MasterSolutionLibrary as suggested by @sdquiring with a better naming.

Tensile/Source/lib/include/Tensile/MasterSolutionLibrary.hpp
```diff
@@ -84,7 +84,17 @@ namespace Tensile
...
         {
-            return library->findBestSolution(problem, hardware, fitness);
+            const int solution_index = Debug::Instance().getSolutionIndex();
+
+            if(solution_index >= 0)
+            {
+                std::cout << "Tensile will use solution index: " << solution_index <<std::endl;
+                std::cout << "Warning: Tensile will only work for a particular transpose and data type." << std::endl;
+                std::cout << "Set TENSILE_SOLUTION_INDEX to a negative number to restore the default behavior." << std::endl;
+                return solutions.at(solution_index);
```
sdquiring, line 94: You still need to check the predicates. `(*solution->problemPredicate)(problem)` and `(*solution->hardwarePredicate)(hardware)` must both return `true` or we should return `nullptr` from here.

sdquiring (commented):
Great, thanks for implementing the changes.  One more quick note though.

sdquiring (approved):
Great! Ready to merge once CI passes.

afanfa:
Something didn't work with my rebase.

PR #1389, opened 2021-08-04
fix addrVgpr naming for atomic GSU

aazz44ss:
Fixing inconsistent naming


zaliu:
@cgmillette The 2 gfx900 pipelines will never finish; the 2 failed hipclang-4.1 pipelines are no longer relevant because we've switched over to compute-rocm-dkms-no-npi-hipclang and all active Tensile pipelines passed.

cgmillette (approved):
LGTM!

PR #1393, opened 2021-08-09
Client, Transform the data type of sources to the type of Accumulator when doing the multiplication.

vin-huang:
At the Client, cast the Data/DestData/ComputeDataType to the type of Accumulator before doing the multiply op when the highPrecisionAccumulate is true.
Exception: I8/I32/I32, I8x4/I32/I32, B/B/S + !HPA

**Issue:** 
  For cases H/H/H + HPA, H/H/S + HPA, H/S/S+ HPA,  B/B/S + HPA, B/S/S+ HPA, 
  When Client is calculating the rC=A*B and D = Alpha * (rC) + Beta * C.
  The type of A, B, Beta, and Alpha are not casting to the higher precision data type even HPA is enabled.
  This may cause the result value which precision not as expected (expected same as the HIP/ASM kernel)

**Solution:**
  Transform the data type of A/B to Accumulator if A!=ACC or B!=ACC, but filter out cases, I8/I32/I32 and I8x4/I32/I32

**Implementation:**
Three cases need to check the need_trasnform
- AxB
- alpha x rC
- Beta x C

The idea here is whether these two input types are consistent or not; if not, then needs to do the type transform.
Also, we need to filter out cases when A=i8 or A=i8x4

Here are conditions to do the transformation
- AxB : (A!=ACC or B!=ACC) and A!=I8 and A!=I8x4
- Alpha x rC : (Alpha!=ACC or rC!=ACC)
- Beta x C : (Beta!=ACC or C!=ACC 

And here is the result of doing the transformation or not, the **expected need cast** is the same as the **result of needAccumCast**. 

  |   |   |   |   | expected need cast | | | result of needAccumCast| | |
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
**A/B** | **C/D** | **Alpha/Beta** | **ACC** | **HPA** | **A x B** | **alpha x rC** | **Beta x C** | **AxB** | **Alpha x rC** | **Beta x C**
B | B | S | B | N | N | :heavy_check_mark: | :heavy_check_mark: | N | :heavy_check_mark: | :heavy_check_mark:
C | C | C | C | N | N | N | N | N | N | N
D | D | D | D | N | N | N | N | N | N | N
H | H | H | H | N | N | N | N | N | N | N
I | I | I | I | N | N | N | N | N | N | N
S | S | S | S | N | N | N | N | N | N | N
Z | Z | Z | Z | N | N | N | N | N | N | N
B | B | S | S | Y | :heavy_check_mark: | N | :heavy_check_mark: | :heavy_check_mark: | N | :heavy_check_mark:
B | S | S | S | Y| :heavy_check_mark: | N | N | :heavy_check_mark: | N | N
H | H | H | S | Y | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark:
H | H | S | S | Y | :heavy_check_mark: | N | :heavy_check_mark: | :heavy_check_mark: | N | :heavy_check_mark:
H | S | S | S |Y| :heavy_check_mark: | N | N | :heavy_check_mark: | N | N
I8 | I | I | I | Y | N | N | N | N | N | N
I8x4 | I | I | I | Y | N | N | N | N | N | N

### DIFF
--- a/Tensile/Source/client/source/Reference.cpp
+++ b/Tensile/Source/client/source/Reference.cpp
@@ -118,6 +118,26 @@ namespace Tensile
             throw std::runtime_error(msg.c_str());
         }
 
+        template <typename Accumulator, typename TypeL, typename TypeR>
+        inline Accumulator multiply(TypeL l, TypeR r)
+        {
+            /* Transform the data type from TypeL/TypeR to Accumulator if TypeL!=ACC or TypeR!=ACC, but filter out cases, I8/I32/I32 and I8x4/I32/I32
+             *
+             * There are three cases of doing multiplication and their conditions to do transform or not are as below.
+             * 1. AxB : (A!=ACC or B!=ACC) and A!=I8 and A!=I8x4
+             * 2. Alpha x rC :  (Alpha!=ACC or rC!=ACC)
+             * 3. Beta x C : (Beta!=ACC or C!=ACC)
+            */
+            constexpr bool needAccumCast
+                = !(std::is_same<TypeL, Accumulator>() && std::is_same<TypeR, Accumulator>())
+                  && !std::is_same<TypeL, Int8>() //case I8/I32/I32, I8 be implicitly cast to int.
+                  && !std::is_same<TypeL, Int8x4>(); //case I8x4/I32/I32, I8x4 overloading the op*.
+
+            using LMultT = std::conditional_t<needAccumCast, Accumulator, TypeL>;
+            using RMultT = std::conditional_t<needAccumCast, Accumulator, TypeR>;
+            return static_cast<Accumulator>(static_cast<LMultT>(l) * static_cast<RMultT>(r));
+        }
+
         template <typename Inputs, typename Accumulator>
         void ReferenceSolution<Inputs, Accumulator>::SolveCPU(ContractionProblem const& problem,
                                                               Inputs const&             inputs,
@@ -292,7 +312,7 @@ namespace Tensile
                                 bVal = Transform<typename Inputs::BType>::Input(
                                     inputs.b[bIndex + (bI * bStride) - zpB.padStart], bConjugate);
 
-                            value += static_cast<Accumulator>(aVal * bVal);
+                            value += multiply<Accumulator>(aVal, bVal);
 
                             if(0)
                             {
@@ -312,12 +332,13 @@ namespace Tensile
                 auto dIndex = d.index(dCoord);
 
                 // Ensure zero*nan returns zero
-                auto beta = static_cast<typename Inputs::DType>(inputs.beta);
-                auto zero = static_cast<typename Inputs::DType>(0);
+                auto beta = inputs.beta;
+                auto zero = static_cast<typename Inputs::BetaType>(0);
 
-                inputs.d[dIndex] = static_cast<typename Inputs::DType>(inputs.alpha)
-                                       * static_cast<typename Inputs::DType>(value)
-                                   + ((beta == zero) ? zero : beta * inputs.c[cIndex]);
+                inputs.d[dIndex] = static_cast<typename Inputs::DType>(
+                    multiply<Accumulator>(inputs.alpha, value)
+                    + ((beta == zero) ? static_cast<Accumulator>(zero)
+                                      : multiply<Accumulator>(beta, inputs.c[cIndex])));
             }
         }
 
@@ -587,8 +608,7 @@ namespace Tensile
                                               << " aIndex=" << aIndex << " bIndex=" << bIndex
                                               << " aVal=" << aVal << " bVal=" << bVal << "\n";
                                 }
-
-                                value += static_cast<Accumulator>(aVal * bVal);
+                                value += multiply<Accumulator>(aVal, bVal);
                             }
                         std::vector<size_t> dCoord(outputTensor.dimensions(), 0);
                         dCoord[formatD.activation().batchPosition()]   = n;
@@ -605,8 +625,8 @@ namespace Tensile
                                       << spatialCoord[1] << "," << spatialCoord[0]
                                       << " dIndex=" << dIndex << " value=" << value << "\n";
                         }
-                        inputs.d[dIndex] = static_cast<typename Inputs::DType>(inputs.alpha)
-                                           * static_cast<typename Inputs::DType>(value);
+                        inputs.d[dIndex] = static_cast<typename Inputs::DType>(
+                            multiply<Accumulator>(inputs.alpha, value));
                     }
         }
 


Tensile/Source/client/source/Reference.cpp
```diff
@@ -118,6 +118,42 @@ namespace Tensile
...
+        constexpr bool is_hpa = !(std::is_same<typename Inputs::AType, Accumulator>() && std::is_same<typename Inputs::BType, Accumulator>() && std::is_same<typename Inputs::DType, Accumulator>())
+                                && !(std::is_same<typename Inputs::AType, Int8>() && std::is_same<typename Inputs::DType, int32_t>())    //case I8/I32/I32
+                                && !(std::is_same<typename Inputs::AType, Int8x4>() && std::is_same<typename Inputs::DType, int32_t>()); //case I8x4/I32/I32
+
+        /* is_hpa and (A!=B or A!=Accumlator or B!=Accumlator) */
+        template<typename A, typename B, typename Inputs, typename Accumulator>
+        constexpr bool need_transform = is_hpa<Inputs, Accumulator> && !(std::is_same<A, Accumulator>() && std::is_same<B, Accumulator>() && std::is_same<A, B>());
+
+        template <typename A, typename B, typename Inputs, typename Accumulator>
+        Accumulator multiply(A a, B b, typename std::enable_if<!need_transform<A, B, Inputs, Accumulator>>::type* = nullptr)
```
rosenrodt, line 134: Replacing `enable_if` with a `bool` reduces code verbosity and remove the need for the template parameter `Input` as your `enable_if` requires it.

Bonus: rearranging `typename Accumulator` to the leftmost side, the rest of them can even be deduced from the function arguments.

```c++
// declaration
template<typename Accumulator, typename A, typename B>
Accumulator multiply(A a, B b, bool need_transform);

// user code
//multiply<typename Inputs::AType, typename Inputs::BType, Inputs, Accumulator>(aVal, bVal);
multiply<Accumulator>(a, b, true);  // need_transform
multiply<Accumulator>(a, b, false); // non need_transform
```

vin-huang: @rosenrodt 
Thanks for your advice, however using bool at the function parameter would be another implement way.
The idea here is to avoid the check at runtime, so use enable_if here to use the template mechanism to let the compiler does the check things. Another reason is because of the function template doesn't allow partial specification so, using the enable_if to generate functions is one way to make the idea happened.

rosenrodt:
Gee we're hitting Github API rate limit again. That's how Integration test failed. Wait for another hour and try restart it

```
HTTP request sent, awaiting response... 403 rate limit exceeded
2021-08-11 07:03:55 ERROR 403: rate limit exceeded.
```

Tensile/Source/client/source/Reference.cpp
```diff
@@ -605,8 +669,9 @@ namespace Tensile
                                       << spatialCoord[1] << "," << spatialCoord[0]
                                       << " dIndex=" << dIndex << " value=" << value << "\n";
                         }
-                        inputs.d[dIndex] = static_cast<typename Inputs::DType>(inputs.alpha)
-                                           * static_cast<typename Inputs::DType>(value);
+                        inputs.d[dIndex] = static_cast<typename Inputs::DType>(
+                            multiply<typename Inputs::AlphaType, Accumulator, Inputs, Accumulator>(
```
jichangjichang, line 673: can we simply use alphatype for alpha computing??

static_cast<typename Inputs::DType>(
inputs.allpha * static_cast<typename Inputs::AlphaType>(value)
)

vin-huang: Using alphatype here should against the case  H/H/H + HPA
In this case, the type of rC is float and alphtype is half.
this is also the reason for not using DType.

Tensile/Source/client/source/Reference.cpp
```diff
@@ -118,6 +118,60 @@ namespace Tensile
             throw std::runtime_error(msg.c_str());
         }
 
+        /* high precision accumulate: dist type != Accumulator or src type != Accumulator, such as H/H/H + HPA, H/H/S + HPA, B/B/S + HPA, B/S/S + HPA
+         * Ignore types: I8/I32/I32, I8x4/I32/I32
+         */
+        template <typename Inputs, typename Accumulator>
+        constexpr bool is_hpa
```
jichangjichang, line 125: Why not just check is_hpa = (D Type !=AccType)?

vin-huang: Consider the case, B/S/S +HPA,
both types of D and ACC are S. If we just check the D and ACC types then the _is_hpa()_ will return false.
However, in this example, the HPA is true and we want to raise the precision when doing _rC = A x B. => rC = float(A) x float(B)_.
So, extended to check the type of A and B will be helpful.

Tensile/Source/client/source/Reference.cpp
```diff
@@ -118,6 +118,60 @@ namespace Tensile
...
+                   && std::is_same<typename Inputs::DType, int32_t>()) //case I8/I32/I32
+              && !(std::is_same<typename Inputs::AType, Int8x4>()
+                   && std::is_same<typename Inputs::DType, int32_t>()); //case I8x4/I32/I32
+
+        /* is_hpa and (A!=B or A!=Accumlator or B!=Accumlator) */
+        template <typename A, typename B, typename Inputs, typename Accumulator>
+        constexpr bool need_transform
+            = is_hpa<
+                  Inputs,
+                  Accumulator> && !(std::is_same<A, Accumulator>() && std::is_same<B, Accumulator>() && std::is_same<A, B>());
```
jichangjichang, line 139: seems don't need to check  A==B.
If A==B, it means that one of condition A!=ACC and condition B!=ACC must be true.

jichangjichang: can we simply use this check??
need_transform = (A!=ACC or B!=ACC or A! = I8 or A!=I8x4)

vin-huang: Yes, simpling the condition is possible under the current type combinations, 

Three cases need to check the need_trasnform.
- AxB
- alpha x rC
- Beta x D

So the idea here is whether these two input types are consistent or not; if not, then needs to do the type transform.
Also, we need to filter out cases when A=i8 or A=8x4

Here is the idea,
- AxB :  (A!=ACC or B!=ACC) and A!=I8 and A!=I8x4
- alpha x rC :   ((alpha!=ACC or rC!=ACC) and A!=I8   and A!=I8x4
- Beta x D : (beta!=ACC or D!=ACC )and A!=I8 and A!=I8x4

And here is the result of doing the transformation or not, the **expected need_transmform** is the same as the **result_need_transform**. 
>NoteL If the need_transform is true, the two inputs will be cast to ACC before doing the multiplication.
>For the case, B/B/S +!HPA, which beta/alpha will cast to the type B, if we want to use higher precision(type S) then still need to specialize this case.


  |   |   |   |   | expected need_transform | | | result need_transform| | |
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
A/B | C/D | Alpha/Beta | ACC | HPA | A x B | alpha x rC | Beta x C | AxB | alpha x rC | beta x D
B | B | S | B | FALSE | FALSE | TRUE | TRUE | FALSE | TRUE | TRUE
C | C | C | C | FALSE | FALSE | FALSE | FALSE | FALSE | FALSE | FALSE
D | D | D | D | FALSE | FALSE | FALSE | FALSE | FALSE | FALSE | FALSE
H | H | H | H | FALSE | FALSE | FALSE | FALSE | FALSE | FALSE | FALSE
I | I | I | I | FALSE | FALSE | FALSE | FALSE | FALSE | FALSE | FALSE
S | S | S | S | FALSE | FALSE | FALSE | FALSE | FALSE | FALSE | FALSE
Z | Z | Z | Z | FALSE | FALSE | FALSE | FALSE | FALSE | FALSE | FALSE
B | B | S | S | TRUE | TRUE | FALSE | TRUE | TRUE | FALSE | TRUE
B | S | S | S | TRUE | TRUE | FALSE | FALSE | TRUE | FALSE | FALSE
H | H | H | S | TRUE | TRUE | TRUE | TRUE | TRUE | TRUE | TRUE
H | H | S | S | TRUE | TRUE | FALSE | TRUE | TRUE | FALSE | TRUE
H | S | S | S | TRUE | TRUE | FALSE | FALSE | TRUE | FALSE | FALSE
I8 | I | I | I | TRUE | FALSE | FALSE | FALSE | FALSE | FALSE | FALSE
I8x4 | I | I | I | TRUE | FALSE | FALSE | FALSE | FALSE | FALSE | FALSE

Tensile/Source/client/source/Reference.cpp
```diff
@@ -118,6 +118,38 @@ namespace Tensile
...
+         * 
+         * There are three cases of doing multiplication and their conditions to do transform or not are as below.
+         * 1. AxB : (A!=ACC or B!=ACC) and A!=I8 and A!=I8x4
+         * 2. alpha x rC :  ((alpha!=ACC or rC!=ACC) and A!=I8 and A!=I8x4
+         * 3. Beta x D : (beta!=ACC or D!=ACC )and A!=I8 and A!=I8x4
+        */
+        template <typename A, typename B, typename Inputs, typename Accumulator>
+        constexpr bool need_transform
+            = !(std::is_same<A, Accumulator>() && std::is_same<B, Accumulator>())
+              && !std::is_same<typename Inputs::AType, Int8>()    //case I8/I32/I32
```
jichangjichang, line 131: Do we really need to consider "Inputs::AType"? Using "A" should be more reasonable and simple.

vin-huang: The idea of using Inputs::AType is wanting to describe clearly that the filter cases are for two specific ContractionInputs. 

And, yes, we can use A instead Inputs::AType, which will get the same results as expected.
Since the value of A comes from
1. Matrix A
2. Alpha
3. Beta

And only case 1 which type has a chance to be I8 or I8x4.
I will use A instead Inputs::AType,

jichangjichang (commented):
Please make sure this change won't make the CI testing time longer.
Please consider using inline for better efficient for the multiply function.

vin-huang:
> Please make sure this change won't make the CI testing time longer.
> Please consider using inline for better efficient for the multiply function.

Modified to use the inline function.

Checking the execution time by running "prec_checkin" locally (on gfx908). 
After comparing the execution time of using the inline method with the original one, the execution time didn't change a lot.

Here's the execution time (minutes) of 3 cases
|Original|w/o inline|w/ inline|
|--|--|--|
|56:37|58:08|55:04|

jichangjichang:
> > Please make sure this change won't make the CI testing time longer.
> > Please consider using inline for better efficient for the multiply function.
> 
> Modified to use the inline function.
> 
> Checking the execution time by running "prec_checkin" locally (on gfx908).
> After comparing the execution time of using the inline method with the original one, the execution time didn't change a lot.
> 
> Here's the execution time (minutes) of 3 cases
> 
> Original	w/o inline	w/ inline
> 56:37	58:08	55:04

LGTM, once you think this PR is ready, please remove the "draft" mark and loop in more tensile teammates for the reviewing/merging process.

zaliu:
Please rebase the latest develop into your PR so that the latest flake8 and clang-format fixes (#1396) are part of this PR.

Please make sure your C++ modifications are ROCm 4.3 clang-format clean.  To verify that, on a machine with ROCm 4.3 installed, run the following command in your local repo:
git ls-files -z *.cc *.cpp *.h *.hpp *.cl *.h.in *.hpp.in *.cpp.in | xargs -0 /opt/rocm/llvm/bin/clang-format -style=file -i

As far as I can tell, your Tensile/Source/client/source/Reference.cpp needs clang-format fixes.  Thanks.

vin-huang:
> Please rebase the latest develop into your PR so that the latest flake8 and clang-format fixes (#1396) are part of this PR.
> 
> Please make sure your C++ modifications are ROCm 4.3 clang-format clean. To verify that, on a machine with ROCm 4.3 installed, run the following command in your local repo:
> git ls-files -z *.cc *.cpp *.h *.hpp *.cl *.h.in *.hpp.in *.cpp.in | xargs -0 /opt/rocm/llvm/bin/clang-format -style=file -i
> 
> As far as I can tell, your Tensile/Source/client/source/Reference.cpp needs clang-format fixes. Thanks.

@zaliu, thanks for your review, updated the patch included branch rebase and clang-format fix.

jichangjichang (approved):
LGTM

cgmillette:
Tensile has clang-format githooks. 

When doing any work on a freshly cloned Tensile repo, please make sure to run the following from the root directory:
``` .githooks/install ```

The above will configure the pre-checkin clang formatting run when commits are made.

File: Tensile/Source/client/source/Reference.cpp [hunk 6]
```diff
@@ -118,6 +118,34 @@ namespace Tensile
...
+        }
+
+        /* For A!=ACC or B!=ACC */
+        template <typename A, typename B, typename Accumulator>
+        inline Accumulator multiply(
+            A a, B b, typename std::enable_if<need_transform<A, B, Accumulator>>::type* = nullptr)
+        {
+            return static_cast<Accumulator>(a) * static_cast<Accumulator>(b);
+        }
+
```
cgmillette, line 148: From a big-picture standpoint, I don't feel these are a good fit here.
* The new need_transform variable is now confusing because it has nothing to do with the above class Transform<T>.
* The new multiply functions are specific to the SolveCPU function and have no other use in the Client namespace
* The calling of a multiply function feels unnatural, as the function template a) cannot be automatically deduced and b) is not in operator form as would normally be expected.

Tensile/Source/client/source/Reference.cpp
```diff
@@ -292,7 +320,9 @@ namespace Tensile
                                 bVal = Transform<typename Inputs::BType>::Input(
                                     inputs.b[bIndex + (bI * bStride) - zpB.padStart], bConjugate);
 
-                            value += static_cast<Accumulator>(aVal * bVal);
+                            value += multiply<typename Inputs::AType,
```
cgmillette, line 323: I would suggest declaring a constexpr flag like before. Then add type aliases for multiplicands: this case for A and B.

``` 
/* Transform the data type of A/B to Accumulator if A!=ACC or B!=ACC, but filter out cases, I8/I32/I32 and I8x4/I32/I32
 * 
 * There are three cases of doing multiplication and their conditions to do transform or not are as below.
 * 1. AxB : (A!=ACC or B!=ACC) and A!=I8 and A!=I8x4
 * 2. Alpha x rC :  (Alpha!=ACC or rC!=ACC)
 * 3. Beta x C : (Beta!=ACC or C!=ACC)
 */
  constexpr bool needAccumCast
      = !(std::is_same<typename Inputs::AType, Accumulator>() && std::is_same<Inputs::BType, Accumulator>())
        && !std::is_same<typename Inputs::AType, Int8>() //case I8/I32/I32
        && !std::is_same<typename Inputs::AType, Int8x4>(); //case I8x4/I32/I32 
        
  using AMultT = std::conditional_t<needAccumCast, Accumulator, typename Inputs::AType>;
  using BMultT = std::conditional_t<needAccumCast, Accumulator, typename Inputs::BType>;
   
  value += static_cast<Accumulator>(static_cast<AMultT>(aVal) * static_cast<BMultT>(bVal));
        
```

vin-huang: @cgmillette Thanks for your suggestion.
The idea of using the type aliases is helpful, which can be used to instead the enable_if in the template's parameter 
I would like to keep the function **_multiply_** to reduce the code because 5 places need to do the CAST.

The new function will look like below. If you have any idea please kindly let me know. Thank you.
```
        template <typename A, typename B, typename Accumulator>
        inline Accumulator multiply(A a, B b)
        {
            constexpr bool needAccumCast
                = !(std::is_same<A, Accumulator>() && std::is_same<B, Accumulator>())
                  && !std::is_same<A, Int8>() //case I8/I32/I32
                  && !std::is_same<A, Int8x4>(); //case I8x4/I32/I32

            using AMultT = std::conditional_t<needAccumCast, Accumulator, A>;
            using BMultT = std::conditional_t<needAccumCast, Accumulator, B>;
            return static_cast<Accumulator>(static_cast<AMultT>(a) * static_cast<BMultT>(b));
         }
```

cgmillette: Ok, if we are to keep this function, then last thing I will request is from what @rosenrodt suggested above:

_Bonus: rearranging typename Accumulator to the leftmost side, the rest of them can even be deduced from the function arguments._

``` 
value += multiply<Accumulator>(a, b);    // A bit cleaner than value += multiply<TypeA, TypeB, Accumulator>(a, b); 
```

vin-huang: Thanks for your review.
Updated the patch.

rosenrodt (approved):

cgmillette (approved):
LGTM

PR #1394, opened 2021-08-12
Print solution index selected after search

afanfa:
Prints the solution index selected when DB mode is enabled. It works for both Binary and Naive search.
This information is useful for a proper use of TENSILE_SOLUTION_INDEX.

### DIFF
--- a/Tensile/Source/lib/include/Tensile/PropertyMatching.hpp
+++ b/Tensile/Source/lib/include/Tensile/PropertyMatching.hpp
@@ -346,7 +346,7 @@ namespace Tensile
                     std::cout << "Considered " << considered << "% of entries." << std::endl;
                 }
 
-                if(T_Debug)
+                if(T_Debug && bestMatch)
                     std::cout << "Solution index selected: " << bestMatch->index << std::endl;
 
                 return std::make_tuple(bestMatch, bestDistance);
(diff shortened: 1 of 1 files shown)


Tensile/Source/lib/include/Tensile/PropertyMatching.hpp
```diff
@@ -346,6 +346,9 @@ namespace Tensile
                     std::cout << "Considered " << considered << "% of entries." << std::endl;
                 }
 
+                if(T_Debug)
+                    std::cout << "Solution index selected: " << bestMatch->index << std::endl;
```
sdquiring, line 350:
```suggestion
                if(T_Debug && bestMatch)
                    std::cout << "Solution index selected: " << bestMatch->index << std::endl;
```

It is possible for `bestMatch` to be null here.

Tensile/Source/lib/include/Tensile/PropertyMatching.hpp
```diff
@@ -414,6 +417,9 @@ namespace Tensile
                     iter++;
                 }
 
+                if(T_Debug)
+                    std::cout << "Solution index selected: " << bestMatch->index << std::endl;
```
sdquiring, line 421:
```suggestion
                if(T_Debug && bestMatch)
                    std::cout << "Solution index selected: " << bestMatch->index << std::endl;
```

sdquiring (commented):
Apologies for the delay, just a minor issue.

sdquiring (approved):

PR #1402, opened 2021-09-08
Remove unused old client code

AlexBrownAMD:
Old client has been phased out for a while. As long as it is no longer used anywhere, we can probably safely clean up the old code from our repo.


zaliu:
@AlexBrownAMD Do we really want to remove the global parameter NewClient?  It's used everywhere, not just in the Tensile repo.  Can we retain NewClient either as a no-op or issue a warning that we only support the new client?

AlexBrownAMD:
> @AlexBrownAMD Do we really want to remove the global parameter NewClient? It's used everywhere, not just in the Tensile repo. Can we retain NewClient either as a no-op or issue a warning that we only support the new client?

Yea thats a good idea. I'll leave the parameter there with a message if set to anything other than 2.

zaliu:
@AlexBrownAMD Please resolve the conflict on Tensile/Source/CmakeLists.txt due to the merging of #1403, thanks.

benjaminulmer (approved):
This is something I was planning on doing once my plate cleared a bit, so thanks for saving me the work.

For some reason when I initially wrote TensileRetuneLibrary.py, I ran into issues if I didn't copy the old client files into the working directory. I would bet this PR fixes whatever that issue was and we could update that file now as well. However, afaik I'm the only one who uses that tuning pipeline, so I don't think those changes are necessary for this PR. 

LGTM assuming CI behaves nicely.

cgmb:
Does this mean rocBLAS should be dropping one branch of this if/else block? https://github.com/ROCmSoftwarePlatform/rocBLAS/blob/4962aa3acfa527be5032e68787e9ba635b03f980/library/src/CMakeLists.txt#L45-L95

AlexBrownAMD:
> Does this mean rocBLAS should be dropping one branch of this if/else block? https://github.com/ROCmSoftwarePlatform/rocBLAS/blob/4962aa3acfa527be5032e68787e9ba635b03f980/library/src/CMakeLists.txt#L45-L95

Yea, old client is no longer used or supported

TorreZuk:
I was going to remove the rocblas side earlier but was waiting on Alex's approval, I can clean it out once this is merged.

AlexBrownAMD:
> I was going to remove the rocblas side earlier but was waiting on Alex's approval, I can clean it out once this is merged.

Thanks Torre! There's an issue with the HostLibraryTest build that I need to resolve, hoping to get to that later today.

zaliu:
@AlexBrownAMD please re-open when you have time to deal with the CI failures.  Thanks.

AlexBrownAMD:
Fixed HostLibraryTests

PR #1404, opened 2021-09-16
Fix Navi gpu target to logic mapping

JeremyAdamHart:
gfx1011 and gfx1012 are mapped to the incorrect logic names, changed them to the correct ones.

### DIFF
--- a/Tensile/Common.py
+++ b/Tensile/Common.py
@@ -268,7 +268,7 @@ architectureMap = {
   'gfx906':'vega20', 'gfx906:xnack+':'vega20', 'gfx906:xnack-':'vega20',
   'gfx908':'arcturus','gfx908:xnack+':'arcturus', 'gfx908:xnack-':'arcturus',
   'gfx90a':'aldebaran', 'gfx90a:xnack+':'aldebaran', 'gfx90a:xnack-':'aldebaran',
-  'gfx1010':'navi10', 'gfx1011':'navi11', 'gfx1012':'navi12', 'gfx1030':'navi21'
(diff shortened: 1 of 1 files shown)


zaliu:
What's the rationale of this change?  Also,  Tests/unit/test_Component.py defines navi12, even though it is not used, should it be changed to navi14?

JeremyAdamHart:
> What's the rationale of this change?

When using navi12 or navi14 kernels, and specifying gfx1011 and gfx1012 in the --architecture list for TensileCreateLibrary, the incorrect mapping results in the appropriate logic files not being picked up. I ran into this issue working on a fork of Tensile, and thought it should be fixed in the mainline instead of just the fork.

> Also, Tests/unit/test_Component.py defines navi12, even though it is not used, should it be changed to navi14?

I saw this as well. I'm not sure if it was intended to refer to navi12 or navi14, so I left it alone for now.

zaliu:
Since the marketing names for gfx1011 and gfx1012 are "Navi 12" and "Navi 14" respectively, I have no objections to this change.  Thanks.

benjaminulmer (approved):
The CI issue with gfx1030 seems to be caused by a kernel segfaulting and is affecting other PRs, so I think this is safe to merge.

zaliu (approved):
The extended pipeline failure on gfx1030 happens randomly, and it is not related to this change.

zaliu:
@JeremyAdamHart please confirm with @sdquiring on the navi12 definition in Tests/unit/test_Component.py to make sure if it is for navi14 or not.  Thanks.

PR #1406, opened 2021-09-21
merge staging 998f23ee into master with GO from CQE

zaliu:


amcamd (approved):
- this commit 998f23ee corresponds to rocBLAS commit bba0b138, and the rocBLAS commit has the GO verdict from Manivannan for merging from staging into master

PR #1407, opened 2021-09-21
Tensile uses static C++ msgpackc header library

cgmillette:
Remove unused shared lib linkage, due to using only C++ header library for msgpackc.
Creates problems downstream with other projects linking TensileHost library

### DIFF
--- a/Tensile/Source/lib/CMakeLists.txt
+++ b/Tensile/Source/lib/CMakeLists.txt
@@ -99,14 +99,9 @@ if(TENSILE_USE_LLVM OR TENSILE_USE_MSGPACK)
 endif()
 
 if(TENSILE_USE_MSGPACK)
+    find_package(msgpack REQUIRED)
     target_compile_definitions(TensileHost PUBLIC -DTENSILE_MSGPACK=1)
-
-    find_package(msgpack)
-    if(TARGET msgpackc-static)
-        target_link_libraries(TensileHost PRIVATE msgpackc-static)
-    else()
-        target_link_libraries(TensileHost PRIVATE msgpackc)
-    endif()
+    target_include_directories(TensileHost PRIVATE "${msgpack_DIR}/include")
 endif()
 
 if(TENSILE_USE_LLVM)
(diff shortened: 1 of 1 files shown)


TorreZuk (commented):
Fine by me if you don't actually link nor need any compiler flags set from the target config.

cgmillette:
This is what the msgpackc readme.md suggests, up until v3.
Once we get to v4, then we will have to apply a target config as such:
```target_link_libraries(your_target_name <PRIVATE/PUBLIC/INTERFACE> msgpackc-cxx)```

TorreZuk:
Okay but you may want to add it in commented out until v4.   windows may force us there before linux as vcpkg seems to grab the latest.  You can maybe make it future readyby using if on the new target msgpackc-cxx is defined, but instead of linking maybe they define compile interface flags you can just apply.

zaliu:
@cgmillette did you try to build rocBLAS with `./install.sh --static`?  Just wondering.  Thanks.

cgmb:
> vcpkg seems to grab the latest

The version you can install from vcpkg is tied to the vcpkg version. However, a-la-cart version selection works in manifest mode. I have a branch that works with manifest mode for rocsolver, but it still needs a bit of cleanup.

cgmillette:
> @cgmillette did you try to build rocBLAS with `./install.sh --static`? Just wondering. Thanks.

I can try this out today

TorreZuk:
In rocblas library/src/CMakeLists.txt we have STATIC_DEPENDS PACKAGE msgpack for which will still find the library for static linking even if nothing is actual used in it.  Thus If you can say for v4 and beyond you won't link against the library and can take target compile/interface options only from their cmake target and not restore the target_link_libraries then we can remove this from rocblas too.

cgmillette:
> > @cgmillette did you try to build rocBLAS with `./install.sh --static`? Just wondering. Thanks.
> 
> I can try this out today

@zaliu Build is successful with latest changes :)

sdquiring (approved):

PR #1408, opened 2021-09-21
Update tensile_client executable to std=c++14

cgmillette:
Update the tensile_client executable to use c++14:
- Linked libraries TensileHost and TensileClient are built with c++14
- Warnings when building main.cpp due to inclusion of headers with constexpr functions requiring c++14 (e.g. BitFieldGenerator object).
- Good to keep code standards consistent


TorreZuk:
rocblas is now using c++17 so if it helps you could also try using that, would be nice if we can match.

cgmillette:
Somehow I remember there was an argument for staying on c++14. Will check ;)

zaliu (approved):

PR #1416, opened 2021-10-01
Remove .dir-locals.el

cgmb:
emacs-level formatting is overruled by clang format, so this file is no longer needed. See https://github.com/ROCmSoftwarePlatform/rocBLAS-internal/pull/871 (rocBLAS 1e13f40).


cgmillette (approved):
Fine with me

zaliu (approved):

PR #1418, opened 2021-10-03
Added parameter which controls the number of files kernels are split into to improve compile speed

lringham:
- Added a parameter to control the number of files that kernels are merged into to better parallelize kernel compilation
    - Regardless of the parallel-jobs value, not all cores were being used to compile kernels when all kernels were in one src file.
    - Merging the kernels into a few src files allowed the compiler to fulled parallelize compilation
    - Can be controlled with --num-merged-files command line arg
- Fixed various spelling mistakes
- Fixed warning related to EmbedLibrary not being found in globalParameters
- Tests passed: Linting, Unit Tests, Client Test


lringham:
Added Cory Bloor as a review because I consulted with him on this issue.

cgmb:
Awesome. When compiling for a single architecture (like on my development machine), the lack of parallelism in the compilation of the Tensile source kernels was the slowest part of compiling rocBLAS (and the tight coupling between rocBLAS and rocSOLVER meant that I had to do that often). I'll give this a try and let you know what sort of improvement I see on my dev machine.

With respect to the design of this feature, there's one thing that you're missing which I would like to see. Ideally, the kernels would be distributed among the source files in such a way that small changes in input result in small changes in output. For example, it would be nice if adding a few source kernels to a single input logic file would only change one generated source file, and the others would remain identical. That would help to increase the frequency of build cache hits once SWDEV-283320 is finished (which will fix problems related to using ccache with device-only compilation). In your current setup, any change in the number of input kernels is going to result in all output source files changing, causing them all to be build cache misses. That said, I'd be glad to have this change regardless of whether you can find a way to improve that or not. I realize it may be quite difficult to achieve that ideal.

I took a look through the code and have provided some nitpicks. I know a thing or two about Python, but I'm not a Tensile developer. If anybody from the team disagrees with my comments, then their advice trumps mine. My perspective is really that of a user. Nevertheless, I really was _very_ picky, so despite the great work you did, I managed to find some suggestions on how to do things better.

PR #1419, opened 2021-10-08
Remove no longer supported benchmarking steps

benjaminulmer:
Several sections of BenchmarkProblems in config files have been unofficially unsupported for quite some time now:
- InitialSolutionParameters
- BenchmarkForkParameters
- JoinParameters
- BenchmarkJoinParameters
 
Previously, Tensile would simply fail with cryptic runtime errors if these steps had non-null values in a config file. This PR officially removes support for them; Tensile will now exit with a helpful error message if any of these steps have non-null values and all the code related to them has been removed.

I also did a fairly extensive refactoring of the relevant files, mostly BenchmarkProblems.py and BenchmarkStructs.py.

These changes should also make future extensions to our benchmarking process easier now that the code is cleaner and all the unnecessary parts are gone.

### DIFF
--- a/Tensile/Tests/extended/stagger_u/big_skinny_A_NN.yaml
+++ b/Tensile/Tests/extended/stagger_u/big_skinny_A_NN.yaml
@@ -28,16 +28,6 @@ BenchmarkProblems:
 
     - # BenchmarkProblemSizeGroup - Standard
       InitialSolutionParameters:
-        - AssertMinApproxSize: [0]
-        - AssertStrideBEqual: [ { 0: 1 } ]
-        - BufferLoad: [True] #
-        - GlobalReadVectorWidth: [1]
-        - UseSgprForGRO: [1]
-        - VectorAtomicWidth: [1]
-        - DirectToLds: True
-        - EnableMatrixInstruction: True
-        - VectorWidth: [1]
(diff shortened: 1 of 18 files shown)


Tensile/Tests/extended/stagger_u/big_skinny_A_NN.yaml
```diff
@@ -28,16 +28,6 @@ BenchmarkProblems:
 
     - # BenchmarkProblemSizeGroup - Standard
       InitialSolutionParameters:
```
AlexBrownAMD, line 30: Do the section headers like this one have to stay in if the params are no longer supported?

AlexBrownAMD (approved):
Looks good to me

Tensile/Tests/extended/stagger_u/big_skinny_A_NN.yaml
benjaminulmer: Hmm, I suppose I did it this way for consistency with the other yaml test configs. Or at least, that's the best reason I can think of for why I didn't just outright remove them.

To answer the question though, no, the section headers are not required. All the other yamls should be updated as well though, and I think there might be some other quirks that should be fixed as well. I'll do that as a separate PR.

PR #1422, opened 2021-10-12
Remove parts of ClientWriter related to old client

AlexBrownAMD:
Prints warning if LibraryClient step used in yaml


benjaminulmer:
There are no internal workflows that make use of it afaik. However, I have highlighted it as a useful option to some other users in the past.

As I see it, there are two useful things it currently does:
1) Invokes TensileCreateLibrary on the results of the Tensile run
2) Creates a config file for using this library with the benchmarking client

As you've pointed out, the way it goes about doing these things isn't necessarily the best. Nonetheless, while neither of these are particularly difficult tasks, I think there is some value in Tensile doing them automatically for new users. 

I don't think this is a hill I'm willing to die on, but I _would_ be willing to take a crack at reworking the LibraryClient step to just do the useful things (and do them in a better way).

zaliu:
Given the lack of consensus, can we defer this PR to after our impending feature complete merge?  If so, can @AlexBrownAMD please temporarily close this PR for now?  Thanks so much.

PR #1424, opened 2021-10-19
Adding fp16 alternate implementation

yoichiyoshida:
Adding an alternate implementation for fp16


benjaminulmer:
After a quick review it looks like all the needed parts are in place. I will do a more detailed review and testing tomorrow morning.

PR #1440, opened 2021-11-01
Added better control of parallelism and memory usage

lringham:
Tensile uses a lot of memory when running TensileCreateLibrary due to parallelized code.
This change allows control of parallelism and thus memory usage by letting the user specify the process count with  -j or --jobs

### DIFF
--- a/Tensile/Source/TensileCreateLibrary.cmake
+++ b/Tensile/Source/TensileCreateLibrary.cmake
@@ -33,7 +33,8 @@ function(TensileCreateLibraryCmake
     Tensile_LIBRARY_FORMAT
     Tensile_MERGE_FILES
     Tensile_SHORT_FILE_NAMES
-    Tensile_LIBRARY_PRINT_DEBUG )
+    Tensile_LIBRARY_PRINT_DEBUG
+    Tensile_CPU_THREADS )
 
 # make Tensile_PACKAGE_LIBRARY and optional parameter
 # to avoid breaking applications which us this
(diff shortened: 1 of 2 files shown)


lringham:
I have profiled the effect on memory usage (for all architectures) with different levels of `--jobs` combined with `--parallel-jobs`.
In general systems with:
- \>= 32GB ram - should be safe with `--parallel-jobs=4` and `--jobs=16` processes (build time ~1hr 16min)
- \>= 64GB ram - should be safe with `-parallel-jobs=10` and `--jobs=48` processes (build time ~45min) 

Screenshot from 2021-11-01 11-11-06

TorreZuk (commented):
I think you need to add the new argument into:
  # Single value settings
  set(oneValueArgs

TorreZuk:
My comment was for the function TensileCreateLibraryFiles as you have some local helper to decode these things.

lringham:
> My comment was for the function TensileCreateLibraryFiles as you have some local helper to decode these things.

I have now added the argument to oneValueArgs in TensileCreateLibraryFiles.

benjaminulmer (approved):
lgtm, but I'm pretty inexperienced with cmake

zaliu:
@lringham Please squash-merge after CI passes.  Thanks.

TorreZuk (approved):
LGTM.  Unfortunately rocblas uses this cmake code and shadows these variables making for some strange effects of unexpected parent scope variable referencing.    A separate PR but as I don't want to change these in rocblas and break spack builds, wonder if we could experiment tensile side with local scoping of the TensileCreateLibraryFiles function arguments via prefix "Tensile" -> "TensileFunctionArg" or similar.   This would ensure no arguments bypass the function call argument system.  I would expect this only arguments to be used to create the command, of course this may reveal some other global var usage.

cgmb:
`CACHE` variables have only a single global namespace, so it's commonly recommended that you always prefix your cache variables with your project name (e.g., Professional CMake §5.10). It mostly hasn't mattered, but this may be a good example of why that recommendation exists.

TorreZuk:
Ben's mention reminded me @lringham if you don't want to use locally scoped variable names in the function (in you are worried it could break other legacy calling usage where parameters are not passed in) I would suggest you could document the function to say you may bypass the function parameter list by setting the Tensile prefixed variables.    Not planning to change rocblas side for these variables.

cgmb:
If the `Tensile_`-prefixed cache variables are a supported method of configuring Tensile, is there somewhere that they are documented?

PR #1442, opened 2021-11-08
Delete unused dockerfiles

saadrahim:
These dockerfiles are from unused OSes and configurations. Cleaning up as they should not be used anymore.


zaliu (approved):

zaliu:
@saadrahim Please squash-merge.  Thanks.

PR #1448, opened 2021-11-23
New JSD metric and metric selection functionality

afanfa:
This PR adds a new distance metric (even though is not technically a metric) and sketches a new functionality in Tensile to select a particular distance metric based on the value of the USE_METRIC environment variable.
The assignment a certain integer value to a metric is currently done statically. A better solution would be to rely on a method (sketched) in the class Distance to retrieve the right index. 
Suggestions on how to implement this new functionality in a clean way are more than welcome.

### DIFF
--- a/Tensile/Source/lib/include/Tensile/Debug.hpp
+++ b/Tensile/Source/lib/include/Tensile/Debug.hpp
@@ -67,6 +67,8 @@ namespace Tensile
 
         bool enableDebugSelection() const;
 
+        bool useJSD() const;
+
         int getSolutionIndex() const;
 
     private:
@@ -75,6 +77,7 @@ namespace Tensile
         int  m_value;
         int  m_value2;
         bool m_naivePropertySearch = false;
+        bool m_jsd                 = false;
         bool m_debugSelection      = false;
         int  m_solution_index      = -1;
 
--- a/Tensile/Source/lib/include/Tensile/Debug.hpp
+++ b/Tensile/Source/lib/include/Tensile/Debug.hpp
@@ -67,7 +67,7 @@ namespace Tensile
 
         bool enableDebugSelection() const;
 
-        bool useJSD() const;
+        int useMetric() const;
 
         int getSolutionIndex() const;
 
@@ -77,7 +77,7 @@ namespace Tensile
         int  m_value;
         int  m_value2;
         bool m_naivePropertySearch = false;
-        bool m_jsd                 = false;
+        int  m_use_metric          = -1;
         bool m_debugSelection      = false;
         int  m_solution_index      = -1;
 
--- a/Tensile/Source/lib/include/Tensile/Debug.hpp
+++ b/Tensile/Source/lib/include/Tensile/Debug.hpp
@@ -67,19 +67,19 @@ namespace Tensile
 
         bool enableDebugSelection() const;
 
-        int useMetric() const;
+        std::string getMetric() const;
 
         int getSolutionIndex() const;
 
     private:
         friend LazySingleton<Debug>;
 
-        int  m_value;
-        int  m_value2;
-        bool m_naivePropertySearch = false;
-        int  m_use_metric          = -1;
-        bool m_debugSelection      = false;
-        int  m_solution_index      = -1;
+        int         m_value;
+        int         m_value2;
+        bool        m_naivePropertySearch = false;
+        bool        m_debugSelection      = false;
+        int         m_solution_index      = -1;
+        std::string m_metric              = "";
 
         Debug();
     };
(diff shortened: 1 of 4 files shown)


benjaminulmer (commented):
Instead of adding an index to each distance metric, could we just have the USE_METRIC variable be a string that matches the `type()` method? This would both make the variable more intuitive to use and would remove the need for the `get_distance_id()` method you sketched.

benjaminulmer:
Also I see this is targeting master. At some point the default branch on GitHub got switched to master instead of develop; could you retarget to develop please.

afanfa:
> Instead of adding an index to each distance metric, could we just have the USE_METRIC variable be a string that matches the `type()` method? This would both make the variable more intuitive to use and would remove the need for the `get_distance_id()` method you sketched.

Sure, I like the idea. I default to numbers because it makes easier for me to test the various options with a loop in a bash script. Using strings instead of numbers for metrics makes more sense and reduces the risk of errors.

zaliu:
@afanfa Please see @eidenyoshida's pinning clang-format to ROCm 4.3.1 [announcement](https://teams.microsoft.com/l/message/19:5818907f6c8b4637a19276b60b3cf5bf@thread.skype/1632348609913?tenantId=3dd8961f-e488-4e60-8e11-a82d994e183d&groupId=ac2278ba-8e20-4f39-b829-5ad102f224ef&parentMessageId=1632348609913&teamName=MLSE%20Libraries&channelName=Development%20Process&createdTime=1632348609913) for the appropriate version of clang-format to use.

benjaminulmer (approved):
It shouldn't be related to your changes, but I restarted the failed integration test just to be safe.

This doesn't change the default behaviour of Tensile at all and will be a useful feature for developing the two-tier solution selection approach. LGTM once CI clears.

afanfa:
thanks @benjaminulmer . I have a question about the error message returned when the metric specified by TENSILE_METRIC is not found:

rocBLAS error: Could not load /home/alex/rocBLAS/build/release/rocblas-install/lib/../../Tensile/library/TensileLibrary.dat

rocBLAS error: Could not initialize Tensile library

It seems like the caller doesn't check for errors in the metric selection. Any suggestions on how to fix this?

zaliu (changes requested):
Logistically, please confirm that this PR can build and pass rocblas-test fully on at least gfx908 and gfx90a.  Thanks.

benjaminulmer:
> thanks @benjaminulmer . I have a question about the error message returned when the metric specified by TENSILE_METRIC is not found:
> 
> rocBLAS error: Could not load /home/alex/rocBLAS/build/release/rocblas-install/lib/../../Tensile/library/TensileLibrary.dat
> 
> rocBLAS error: Could not initialize Tensile library
> 
> It seems like the caller doesn't check for errors in the metric selection. Any suggestions on how to fix this?

I guess there's two sides to this. On the [rocBLAS side](https://github.com/ROCmSoftwarePlatform/rocBLAS-internal/blob/433ef8dd1ce0539196c2de94dfe7ef41b6674f77/library/src/tensile_host.cpp#L668), we don't do any checks around loading the Tensile library besides success/failure. However, on the [Tensile side](https://github.com/ROCmSoftwarePlatform/Tensile/blob/75b9aefe5981d85d1df32ddcebf32dab52bfdabd/Tensile/Source/lib/source/Tensile.cpp#L54), we're not communicating any information about what caused the failure, so rocBLAS doesn't have any other option currently.

Step 1 would be to fix the Tensile side. I would need to look closer into the (de)serialization to see exactly how to get the error messages out, but I don't think it should be too difficult. We then would need to decide on how `LoadLibraryFile` should communicate these errors with the caller. Once the necessary changes are in place in Tensile, we can update rocBLAS to make use of them.

benjaminulmer:
> Logistically, please confirm that this PR can build and pass rocblas-test fully on at least gfx908 and gfx90a. Thanks.

Hi Alex.

Is there a more "official" way to do this other than running manually? Either way, I can see to this task.

zaliu:
> Is there a more "official" way to do this other than running manually?

On each of gfx908 and gfx90a, clone the develop branch of Tensile, then:
```
git remote add upstream https://github.com/ROCmSoftwarePlatform/Tensile.git
git fetch upstream pull/1448/head:pr-1448
git checkout pr-1448
```
Then build rocBLAS with the local Tensile repo.  Nothing "official" here, but pretty straightforward.  Thanks.

afanfa:
I've successfully tested this PR on gfx90a. I can successfully compile and run rocblas-bench and the new functionality works as expected.

benjaminulmer:
Sorry I got distracted by some things that came up. I just started the tests on 908 and 90a; will update when they're done. I don't expect any issues, however.

benjaminulmer:
rocblas-test passed fully on 908 and 90a

zaliu (approved):

zaliu:
Please squash-merge when you're ready.  Thanks.

PR #1453, opened 2021-12-10
Updating Tensile document

babakpst:


wbgilmartin (approved):
LGTM

tuning_docs/tensile_tuning.tex
```diff
@@ -355,10 +387,13 @@ \subsubsection{BenchmarkProblems}
...
 
+\textbf{BenchmarkForkParameters, JoinParameters, BenchmarkJoinParameters} are obsolete items in this section. However, they still should be in this section.
+
+
+\subsubsection{Library Logic}
 \label{sec:LibraryLogic}
-\subsubsection{LibraryLogic}
 
-The \emph{Library Logic} is a collction of one or more yaml files containing benchmarking output results (usually one per gfx architecture, per benchmark problem) created by Tensile and is the final product of the benchmarking and tuning process. It consists of kernel meta-data (called the solution), mappings to problems that it can solve and performance data at particular sizes.
+The Library Logic is a collction of one or more yaml files containing benchmarking output results (usually one per gfx architecture, per benchmark problem) created by Tensile and is the final product of the benchmarking and tuning process (dropped in 3\_LibraryLogic). It consists of kernel meta-data (called the solution), mappings to problems that it can solve and performance data at particular sizes. If we skip this section in the configuration file, the library logic file (and 3\_LibraryLogic dir) will not be generated.
```
benjaminulmer, line 396:
```suggestion
The Library Logic is a collection of one or more yaml files containing benchmarking output results (usually one per gfx architecture, per benchmark problem) created by Tensile and is the final product of the benchmarking and tuning process (dropped in 3\_LibraryLogic). It consists of a list of solutions, a list of problem sizes, and a mapping from each problem size to the fastest solution. If we skip this section in the configuration file, the library logic file (and 3\_LibraryLogic dir) will not be generated.
```

benjaminulmer: The information here wasn't really correct imo. I think this is more accurate to what a Library Logic file actually contains.

tuning_docs/tensile_tuning.tex
```diff
@@ -373,9 +408,16 @@ \subsubsection{LibraryLogic}
...
+This describes the target platform for which the final library is targeted and will generate a corresponding \hyperref[sec:logicFile]{logic file} for this specific platform and device specification. \newline
+
+\noindent
+There are three main parts to this and they are mostly self describing. The ArchitectureName specifies the graphics architecture and the number ``906'' is the instruction set ``ISA'' id. The schedule is the GPU codename. The Device list is the specific device ids. Other example target platform specifications and devices are listed in \hyperref[sec:appendixD]{Appendix D}. \newline
+
 
 \noindent \textbf{Generated Logic File Example:}
+
+\noindent
+Notice here that the top of this logic file contains the same information as described in the LibraryLogic section of the configuration file. The content of this file will be discussed in section \ref{sec:logicFile} \newline
```
benjaminulmer, line 420:
```suggestion
Notice that the top of this logic file contains the same information as described in the LibraryLogic section of the configuration file. The content of this file will be discussed in section \ref{sec:logicFile} \newline
```

tuning_docs/tensile_tuning.tex
```diff
@@ -390,7 +432,20 @@ \subsubsection{LibraryLogic}
 - [Device 66a0, Device 66a1, Device 66a7]
 \end{minted}
 
-There are three main parts to this and they are mostly self describing. The ArchitectureName specifies the graphics architecture and the number ``906'' is the instruction set ``ISA'' id. The schedule is the gpu codename. The Device list is the specific device ids. Other example target platform specifications and devices are listed in \hyperref[sec:appendixD]{Appendix D}.
+
+\subsubsection{LibraryClient}
+\label{sec:LibClient}
+
+\noindent
+As of now, we do not need to have this section for the tuning process.
```
benjaminulmer, line 440:
```suggestion
The LibraryClient section is optional. Including it in the tuning config will build the Tensile library from the library logic files after tuning finishes and then runs the fastest solution for each tuned size.
```

benjaminulmer: Might as well explain what it actually does.

tuning_docs/tensile_tuning.tex
```diff
@@ -495,8 +553,8 @@ \subsubsection{Build Artifacts}
 \subsubsection{The Tensile Logic File}
 \label{sec:logicFile}
 
-
-The results of the tuning generate a Tensile logic file. This file is used to generate the final Tensile library using the TensileCreateLibrary application. It contains the library and kernel specification as well as the size mappings. An example:
+\noindent
+The results of the tuning generate a Tensile logic file (or multiple files) in build*/3\_LibraryLogic. This file is used to generate the final Tensile library using the TensileCreateLibrary application. This is a separate application (still in the Tensile repo) and the library logic file generated by the Tensile is the input file to TensileCreateLibrary application. This application will be discussed in section \ref{sec:TensileLib}. The logic file contains the library and kernel specification as well as the size mappings. Here is a sample logic file:
```
benjaminulmer, line 557:
```suggestion
The result of tuning is the Tensile library logic file(s) in build*/3\_LibraryLogic. These file are used to generate the final Tensile library using the TensileCreateLibrary application. This is a separate application (still in the Tensile repo) which takes the library logic files generated by Tensile as input. This application will be discussed further in section \ref{sec:TensileLib}. Here is a sample logic file:
```

benjaminulmer: Cleaning up a bit and removing redundant information.

tuning_docs/tensile_tuning.tex
```diff
@@ -579,61 +637,97 @@ \subsubsection{The Tensile Logic File}
 	- [kernel index, measured efficiency].
 \end{center}
 
-\section{Validating Using rocblas-bench}
+\noindent
+Note that as of now, rocBLAS does not hold the ld information in the final library, even if we define them in the configure file.
```
benjaminulmer, line 641:
```suggestion
Note that as of now, rocBLAS does not store the leading dimension information in the final library, even if we define them in the configure file.
```

tuning_docs/tensile_tuning.tex
```diff
@@ -579,61 +637,97 @@ \subsubsection{The Tensile Logic File}
...
+	$\langle architecture\_name \rangle \_\langle Contraction\_Type \rangle\_ \langle Data\_Type \rangle$
+\end{center}
+
+\noindent
+$\langle architecture\_name \rangle$ refers the GPU architecture, for example for Vega 20, the architecture name is vega20. $\langle Contraction\_Type \rangle$ refers to the Einstein tensor convention as discussed in section \ref{sec:einstein}. $\langle Data\_Type \rangle$ is based on the ProblemType in the configuration file. For example, SB refers to a Batched Single precision DataType. HSS\_BH means Half-precision for DataType (A/B), Single-precision for DestDataType (C/D), and Single-precision for ComputeDataType. Notice also that the exact same naming convention will be used in the rocBLAS library.
+
+\section{Integrating new libraries in rocBLAS}
+Once we have library logic files for new sizes, we need to merge and update rocBLAS libraries. This a two-fold process: 
+\begin{itemize}
+\item First, we need to confirm that the new logic file is generating the correct results and the performance of the new libraries is more that the old library. We use rocBLAS-bench tool for this part. See the next section.
```
benjaminulmer, line 661:
```suggestion
\item First, we need to confirm that the new logic file is generating the correct results and the performance of the new library is better than the old library. We use rocBLAS-bench tool for this part. See the next section. 
```

tuning_docs/tensile_tuning.tex
```diff
@@ -579,61 +637,97 @@ \subsubsection{The Tensile Logic File}
...
 The git repository for rocBLAS is found at \url{https://github.com/ROCmSoftwarePlatform/rocBLAS}. Either the develop or master branch version may be used for testing.
 
 \begin{lstlisting}[language=bash]
-git clone -b master|develop https://github.com/ROCmSoftwarePlatform/rocBLAS.git
+$ git clone -b master|develop https://github.com/ROCmSoftwarePlatform/rocBLAS.git
 \end{lstlisting}
 
-\noindent
-The library can be installed using the included installation script:
+To use rocBLAS-bench we need to install it locally. The library can be installed using the included installation script (the installation may last quite some time.):
```
benjaminulmer, line 681: Not strictly true; the rocblas-bench (and other clients) can be built using the installed version of rocBLAS without having to first rebuild rocBLAS. I think the option is --clients-only.

tuning_docs/tensile_tuning.tex
```diff
@@ -579,61 +637,97 @@ \subsubsection{The Tensile Logic File}
...
-cd rocBLAS
-./install.sh -d -c
+$ cd rocBLAS
+$ ./install.sh -h
+$ ./install.sh -d -c -a <architecture> --cmake_install
 \end{lstlisting}
 
 \noindent
-where the c option builds the client applications, which includes rocblas-bench, and the d option will build and install the dependencies (the d option only need to be included once on each system that rocblas is being build).
+where the c option builds the client applications, which includes rocBLAS-bench, and the d option will build and install the dependencies (the d option only need to be included once on each system that rocBLAS is being build). -a refers to the GPU architecture. For example: gfx906:xnack-, gfx908:xnack-. Without this flag, rocBLAS installs/tunes for all architectures, and installation lasts even longer. To skip rocBLAS Tensile use -n flag (You cannot benchmark GEMM then). The cmake flag allows the rocBLAS to install cmake if it is older than required.
```
benjaminulmer, line 690:
```suggestion
where the c option builds the client applications, which includes rocBLAS-bench, the d option builds and installs the dependencies (only needed to be included once on each system that rocBLAS is being built), and the a option specifies the GPU architecture e.g. gfx906:xnack-, gfx908:xnack-. Without this flag, rocBLAS installs for all architectures, and installation lasts even longer. The cmake flag allows the rocBLAS to install cmake if it is older than required.
```

benjaminulmer: Cleaning up a bit. Removed the bit about the -n flag; I think this just adds confusion since this is a Tensile doc and you would only use rocBLAS if it was built with Tensile.

tuning_docs/tensile_tuning.tex
```diff
@@ -579,61 +637,97 @@ \subsubsection{The Tensile Logic File}
...
 gobble=0,breaklines,
 frame=single,
 linenos
 ]{yaml}
 
-- { rocblas_function: "rocblas_sgemm", transA: 'N', transB: 'N', M: 1024, N: 1264, K: 345, lda: 1024, ldb: 345, ldc: 1024, cold_iters: 2, iters: 10 }
-- { rocblas_function: "rocblas_sgemm", transA: 'N', transB: 'N', M: 1025, N: 1064, K: 148, lda: 1025, ldb: 148, ldc: 1025, cold_iters: 2, iters: 10 }
-- { rocblas_function: "rocblas_sgemm", transA: 'N', transB: 'N', M: 2450, N: 512, K: 1023, lda: 2450, ldb: 1023, ldc: 2450, cold_iters: 2, iters: 10 }
-- { rocblas_function: "rocblas_sgemm", transA: 'N', transB: 'N', M: 2345, N: 2222, K: 345, lda: 2345, ldb: 345, ldc: 2345, cold_iters: 2, iters: 10 }
+- { rocBLAS_function: "rocBLAS_sgemm", transA: 'N', transB: 'N', M: 1024, N: 1264, K: 345, lda: 1024, ldb: 345, ldc: 1024, cold_iters: 2, iters: 10 }
```
benjaminulmer, line 711: I'm not sure if these are case sensitive or not. Could you double check that?

tuning_docs/tensile_tuning.tex
```diff
@@ -579,61 +637,97 @@ \subsubsection{The Tensile Logic File}
...
 
 \end{minted}
 
 \noindent
-This can be executed in rocblas-bench as follows:
+This can be executed in rocBLAS-bench as follows:
 
 \begin{lstlisting}[language=bash]
-./rocblas-bench --yaml problem-sizes.yaml
+./rocBLAS-bench --yaml problem-sizes.yaml
```
benjaminulmer, line 722: This definitely _is_ case sensitive. I'm guessing a find and replace got out of hand?

tuning_docs/tensile_tuning.tex
```diff
@@ -817,8 +911,9 @@ \subsection{Code Object Libraries}
...
+The Tensile library's independence from rocBLAS enables us to build and use the libraries as part of a separate workflow. This avoids much of the long wait times related to the rocBLAS build. The rocBLAS-bench application can reference a library by setting the environment variable rocBLAS\_TENSILE\_LIBPATH. When this variable is set to a specific version of the Tensile library, rocBLAS will load that version at runtime; otherwise, the prebuilt version of the library will be loaded. This means that only one version of rocBLAS needs to be built for testing when comparing the results of a new version of the Tensile library with a reference version. When using this method of referencing a library, it is important that the same version of TensileCreateLibrary from the rocBLAS build is used to create the library; otherwise, there could be conflicts.
```
benjaminulmer, line 916:
```suggestion
The Tensile library's independence from rocBLAS enables us to build and use the libraries as part of a separate workflow. This avoids much of the long wait times related to the rocBLAS build. The rocBLAS-bench application can reference a library by setting the environment variable ROCBLAS\_TENSILE\_LIBPATH. When this variable is set to a specific version of the Tensile library, rocBLAS will load that version at runtime; otherwise, the prebuilt version of the library will be loaded. This means that only one version of rocBLAS needs to be built for testing when comparing the results of a new version of the Tensile library with a reference version. When using this method of referencing a library, it is important that the same version of TensileCreateLibrary from the rocBLAS build is used to create the library; otherwise, there could be conflicts.
```

benjaminulmer: More find and replace.

tuning_docs/tensile_tuning.tex
```diff
@@ -846,10 +941,10 @@ \subsection{Using the Tensile Library For Testing}
 
 \subsubsection{Example Usage}
 
-Once the library is generated, it can be used in rocBLAS by setting the environment variable ROCBLAS\_TENSILE\_LIBPATH. Execute rocblas-bench as follows:
+Once the library is generated, it can be used in rocBLAS by setting the environment variable rocBLAS\_TENSILE\_LIBPATH. Execute rocBLAS-bench as follows:
```
benjaminulmer, line 944:
```suggestion
Once the library is generated, it can be used in rocBLAS by setting the environment variable ROCBLAS\_TENSILE\_LIBPATH. Execute rocBLAS-bench as follows:
```

tuning_docs/tensile_tuning.tex
```diff
@@ -846,10 +941,10 @@ \subsection{Using the Tensile Library For Testing}
...
 
 \begin{lstlisting}[language=bash]
-ROCBLAS_TENSILE_LIBPATH=path/to/drop/new/tensile/library ./rocblas-bench --yaml problem_sizes.yaml
+rocBLAS_TENSILE_LIBPATH=path/to/drop/new/tensile/library ./rocBLAS-bench --yaml problem_sizes.yaml
```
benjaminulmer, line 947:
```suggestion
ROCBLAS_TENSILE_LIBPATH=path/to/drop/new/tensile/library ./rocblas-bench --yaml problem_sizes.yaml
```

tuning_docs/tensile_tuning.tex
```diff
@@ -931,28 +1029,28 @@ \subsection{Automation Workflow}
...
+rocBLAS logging provides information about the rocBLAS functions which get called during an application as well as the size information of the calls. This logging is enabled by setting the rocBLAS\_LAYER environment variable. The two settings that get used as input into the tuning automation are rocBLAS\_LAYER=2 and rocBLAS\_LAYER=4; these each produce output about the rocBLAS calls in different formats.
 
 \begin{lstlisting}[language=bash,breaklines=true]
-ROCBLAS_LAYER=2
+rocBLAS_LAYER=2
 
 # generates output in the form of
-rocblas-bench -f gemm -r s --transposeA N --transposeB N -m 1024 -n 1024 -k 1024 --beta 0.0 --alpha 1.0 --lda 1024 --ldb 1024 --ldc 1024
+rocBLAS-bench -f gemm -r s --transposeA N --transposeB N -m 1024 -n 1024 -k 1024 --beta 0.0 --alpha 1.0 --lda 1024 --ldb 1024 --ldc 1024
```
benjaminulmer, line 1038: More find and replace here; I'm too lazy to tag them all

tuning_docs/tensile_tuning.tex
```diff
@@ -962,7 +1060,7 @@ \subsubsection{The provision\_tuning.sh Script}
 
 
 \begin{lstlisting}[language=bash,breaklines=true]
-./tuning/scripts/provision_tuning.sh tensile_tuning logs/inception-rocblas-configs_unique.log tf_inception.yaml vega20
+./tuning/scripts/provision_tuning.sh tensile_tuning logs/inception-rocBLAS-configs_unique.log tf_inception.yaml vega20
```
benjaminulmer, line 1063:
```suggestion
./tuning/scripts/provision_tuning.sh tensile_tuning logs/inception-rocblas-configs_unique.log tf_inception.yaml vega20
```

benjaminulmer (changes requested):
I put most of the stuff I would liked changed as suggest changes. There are quite a few place where your find and replace on "rocblas" got out of hand; I tagged some of them, but there are more.

babakpst:
Thank @benjaminulmer for your detailed comments and for catching the errors. I implemented all your comments/suggestions and fixed some other typos.

benjaminulmer (approved):
Good work and thanks for addressing all my comments.
LGTM

PR #1460, opened 2022-01-11
Initial changes to support two-tier solution selection

benjaminulmer:
### The two-tier solution selection proposal
Currently, the solution selection logic for handling exact size matches and the selection logic for non-exact matches (sometimes called known and unknown sizes, respectively) are _not_ independent. That is, adding tuning results for a new size changes the distance matching table and consequently changes how unknown sizes get mapped to solutions. This interdependence means special workarounds are necessary if a kernel should _only_ be used for the exact size match and not participate in general solution selection for unknown sizes. Furthermore, recent experiments show that we can get better performance, on average, for unknown sizes by using different distance tables for unknown sizes or other methods all together.

The proposed solution to both of these issues is to have two tiers of solutions selection: primary and fallback. The primary tier will handle exact size matches and operates almost identically to the current distance based solution selection---the only difference being it will only return exact matches. The fallback tier handles these non-exact matches and will use a selection method optimized to perform well on the full range of possible sizes.

### What's in this PR
This PR contains the initial changes necessary for the two-tier solution selection proposal. Specifically, it allows for using a two-tier solution selection method in place of what would previously be a single `MatchingLibrary`; however, it does this in a backwards compatible way where a single `MatchingLibrary` is still possible (and is the default behaviour if logic files are unchanged).

What this PR _doesn't_ contain is any new implementation for the fallback selection method. Currently, any fallback selection method must simply be a 'MatchingLibrary', potentially with a different distance metric.

### Design
The bottom of the solution selection hierarchy, which was previously a single `MatchingLibrary`, is now a `PredicateLibrary` containing one or more `MatchingLibrary` objects. For the default behaviour, there is a single `MatchingLibrary` using the default Euclidean distance and a true predicate. For the two-tier behaviour, the primary tier is a `MatchingLibrary` with the new `Equality` distance and a new `EqualityMatching` predicate. This new predicate is functionally a true predicate, but has to be a distinct type to ensure the rows are ordered correctly when merging libraries (specifically, we need the primary tier to be before the fallback one).

On the Tensile library side, the primary tier is implemented as a template specialization of `DistanceMatchingTable` for the case where the distance metric is the new `Equality` distance; a common base class allows minimal code duplication between the generic version and the specialization. The only difference in behaviour between the two is for `findBestKeyMatch`, where the specialization will return null if there isn't an exact size match in the library. This null return value prompts the solution selection algorithm to continue exploring the hierarchy, with the next search point being the next row in the `PredicateLibrary`: the fallback tier.

I have done initial testing but it was far from exhaustive. Assuming CI looks good on Tensile, I will run tests with rocBLAS to make sure the default behavior isn't getting impacted by these changes.


Tensile/Source/lib/include/Tensile/PropertyMatching.hpp
```diff
@@ -116,45 +116,235 @@ namespace Tensile
...
+            virtual std::tuple<ReturnValue, double> findBestKeyMatch(Key const& key,
+                                                                     Transform transform) const = 0;
+
+            virtual std::tuple<ReturnValue, double>
+                findBestMatch(Object const& object, Transform transform) const override
+            {
+                return findBestKeyMatch(keyForProblem(object), transform);
+            }
+
+            virtual ReturnValue findBestEvaluationSolution(Object const&   object,
```
wbgilmartin, line 166: This interface was designed to provide a run time switch, using an environment variable, to enable the experimental TAM selection method.  For future work we need to understand how to enable a clean way to select at runtime between two or more solutions methods. Does the tooling here provide such a mechanism?

wbgilmartin (approved):
LGTM. This achieves the stated objectives and everything passes.

afanfa:
The PR looks good, I only have one question related to the impact of this on the TENSILE_SOLUTION_INDEX environment variable used to select a specific solution by providing the index number. Would this functionality still work when this PR gets merged?

Tensile/Source/lib/include/Tensile/PropertyMatching.hpp
benjaminulmer: I don't think this PR adds any functionality that will be useful for changing the selection at runtime, at least not directly; all the changes are for build-time steps. I tried to make sure the changes I made wouldn't affect the TAM method, but I still need to verify it still works as expected. I will do this before I merge.

I know @afanfa also expressed desire for switching parts of solution selection at runtime, so I agree it's worth coming up with a plan to do this nicely and in a way that doesn't significantly impact selection speed.

benjaminulmer:
> The PR looks good, I only have one question related to the impact of this on the TENSILE_SOLUTION_INDEX environment variable used to select a specific solution by providing the index number. Would this functionality still work when this PR gets merged?

I don't expect this PR will affect that functionality, but I will confirm before I merge.

zaliu (approved):
It would be great if a full rocBLAS build and test on one recent gfx architecture with this PR can be done before you merge.  Thanks.

afanfa (approved):
LGTM

benjaminulmer:
Yes I'm planning on doing some more manual verification before merge, including a rocblas test run. This PR does _not_ need to be merged before the 5.1 FC; would you like me to close it in the meantime while I finalize the testing?

benjaminulmer:
Everything looks good to me; I'm going to go ahead and merge
- rocBLAS nightly test suite passed fully on gfx90a
- I've manually verified that TENSILE_TAM_SELECTION_ENABLE still works with these changes
- The TENSILE_SOLUTION_INDEX variable is used higher up in the solution selection hierarchy than these changes affect, so that should still work as before

PR #1463, opened 2022-01-17
ROCm 5.1 merge master into develop

zaliu:


amcamd (approved):
Confirmed that version changed from 31 to 32 in CMakeLists.txt, __init__.py, and TensileConfigVersion.cmake files

PR #1464, opened 2022-01-19
Add TensileUpdateLibrary for updating old library logic files

benjaminulmer:
As Tensile is updated with new solution parameters, solutions from old library logic files will not have values for said new parameters. On the surface this is not an issue, as the missing parameters will be assigned default values when being parsed by Tensile. However, these missing parameters _can_ lead to issues with the merge utility script, where solutions are treated as different when they are actually the same due to missing parameters in one of the solutions. As a result, duplicate solutions can be present after a merge. 

Thus far, these duplicate solutions have not been an issue, as they get identified as duplicates and handled properly during TensileCreateLibrary. Recently though, this did cause issues for a rocBLAS tuning PR that had to update the Fp16AltImpl files. To solve the problem, I wrote a simple program to update the solutions in a given library logic file to have all the most current solution parameters. After cleaning it up to a point that makes it generally useful to others, I figured we might as well put it into mainline Tensile. 

By leveraging existing Tensile functions, the program is quite simple. In short, it parses the problem type and solution objects from the library logic file, then turns these objects back into their yaml representations and updates the file.

### DIFF
--- a/Tensile/LibraryIO.py
+++ b/Tensile/LibraryIO.py
@@ -250,9 +250,6 @@ def createLibraryLogic(schedulePrefix, architectureName, deviceNames, logicTuple
                 solutionState["ProblemType"]["DestDataType"].value
         solutionState["ProblemType"]["ComputeDataType"] = \
                 solutionState["ProblemType"]["ComputeDataType"].value
-
-        solutionState["ISA"] = list(solutionState["ISA"])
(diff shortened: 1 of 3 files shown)


zaliu (approved):
LGTM.  Thanks for polishing the script and making it available.  Please squash-merge.

PR #1465, opened 2022-01-19
Let TensileRetuneLibrary use sizes from separate file

benjaminulmer:
Just a small update to the TensileRetuneLibrary tool that lets you specify a separate file containing sizes instead of using the ones present in the library logic file.

### DIFF
--- a/Tensile/TensileRetuneLibrary.py
+++ b/Tensile/TensileRetuneLibrary.py
@@ -130,7 +130,10 @@ def TensileRetuneLibrary(userArgs):
     argParser.add_argument("OutputPath", type=os.path.realpath,
                            help="Where to run benchmarks and output results")
     argParser.add_argument("SizeFile", type=os.path.realpath, nargs="?",
-                           help="Yaml file with sizes to tune", default=None)
+                           help="Yaml file with sizes to tune; same format as the 'ProblemSizes' "
+                           "section of a regular Tensile config "
+                           "(https://github.com/ROCmSoftwarePlatform/Tensile/wiki/Benchmark-Protocol)",
+                           default=None)
     argParser.add_argument("--update-method", "-u", dest="updateMethod",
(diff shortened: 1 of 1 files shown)


Tensile/TensileRetuneLibrary.py
```diff
@@ -50,13 +50,16 @@ def parseCurrentLibrary(libPath, skipRK):
...
-        if skipRK:
-            sol = solutions[mapping[0]]
-            if sol["ReplacementKernel"]:
-                continue
+    if sizePath is None:
+        for (size, mapping) in exactLogic:
+            if skipRK:
+                sol = solutions[mapping[0]]
+                if sol["ReplacementKernel"]:
+                    continue
```
JeremyAdamHart, line 59: This isn't related to your change, but should skipRK also be skipping custom kernels? The isCustomKernelConfig function could be used to test for that as well.

JeremyAdamHart (commented):
This looks like a handy feature. I'd considering adding a small blurb on how the size yaml file should be structured, as far as I can infer, it's just a list of the sizes without the solutions index or efficiencies, but it'd nice to have it mentioned explicitly

Tensile/TensileRetuneLibrary.py
benjaminulmer: The option only exists because I needed it for a specific use case, but if we need to filter out custom kernels in the future for whatever reason this is good to know about.

benjaminulmer:
> This looks like a handy feature. I'd considering adding a small blurb on how the size yaml file should be structured, as far as I can infer, it's just a list of the sizes without the solutions index or efficiencies, but it'd be nice to have it mentioned explicitly

Good call. I expanded the help message to explain the format for the file.

JeremyAdamHart (approved):
LGTM

PR #1468, opened 2022-01-21
Fp16 bug HBH

babakpst:
Keep the alpha type as fp32 for HBH case.


zaliu:
Offline testing of this PR with rocBLAS-internal [#1054](https://github.com/ROCmSoftwarePlatform/rocBLAS-internal/pull/1054) passed rocblas-test fully on gfx906, gfx908 and gfx90a.  It is ready for review.  Thanks.

Tensile/Components/Signature.py
```diff
@@ -71,7 +71,11 @@ def getDstValueType(kernel, cov):
     return dstValueType
 
 def getCptValueType(kernel, cov):
-    cptValueType = cptValueTypeDict[kernel["ProblemType"]["DataType"].toNameAbbrev()]
+    if kernel["ProblemType"]["DataType"].isHalf() and kernel["ProblemType"]["HighPrecisionAccumulate"]:
+        cptValueType = cptValueTypeDict[kernel["ProblemType"]["ComputeDataType"].toNameAbbrev()] # for HHS case, this will be handled in HBH kernels.
+    else:
+        cptValueType = cptValueTypeDict[kernel["ProblemType"]["DataType"].toNameAbbrev()]
```
cgmillette, line 77: Assuming this function is getting the compute data type as the name implies, why is this not always

cptValueType = cptValueTypeDict[kernel["ProblemType"]["ComputeDataType"].toNameAbbrev()]

?

Tensile/SolutionStructs.py
```diff
@@ -968,6 +973,7 @@ def checkIfSupportedGEMMType(self):
 
     # TODO- Migrate ([H/H/H]+HPA) to ([H/H/S]+HPA)
     # Note that we need to do a little change in rocBLAS and logic yaml
+    # Currently, ([H/H/S]+HPA) is implemented, but due to some kenerl naming conflict, we use HBH kernels instead HHS_BH.
```
cgmillette, line 976: Small typo 'kenerl' -> 'kernel'

Tensile/Components/Signature.py
babakpst: Good point. It confused me while I was working on it, too. cptValueType was based on DataType, before my commit, and, initially, I modified the def to define the cptValueType based on computeDataType, (as the name suggests). But I faced some errors later on for other types.

cgmillette: At least previously I would assume the compute type would follow OutputDataType from legacy code.

Can anyone else weigh in on this?

benjaminulmer: OK I think I finally understand what's going on here, as I initially had the same thought as Chris.
It looks like this code is from before we made compute data type explicit in Tensile. So instead of just pulling the compute data type directly, this function uses a dict to map to the correct type based on the (input) data type. Before this fix this wasn't a problem, but now we need to map F16 to either F16 or F32 depending on if it's HPA or not.

The new code produces the correct result (returning F32 for F16+HPA), but the way it does it is quite confusing. Short of refactoring this entire file, I would suggest bypassing the dict entirely and just setting cptValueType directly
```c++
if kernel["ProblemType"]["DataType"].isHalf() and kernel["ProblemType"]["HighPrecisionAccumulate"]:
    cptValueType = "F32"
else:
    cptValueType = cptValueTypeDict[kernel["ProblemType"]["DataType"].toNameAbbrev()]

File: Tensile/Common.py [hunk 3]
```diff
@@ -339,7 +339,7 @@ def getArchitectureName(gfxName):
 
 # These type are newly supported and we would like to use a better file naming for them: _TiToTc_
 # For the rest of the typed, we keep them with old existing naming.
-typesUsingNewNaming = [ ('H','H','S'), ('H','S','S'), ('B','S','S'),('I8','I','I')]
+typesUsingNewNaming = [ ('H','S','S'), ('B','S','S'),('I8','I','I')]   # ('H','H','S') is removed bcs we are merging this case in HBH
```
benjaminulmer, line 342: "merging" implies to me that the old HBH functionality still exists. Something like "...bcs it is replacing the legacy HBH type"

Tensile/SolutionStructs.py
benjaminulmer, line 976: I think this PR accomplished the TODO, so you can just remove this whole block of comments (including the new one you added)

Tensile/SolutionStructs.py
```diff
@@ -850,10 +850,15 @@ def __init__(self, config):
       else:
         if "DataType" in config:
           self["ComputeDataType"] = DataType(config["DataType"])
+          self["DestDataType"] = DataType(config["DataType"])
         else:
           printExit("NO compute data type, or dest data type, or data type specified")
           self["DataType"] = DataType(0)
 
+    # Modifying ComputeDataType for HHS_BH: if (HHH+HPA), convert it to HHS_BH by setting ComputeDataType to s.
+    if self["ComputeDataType"].isHalf() and DataType(config["DataType"]).isHalf() and self["HighPrecisionAccumulate"]:
```
benjaminulmer, line 859: What do we do in the case of BBB+HPA? Do we reject it, or silently change the compute type like this? Since we're updating all the logic files in rocBLAS, I think rejecting might be better. At the very least, I think a warning message would be good.

benjaminulmer (changes requested):
My understanding of the history around this change:
1) For whatever legacy reason, F16+HPA (HBH under the old naming scheme) in Tensile used F16 as the compute type, which resulted in alpha/beta being cast from F16 to F32 
2) Proper HHS support was added to Tensile, but the original functionality for HBH was maintained for compatibility with rocBLAS until we could update it
3) Nothing happened for a long time, and now this PR and the corresponding rocBLAS one are meant to finally have rocBLAS use the proper HHS_BH kernels for F16+HPA, where alpha/beta are _not_ cast from F16

My understanding of what this PR does:
- Removes support for HHH + HPA kernels; if you try to request HHH + HPA, it will update it to HHS + HPA
- Changes the naming so that HHS + HPA uses the 'old' style naming instead of the 'new' style (i.e. instead of HHS_BH it is just HBH)
  - I'm assuming this is either needed for some sort of backward compatibility or done for consistency with how BF16 kernels are named

I've pointed out a few places where I think the comments are a bit misleading, or at least were confusing to me at first. I would also suggest giving this PR (and resulting commit message when merged) a slightly more descriptive name; perhaps something like "Remove legacy HHH+HPA kernels and change HHS_BH to be called HBH"

Please correct me if I got anything wrong.

cgmillette (changes requested):
If this is indeed a case where HHH + HPA is being replaced by HHS, then I strongly urge the rethinking using HBH to cram this in.

The newer naming convention <Ti\To\Tc> is in place such that kernel names are more descriptive. I am not in favor of regressing to older naming conventions, and would rather move to use HHS in naming.

cgmillette (approved):
Considering the scope of the current work, as discussed with @zaliu there is some additional risk with making bigger moves on the naming conventions. These extra things will deserve tracking in separate work items:

- First for the cptValue type awkwardness in Signature.py
- Second for managing shift from HBH to HHS kernel naming as applicable.

Please also address @benjaminulmer 's concerns and await his approval

Tensile/SolutionStructs.py
benjaminulmer: Looks like BBB isn't supported regardless of HPA or not, so modelling the behaviour off of that case isn't an option. I'd be OK with either a printExit or printWarning here depending on what you and/or @zaliu thinks is more appropriate

Tensile/Common.py
babakpst: HBH functionality still exists and this PR fixes its behavior, by not converting the datatype of alpha/beta to fp16.  The proper name for this case is HHS_BH, but due to some naming conflict (as I was told by Chris and Alex), we need to stick to HBH for this case. So, indeed we are merging the HHS_BH in HBH.

Tensile/SolutionStructs.py
babakpst: I don't think we reject BBB+HPA this. I guess we treat it the same way (but I am not sure about it).

Tensile/SolutionStructs.py
babakpst: This PR did not fully accomplish the TODO, as we eventually need to move HBH to HHS_BH to have a consistent naming convention. This has been mentioned in Chirs' comment down there (second bullet).

babakpst:
@zaliu- On @cgmillette remarks: I already talked to Alex about the two concerns you mentioned. For the scope of this task, we decided not to move HBH to HHS_BH. I would be glad to continue working on the two concerns.

babakpst:
@benjaminulmer On your comments:
3-..... This PR and the corresponding rocBLAS one are NOT meant to finally have rocBLAS use the proper HHS_BH kernels, it fixes the wrong behavior of HBH, by not converting alpha/beta to F16 and back to F32. rocBLAS logic files are still HBH, and no HHS_BH will be added until we fully migrate HBH to HHS_BH.

Removes support for HHH + HPA kernels .... . The PR does not remove support for HBH.

Tensile/Components/Signature.py
```diff
@@ -71,7 +71,11 @@ def getDstValueType(kernel, cov):
     return dstValueType
 
 def getCptValueType(kernel, cov):
-    cptValueType = cptValueTypeDict[kernel["ProblemType"]["DataType"].toNameAbbrev()]
+    if kernel["ProblemType"]["DataType"].isHalf() and kernel["ProblemType"]["HighPrecisionAccumulate"]:
+        cptValueType = cptValueTypeDict[kernel["ProblemType"]["ComputeDataType"].toNameAbbrev()] # for HHS case, this will be handled in HBH kernels.
```
benjaminulmer, line 75:
```suggestion
        cptValueType = "F32"
```

Tensile/SolutionStructs.py
```diff
@@ -850,10 +850,15 @@ def __init__(self, config):
...
+      self["ComputeDataType"] = DataType('s')
```
benjaminulmer, line 860:
```suggestion
      print2("DataType == f16 and HPA == True; setting compute data type to f32")
      self["ComputeDataType"] = DataType('s')
```

bump-version.sh
```diff
@@ -4,7 +4,7 @@
 # Version will be bumped in Tensile/__init__.py and in .yaml files
 
 OLD_VERSION="4.31.0"
```
benjaminulmer, line 6: Does this need to be updated to "4.32.0" @zaliu?

benjaminulmer (approved):
Just the one comment with regards to the version bump. Once that's confirmed it LGTM

zaliu (approved):
Please squash-merge after @benjaminulmer OK'ed your latest changes at his request.  Thanks.

bump-version.sh
zaliu: If we actually used the script, yes, we would change 4.31.0 to 4.32.0.  Otherwise the next feature complete merge will move 4.32.1 to OLD and 4.31.0 would be gone.  That said, it's a good point.

PR #1471, opened 2022-02-14
Optimize MessagePackLoadLibraryFile

cgmb:
This change provides a 13% reduction in the time spent in MessagePackLoadLibraryFile. Most of this optimization comes from changing how the file is read. The construct

```cpp
std::vector<uint8_t> data((std::istreambuf_iterator<char>(in)),
                          std::istreambuf_iterator<char>());
```

is--from a performance perspective--the worst commonly used way to read a whole file into memory in C++ ([see benchmarks](https://github.com/Sqeaky/CppFileToStringExperiments/blob/4da20ab67767fa557c36575918ef90ef3295fe7f/README.md)). It is an order of magnitude slower than a call to fread. Simply replacing that call with a naive C function to read the whole file provides a ~10% overall speedup in MessagePackLoadLibraryFile.

As an additional optimization, the [msgpack streaming API](https://github.com/msgpack/msgpack-c/wiki/v2_0_cpp_unpacker#msgpack-controls-a-buffer) can be used to parse the data in chunks. Interleaving the file reads with parsing allows the OS prefetcher some time to cache the next chunk while the CPU does useful parsing work. I didn't carefully tune the chunk size, but 512 KiB worked well on my machine.

The use of the streaming API was a somewhat more complex optimization, and it only provided perhaps another 5% overall speedup, but it seemed worthwhile nonetheless. There remain enormous optimizations to be found within PointerMappingTraits and the file read/parse time will become a larger relative portion of the overall function runtime once the PointerMappingTraits portion is optimized.

## Benchmarking Process
I disabled the YAML tests with
```patch
diff --git a/HostLibraryTests/llvm/LibraryPerformance_test.cpp b/HostLibraryTests/llvm/LibraryPerformance_test.cpp
index 67a59d80..096ec33a 100644
--- a/HostLibraryTests/llvm/LibraryPerformance_test.cpp
+++ b/HostLibraryTests/llvm/LibraryPerformance_test.cpp
@@ -308,12 +308,12 @@ std::vector<LibraryPerformanceTest::ParamType> GetLibraries(std::string const& e
 std::vector<LibraryPerformanceTest::ParamType> GetParams()
 {
     std::vector<LibraryPerformanceTest::ParamType> rv;
-
+/*
 #ifdef TENSILE_YAML
     auto yamlParams = GetLibraries("yaml");
     rv.insert(rv.end(), yamlParams.begin(), yamlParams.end());
 #endif
-
+*/
 #ifdef TENSILE_MSGPACK
     auto datParams = GetLibraries("dat");
     rv.insert(rv.end(), datParams.begin(), datParams.end());
```
then executed the tests before and after my change with
```bash
for i in {0..10}; do
  after/TensileTests --gtest_filter='*LibraryPerformanceTest.Load*' >> after.txt
  before/TensileTests --gtest_filter='*LibraryPerformanceTest.Load*' >> before.txt
done
```

## Results

| Version | Mean Time (ms) | σ (ms)
| - | - | - |
| Before | 8958 | 204 |
| After | 7687 | 192 |

## Future Optimizations

File loading and unpacking was 41.8% of the time spent in `MessagePackLoadLibraryFile` before optimization (24.0% in the `std::vector` constructor doing the read, 17.8% in `msgpack::v2::unpack`). After optimization, it is 27.6%. The file IO should be the slowest part of the library load operation, so I don't think I can optimize it ~any further~ much further. At this point, 72.4% of remaining time is spent in `Tensile::Serialization::MappingTraits<Tensile::MasterSolutionLibrary<Tensile::ContractionProblem, Tensile::ContractionSolution>, Tensile::Serialization::MessagePackInput, Tensile::Serialization::EmptyContext>::mapping`. If you chase down what it's actually doing, nearly all time is spent allocating strings and hash table nodes. This could be significantly optimized by:

1. Using an arena allocator for the solution library data. Since nearly all time in the construction of the MasterSolutionLibrary is spent on allocation, I would expect significant speedups by replacing the general-purpose system allocator with something designed for purpose. There's much less bookkeeping required when you know all the objects being created are going to have the same lifetime.
2. Using a more efficient hashtable implementation. `std::unordered_map` is limited by its interface to a chained bucket hash table. [This benchmark](https://martin.ankerl.com/2019/04/01/hashmap-benchmarks-03-01-result-InsertHugeInt/) may help shop for a better implementation.
3. Using an enumeration for the hash table key rather than strings. The keys are all known, so there's no need to support arbitrary strings.

With changes to the data structures on disk and in memory, the loading of the library file could probably be made about as fast as fread (i.e., there remains an order of magnitude improvement available). However, given that `hipModuleLoad` will become a bottleneck long before that point, such drastic changes are probably not necessary.

## Why is this still a Draft?
~I'm cleaning up the error handling, and maybe seeing if I can get the same benefit using `ifstream::read` instead of `fread`.~

It is a draft no more! The performance of the `ifstream::read` version is roughly the same as the `fread` version in my trial. After 33 trials, the results were:

| Version | Mean Time (ms) | σ (ms)
| - | - | - |
| Original | 8997 | 207 |
| fread | 7861 | 179 |
| ifstream | 7860 | 188 |

### DIFF
--- a/Tensile/Source/lib/source/msgpack/MessagePack.cpp
+++ b/Tensile/Source/lib/source/msgpack/MessagePack.cpp
@@ -2,7 +2,7 @@
  *
  * MIT License
  *
- * Copyright 2020 Advanced Micro Devices, Inc.
+ * Copyright 2020-2022 Advanced Micro Devices, Inc.
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
@@ -28,7 +28,7 @@
 
 #include <Tensile/msgpack/Loading.hpp>
 
-#include <stdio.h>
+#include <fstream>
 
 namespace Tensile
 {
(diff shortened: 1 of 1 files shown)


benjaminulmer (approved):
Great work Cory; very thorough.

sdquiring (approved):
Looks good! Also appreciate your additional suggestions.

PR #1473, opened 2022-03-02
Fix: Segmentation fault when run i8 datatype with TENSILE_DB=0x80

jichangjichang:
Setting TENSILE_DB=0x80 would let tensile runtime try to print tensor buffer address of **(TYPE \*)ptr**.
If datatype is i8, it would try to print **(unsigned char \*) ptr** as string but not buffer address, then cause the crash.

### DIFF
--- a/Tensile/KernelWriter.py
+++ b/Tensile/KernelWriter.py
@@ -1095,11 +1095,11 @@ class KernelWriter(metaclass=abc.ABCMeta):
       ####
       if self.numVgprBuffer >= kernel["LoopIters"]:
         for vacancy in self.localReadsVacancy:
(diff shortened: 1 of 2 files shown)


rosenrodt (approved):
LGTM

jichangjichang (commented):
This make precheck_in test fail.
Please check Tensile/Tests/pre_checkin/no_load_loop/sgemm_nll_asm_nt.yaml

poyenc:
I've removed the only-change-comment commits

zaliu (approved):
LGTM.  Please squash-merge when ready.  Thanks.

PR #1475, opened 2022-03-07
Add Denorm testing for Assembly Kernels

yoichiyoshida:
- Adding denorm testing for assembly kernels
- Adding header to force assembly kernels to keep denorms and not flush to 0

### DIFF
--- a/Tensile/ClientWriter.py
+++ b/Tensile/ClientWriter.py
@@ -56,6 +56,7 @@ class DataInitName(Enum):
   NegOne = 17
   Max = 18
   DenormMin = 19
+  DenormMax = 20
 
 class ClientLogLevel(Enum):
   Error = 0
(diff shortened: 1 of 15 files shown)


cgmb:
> Adding header to force assembly kernels to keep denorms and not flush to 0

Does this mean that assembly kernels for rocBLAS are currently using FTZ by default?

yoichiyoshida:
> > Adding header to force assembly kernels to keep denorms and not flush to 0
> 
> Does this mean that assembly kernels for rocBLAS are currently using FTZ by default?

Before we did not explicitly set the flags, so depended on the compiler/assembler defaults for FTZ behaviour. In the current version, the assembler defaults are to not FTZ for 16/64 but FTZ for 32, for instructions that follow the flags (mostly FMA, but SGEMM MFMA as well). The compiler (for source kernels) was setting not FTZ for 32, differing with the assembler. With this new change we will force the setting to never flush for all assembler cases.

zaliu (approved):
All pipelines passed at least once.  Please squash-merge after @AlexBrownAMD reviewed and OKed this PR.  Thanks!

zaliu:
@bragadeesh Please decide if we want to let this PR in.  I recommend it.  Thanks.

bragadeesh (approved):

AlexBrownAMD (approved):
LG

PR #1477, opened 2022-03-08
Add option to write different architectures to different TensileLibrary files

JeremyAdamHart:
A large contributor to the time required to initialize Tensile is the parsing of the TensileLibrary.dat/yaml file. While only the code object files for the used architecture are loaded with hipModuleLoad, Tensile reads and processes all of the architectures found in the TensileLibrary file (unless only a single architecture is built).

To reduce the time to load the library file, I've added a separate-architectures option that writes each architecture into a different TensileLibrary_gfxXXX.dat file. This has been tested and works with TensileCreateLibrary and rocBLAS (pending some modifications to use the option).

Some of this functionality is similar to the PackageLibrary option which I notice hasn't been updated in a while and is listed for possible deprecation. When using the PackageLibrary option, TensileCreateLibrary fails to build. I've put my changes alongside this option, although some of the functionality is the same.

### DIFF
--- a/Tensile/TensileCreateLibrary.py
+++ b/Tensile/TensileCreateLibrary.py
@@ -1345,6 +1345,8 @@ def TensileCreateLibrary():
                           default=-1, help="Number of parallel jobs to launch.")
   argParser.add_argument("--verbose", "-v", dest="PrintLevel", type=int,
                           default=1, help="Set printout verbosity level.")
+  argParser.add_argument("--separate-architectures", dest="SeparateArchitectures", action="store_true", 
+                         default=False, help="Separates TensileLibrary file by architecture")
 
   argParser.add_argument("--global-parameters", nargs="+", type=splitExtraParameters, default=[])
   args = argParser.parse_args()
@@ -1360,6 +1362,7 @@ def TensileCreateLibrary():
   arguments["RuntimeLanguage"] = args.RuntimeLanguage
   arguments["CodeObjectVersion"] = args.CodeObjectVersion
   arguments["Architecture"] = args.Architecture
+  arguments["SeparateArchitectures"] = args.SeparateArchitectures
   arguments["CxxCompiler"] = args.CxxCompiler
   if args.CmakeCxxCompiler:
     os.environ["CMAKE_CXX_COMPILER"] = args.CmakeCxxCompiler
--- a/Tensile/TensileCreateLibrary.py
+++ b/Tensile/TensileCreateLibrary.py
@@ -1105,6 +1105,8 @@ def buildObjectFilePaths(prefixDir, solutionFiles, sourceKernelFiles, asmKernelF
   asmKernelPaths = []
   sourceLibPaths = []
   asmLibPaths = []
+  libMetadataPaths = []
+
 
   # Build full paths for source kernel files
   sourceKernelDir = ""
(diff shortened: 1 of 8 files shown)


cgmb:
Just to be clear, `gfxXXX` in `TensileLibrary_gfxXXX.dat` is just the processor and not the whole [target id](https://clang.llvm.org/docs/ClangOffloadBundler.html#target-id)?

JeremyAdamHart:
> To be clear, `gfxXXX` in `TensileLibrary_gfxXXX.dat` is just the processor and not the whole [target id](https://clang.llvm.org/docs/ClangOffloadBundler.html#target-id)?

That's correct, I just created libraries with "gfx908:xnack-;gfx90a:xnack_;gfx90a:xnack-" and got only TensileLibrary_gfx90a.datand TensileLibrary_gfx908.dat

benjaminulmer:
Haven't had a chance to look at the code in detail yet, but a few comments in the meantime

1) I did run a quick test to verify this works when we have multiple hardware rows in the solution library for a given architecture (e.g. 104CU vs not for aldebaran)
2) Is the logic for merging in the host library going to be a separate PR (I see some C++ code was removed before you made this one)? Do we even need that logic---is that a use case we need to support?
3) How similar is this change to the PackageLibrary option? If PackageLibrary doesn't work and this will more-or-less do the same thing, I'd prefer to just consolidate things so we don't have broken options

JeremyAdamHart:
@benjaminulmer 
2. The c++ library merging code (dynamic merging just to differentiate it from our static merging) seems like it's unnecessary in the case of rocBLAS, as rocBLAS initialize takes a device ID when it's initializing. Aside from that, it sounds like using multiple GPUs with different architectures in the same instance of Tensile isn't a use case that there's really a demand for. The dynamic merging code is working though and I've moved it out to be used in a future PR where it'll definitely be necessary (It's generic enough that it'd still support merging different architectures).
3. In my reading of the PackageLibrary option, it creates separate folders for each target and puts the .co/hsaco files in them (As well as the related TensileLibrary.yaml file). The separate-architectures option puts them all in the same directory as usual, and renames the indices so there's no conflicts. The latter's a little easier for rocBLAS to locate the relevant files, but not enough for that to be a determining factor. I can't see any changes related to PackageLibrary (and looking at the code I don't think it's worked for close to 2 years) so my view is that we should just chose whatever organization we prefer

benjaminulmer:
2. That makes sense
3. I see. In that case, my vote would be all in one directory (i.e. the way you did it in this PR). 
    - we already use the _gfxXXX convention for .co files
    - having the location of _gfxXXX.co files depend on the value of --separate-architectures seems like a bad idea 
    - the directory that contains all these files isn't meant to be a "human readable" directory, so adding a few extra files isn't an issue
 
Unless anyone else is opposed, I think we should just remove the package-library option. Maybe @sdquiring wants to weigh in?

Tensile/TensileCreateLibrary.py
```diff
@@ -1506,13 +1528,19 @@ def splitExtraParameters(par):
         masterFile = os.path.join(archPath, "TensileLibrary")
         newMasterLibrary.applyNaming(kernelMinNaming)
         LibraryIO.write(masterFile, Utils.state(newMasterLibrary), args.LibraryFormat)
+  elif globalParameters["SeparateArchitectures"]:
+    for archName, newMasterLibrary in masterLibraries.items():
+      if (archName in archs):
```
lringham, line 1533: Don't need parentheses here.

lringham (commented):
LGTM!

Tensile/TensileCreateLibrary.py
JeremyAdamHart: Thanks for catching that, I fixed that and removed a commented out line

benjaminulmer (approved):
Not a huge fan of "libMetaData" as a name, but that's nitpicky enough that I'm not going to block merging. This file is long overdue for a good refactor/clean anyways, so we can change it then (if we ever get to it).

PR #1492, opened 2022-03-30
Code generator refactors

AlexBrownAMD:
Moved MAC code into components and updated code with named text formatting
Moved MemoryInstruction and RegisterPool classes into separate files
Added new asm caps for F64 and VOP3 dot4
Removed some dead code and refactored some functions with duplicated code to combine them


nakajee:
Is this just for refactoring? Or, does this fix any issue?
Did you confirm generated code was same as before your change?

AlexBrownAMD:
> Is this just for refactoring? Or, does this fix any issue? Did you confirm generated code was same as before your change?

This is mostly refactoring. A minor functional change is that MAC code is now selected based on asm caps rather than architecture (with added caps for fma_f64 and VOP3 version of dot4), which could allow a code path to be used on additional architectures that support it.

nakajee:
> > Is this just for refactoring? Or, does this fix any issue? Did you confirm generated code was same as before your change?
> 
> This is mostly refactoring. A minor functional change is that MAC code is now selected based on asm caps rather than architecture (with added caps for fma_f64 and VOP3 version of dot4), which could allow a code path to be used on additional architectures that support it.

Thanks for your reply.
So far, I did not find any issue on your code change.
If you confirmed generated assembly code was same as before for these MAC scenarios, I am OK with this change.

AlexBrownAMD:
I haven't done a diff on generated code yet, but I will try some. If there are any changes to assembly code, it should just be in formatting.

AlexBrownAMD:
Generated asm code is functionally identical, some minor updates to whitespace and additional comments.

The CI error is due to a timeout, cannot reproduce any error locally. Looks like the last few builds on develop also timed out.

Tensile/KernelWriterAssembly.py
```diff
@@ -14144,21 +13428,20 @@ def MulMIoutAlphaToArch(self, kernel):
         # c.imag = a.real * b.imag + a.imag * b.real = a.real * b.imag + tmp2
         imod.addText("v_fma_f64 %s, %s, %s, %s%s" % (vgpr("ValuC+__placeholder__ +2",2), sgpr("Alpha+0",2), vgpr("ValuC+%u"%(srcIdx+accImOffset),2), vgpr(vtmp2,2), self.endLine))
         self.codeMulAlpha.itemList[destIdx] = imod
-        #self.vgprPool.checkIn(vtmp1)
-        #self.vgprPool.checkIn(vtmp2)
 
     return kStr
 
   # Perform 32-bit scalar mul and save u64 result in two SGPR
```
nakajee, line 13434: It is better to change the comment here to cover signed case.

yoichiyoshida (approved):

nakajee (approved):
Looks good

Issue #1494, opened 2022-04-01
`getKernel` should notify not found if no module


v01dXYZ:
In `HipSolutionAdapter.cpp`, `getKernel(hipFunction_t& rv, string name)` is used to search in the kernels/modules for the right function matching the provided name. In the corner case of no module, `getKernel` will return `hipSuccess`  instead of `hipErrorNotFound`.

https://github.com/ROCmSoftwarePlatform/Tensile/blob/cbdcbb9083035c85467276585ed743e8166298f5/Tensile/Source/lib/source/hip/HipSolutionAdapter.cpp#L168

v01dXYZ:
When using `tensile_client` without `code-object`, the `loadEmbeddedCodeObjects()`  method will be called. But since this object is empty, the m_modules attribute is left empty. So it's not even a rare case.

PR #1496, opened 2022-04-05
Return hipErrorNotFound if no modules loaded

benjaminulmer:
Addresses #1494 

Old behaviour: if there were no modules (e.g. if the code-object argument(s) were missing), `getKernel` would incorrectly return `hipSuccess` which would result in a segfault

New behaviour: `getKernel` correctly returns `hipErrorNotFound` in this case

### DIFF
--- a/Tensile/Source/lib/source/hip/HipSolutionAdapter.cpp
+++ b/Tensile/Source/lib/source/hip/HipSolutionAdapter.cpp
@@ -2,7 +2,7 @@
  *
  * MIT License
  *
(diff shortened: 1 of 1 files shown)


AlexBrownAMD (approved):
Looks good! Were any test cases silently failing before this change?

benjaminulmer:
> Looks good! Were any test cases silently failing before this change?

I'm not aware of any tests that this affects, but it's worth looking into.

cgmillette (approved):
LGTM! Makes sense for the fall-through on modules as well

JeremyAdamHart (approved):

PR #1497, opened 2022-04-06
Upgrades to Tensile BenchmarkProblems

benjaminulmer:
This PR has four main changes to the BenchmarkProblems process, two of which---3 and 4---are closely related.
1. Support for a new config format being used by the SolutionSelection workgroup
2. An additional way to specify solution parameters: ~~GroupForkParameters~~ ForkParameter Groups
3. ~~Ability to run BenchmarkProblems with no sizes---this changes the old behaviour of using default sizes if none are specified~~
4. Ability to use Solutions generated from a previous run with a new set of sizes

### 1: New Config Format
The main Tensile app now accepts a new config format in addition to the standard one, specified by using the `--alternate-format` option. The new format is actually two yaml files, with the size list separated out into its own file. The PR includes an example for each of these files.

For the "main" config file, it has a similar structure to the regular Tensile config files. The main difference is it only supports a single benchmark problem, with the fields that would normally be specific to a certain benchmark problem (e.g. BenchmarkCommonParameters and ForkParameters) instead being top level fields. Additionally, the "ProblemType" part of the config is labelled as such.

The "sizes" file is simply a list of exact sizes as they would normally appear in `BenchmarkFinalParameters: - ProblemSizes:` section of a config. This is also the same format as used by TensileRetuneLibrary when specifying a size file. 

The main motivation for the new format is being able to split the sizes part of the config into its own file (more on this in 3 & 4). In this use case, it doesn't make sense to have more than one benchmark problem, so a simplified base config is used alongside the size list. Additionally, the default config format is saddled with a bunch of legacy which can make it difficult parse and edit; the new format is easier (imo) to work with both manually and programatically. This new format is what the SolutionSelection workgroup will be using, but I would still consider it "experimental" and subject to some change.

### 2: ~~GroupForkParameters~~ ForkParameter Groups
edit: GroupForkParameters have been renamed to just "Groups" and appear inside "ForkParameters".

GroupForkParameters are a means to give more control over the list of solutions benchmarked by allowing only certain combinations of parameters to be used. The motivation and process is most clearly explained through an example:

Say you only want to benchmark solutions with a 64x64 macrotiling on vega20; specifically, you're interested in the combinations `(WorkGroup: [8, 8, 1], ThreadTile: [8, 8])` and `(WorkGroup: [16, 16, 1], ThreadTile: [4, 4])`. To achieve this normally, you would have to use ForkParameters:
```yaml
ForkParameters:
- WorkGroup:
  - [8, 8, 1]
  - [16, 16, 1]
- ThreadTile:
  - [8, 8]
  - [4, 4]
```
However, this results in four total combinations of WorkGroup and Threadtile: double the number of combinations we care about. Using GroupForkParameters, you can specify only the combinations of parameters you wish to benchmark
```yaml
GroupForkParameters:
- - WorkGroup: [8, 8, 1]
    ThreadTile: [8, 8]
  - WorkGroup: [16, 16, 1]
    ThreadTile: [4, 4] 
```
Even in this simple example this would reduce the total number of solutions benchmarked by (up to, depending on what parameter combinations are valid) a factor of two. More elaborate cases with more parameters in the group and/or more entries in the group can have an even larger factor compared to the equivalent ForkParameters expression.

Additionally, multiple GroupForkParameters lists can be used _and_ they can be used in combination with regular ForkParameters. For example:
```yaml
ForkParameters:
- LdcEqualsLdd: [1, 0]
- FractionalLoad: [0, 1]
- TransposeLDS: [1, 0]
GroupForkParameters:
- - WorkGroup: [8, 8, 1]
    ThreadTile: [8, 8]
  - WorkGroup: [16, 16, 1]
    ThreadTile: [4, 4] 
- - DepthU: 16
    WorkGroupMapping: 1
  - DepthU: 16
    WorkGroupMapping: 15
  - DepthU: 8
    WorkGroupMapping: 8
```
In this example, the total number of solutions (before removing invalids) would be (2 * 2 * 2) * (2 * 3) = 48. To get the same solution set with only ForkParameters would require (2 * 2 * 2 * 2 * 2  * 2 * 3) = 192. 

### 3 & 4: Reusing solutions with new list of sizes
For tuning exercises with a large set of solutions, actually generating the solutions can be quite time consuming. When rerunning Tensile, this time consuming work is always repeated, even if the config file being used has not been changed in a way that would result in a different set of solutions being benchmarked. 

This PR adds a caching mechanism where solution generation---which includes both the generation of the Solution objects _and_ the building of these solutions into code objects and a Tensile library---are only redone if the solution parameters in the config have changed. This allows the set of sizes and/or the global parameters of the config to be changed, and then Tensile can be rerun and new benchmarks generated using the already existing solutions and code objects.

Changed in this PR, if no sizes are present when running Tensile, it will still do the solution and library generation steps, but will no longer benchmark with a set of default sizes. This is to facilitate a workflow where the solution and library generation are done ahead of time, and then Tensile calls with different size lists are dispatched to different GPUs and/or machines to parallelize tuning without redoing work unnecessarily. 

There is also a `--no-cache` option to bypass this mechanism if you want to force redoing these steps, such as when testing changes to the code generator. 

### Main Feedback Areas
- name for GroupForkParameters is maybe not perfectly clear, but was the best I could come up with
- is removing the old behaviour of using default sizes when no size list is present OK?
- the cache mechanism is currently opt-out, but perhaps opt-in would be better?


babakpst:
@benjaminulmer 
I like the concept of GroupForkParameters. Can we have a fork within the GroupForkParameter? For instance, for the example that you have, can we rewrite it as follows: 
```yaml
ForkParameters:
- LdcEqualsLdd: [1, 0]
- FractionalLoad: [0, 1]
- TransposeLDS: [1, 0]
GroupForkParameters:
- - WorkGroup: [8, 8, 1]
    ThreadTile: [8, 8]
  - WorkGroup: [16, 16, 1]
    ThreadTile: [4, 4] 
- - DepthU: 16
    WorkGroupMapping: [1,15]  # here
  - DepthU: 8
    WorkGroupMapping: 8L
```

benjaminulmer:
@babakpst 
> Can we have a fork within the GroupForkParameter?

No, but that's a really good idea! Although, I have a feeling it would make the parsing much more complicated, especially if there are multiple forks within a group. It might be more appropriate to try and add that functionality in a future PR, along with some of the checks I mentioned in a TODO comment.

babakpst:
@benjaminulmer For the default sizes, I am not sure how many people are using this. I am fine if you remove this functionality, but that does not seem to be backward compatible.

babakpst:
> The "sizes" file is simply a list of exact sizes as they would normally appear in

Does that mean the Range is not acceptable?

Tensile/BenchmarkProblems.py
```diff
@@ -236,58 +242,103 @@ def benchmarkProblemType(problemTypeConfig, problemSizeGroupConfig, problemSizeG
...
+        cachePath = os.path.join(stepBaseDir, "cache.yaml")
+        pushWorkingPath("source")
 
-        # handle no valid solutions
-        if len(solutions) == 0:
-            msg = "Your parameters resulted in 0 valid solutions."
-            if globalParameters["PrintSolutionRejectionReason"]:
-                msg += "\nExamine reject and backtrace messages above to see why" \
-                        "and where solutions were rejected."
+        cache = False
```
babakpst, line 258: A future user may not immediately catch why we need this cache param here. A short comment would be great.

babakpst (approved):
LGTM

wbgilmartin (approved):
LGTM

benjaminulmer:
@babakpst 
> GroupForkParameter is basically a sub-category of ForkParameters and makes better sense if we list it under ForkParameters

The name is definitely not super clear as is. They do work very similarly to ForkParameters, so maybe listing the groups in there does make more sense. From a parsing perspective at least, it should make very little difference. I'd still love more feedback from others on this before any final decisions are made.

> For the default sizes, I am not sure how many people are using this. I am fine if you remove this functionality, but that does not seem to be backward compatible.

Yes, this is a breaking change. I'm not aware of anyone using the default sizes, and in my opinion they cover such a small set of sizes I find it hard to imagine them being useful. I think they're mostly a vestigial element from the old multistep tuning where it may have been convenient to not specify the sizes for each step. 

> The "sizes" file is simply a list of exact sizes as they would normally appear in

This was a poor choice of words (I always forget range sizes exist). This file can have sizes specified in any format that is valid normally, including range sizes.

benjaminulmer:
Removing the default sizes broke some tests. However, I've realized the desired behaviour I wanted for this PR could already be achieved by including the size portion of a config but leaving it empty, so I've restored the old behaviour.

babakpst:
@benjaminulmer I had a similar issue with the tests. I guess we need to update some of our tests.

cgmillette (commented):
@benjaminulmer Would it be impractical to parse 'GroupForkParameters' as just another nested list? Instead of having to be explicit about the grouping section, a nested list might be just as logical. Just not sure if this would be too much parsing spaghetti.

For example:
 
ForkParameters:
- LdcEqualsLdd: [1, 0]
- FractionalLoad: [0, 1]
- - AssertMinApproxSize: 3   # <---- Nested list understood as a single 'fork parameter' entity
    DepthU: 16
    InnerUnroll: 1
    WorkGroup: [32, 8, 1]
    WorkGroupMapping: 8

benjaminulmer:
@cgmillette 
You would have to do a pre processing step to separate the regular fork parameters from the groups, something like 
```python
forkRaw = getNonNoneFromConfig("ForkParameters", [])
self.paramGroups = []
for item in forkRaw:
    if type(item) is list:
        self.paramGroups.append(item)
for item in self.paramGroups:
    forkRaw.remove(item)

forkParams = dict(itertools.chain(*[x.items() for x in forkRaw]))
```
which isn't the most unwieldy code in the world, but is certainly a bit more verbose and (imo) a bit less self documenting than simply grabbing something by name.

Tensile/BenchmarkProblems.py
benjaminulmer: I updated the comment a few lines above this and changed the variable name to hopefully be more clear

benjaminulmer:
After some discussion, the general consensus is to go with Babak's proposal and have the Groups be listed under ForkParameters.

PR #1505, opened 2022-05-06
Split global read

AlexBrownAMD:
Add SplitGlobalRead option to create new global load patterns. If one wavefront accesses multiple rows, SplitGlobalRead enables rows to be arranged differently across lanes to avoid conflicts.

### DIFF
--- a/Tensile/Common.py
+++ b/Tensile/Common.py
@@ -396,6 +396,11 @@ validParameters = {
     "WaveSeparateGlobalReadA":    [ 0, 1 ],
     "WaveSeparateGlobalReadB":    [ 0, 1 ],
 
+    # Splits global read addresses within a wave into a number of smaller groups
+    # The default value of 1 reads the most contiguous elements across lanes,
+    # but higher values may help avoid bank conflicts
+    "SplitGlobalRead":            [1, 2, 4, 8, 16, 32],
+
     # PrefetchGlobalRead = 1:
     # Requires 2X LDS space, and VGPRs for buffering data on way into LDS
     #   prefetch / double-buffer reads from global memory -> vgprs -> lds.
@@ -1228,13 +1233,14 @@ defaultBenchmarkCommonParameters = [
     {"MaxOccupancy":              [ 40 ] },
     {"VectorWidth":               [ -1 ] },
     {"VectorStore":               [ -1 ] },
-    {"StoreVectorWidth":         [ -1 ] },
+    {"StoreVectorWidth":          [ -1 ] },
     {"GlobalReadVectorWidth":     [ -1 ] },
     {"LocalReadVectorWidth":      [ -1 ] },
     {"GlobalReadCoalesceVectorA": [ True ] },
     {"GlobalReadCoalesceVectorB": [ True ] },
-    {"WaveSeparateGlobalReadA":    [ 0 ] },
-    {"WaveSeparateGlobalReadB":    [ 0 ] },
+    {"WaveSeparateGlobalReadA":   [ 0 ] },
+    {"WaveSeparateGlobalReadB":   [ 0 ] },
+    {"SplitGlobalRead":           [ 1 ] },
     {"GlobalReadCoalesceGroupA":  [ True ] },
     {"GlobalReadCoalesceGroupB":  [ True ] },
     {"PrefetchGlobalRead":        [ 1 ] },
--- a/Tensile/Common.py
+++ b/Tensile/Common.py
@@ -399,7 +399,7 @@ validParameters = {
     # Splits global read addresses within a wave into a number of smaller groups
     # The default value of 1 reads the most contiguous elements across lanes,
     # but higher values may help avoid bank conflicts
-    "SplitGlobalRead":            [1, 2, 4, 8, 16, 32],
+    "SplitGlobalRead":            [1, 2, 4, 8],
 
     # PrefetchGlobalRead = 1:
     # Requires 2X LDS space, and VGPRs for buffering data on way into LDS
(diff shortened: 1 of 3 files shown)


Tensile/SolutionStructs.py
```diff
@@ -3214,6 +3214,24 @@ def assignDerivedParameters(state):
     state["LVCB"] = roundupRatio(state["LSCB"] , state["GlobalLoadVectorWidthB"])
     state["LVPB"] = roundupRatio(state["LSPB"] , state["GlobalLoadVectorWidthB"])
 
+    if state["SplitGlobalRead"] > 1:
+      if state["DirectToVgprA"] or state["DirectToVgprB"]:
+        reject(state, "SplitGlobalRead does not work with DirectToVgpr")
+      if state["GlobalReadCoalesceGroupA"]:
```
nakajee, line 3220: Don't we need to check B?

Tensile/SolutionStructs.py
```diff
@@ -3214,6 +3214,24 @@ def assignDerivedParameters(state):
...
+          # Fractional load use the more accurate lsc, multiply by VW later
+          divisorName = "LSCA"
+      else:
+        if state["GlobalReadCoalesceVectorA"]:
+          divisorName = "LSPA"
+        else:
+          divisorName = "LVPA"
+      divisor = state[divisorName]
+      if state["SplitGlobalRead"] >= divisor:
+        reject(state, "SplitGlobalRead must be less than lvc/lsc/lvp/lsc")
```
nakajee, line 3233: lsc appears twice.
lsp is not included.

Tensile/SolutionStructs.py
nakajee: These parameters can be different for B, right?

Tensile/SolutionStructs.py
```diff
@@ -3214,6 +3214,24 @@ def assignDerivedParameters(state):
     state["LVCB"] = roundupRatio(state["LSCB"] , state["GlobalLoadVectorWidthB"])
     state["LVPB"] = roundupRatio(state["LSPB"] , state["GlobalLoadVectorWidthB"])
 
+    if state["SplitGlobalRead"] > 1:
+      if state["DirectToVgprA"] or state["DirectToVgprB"]:
```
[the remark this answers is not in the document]
AlexBrownAMD: Good call, that feature is not implemented yet, I've made a ticket for SplitGlobalRead + DirectToLds support. Just pushed a change to add a reject for this case.

Tensile/SolutionStructs.py
AlexBrownAMD: Fixed

Tensile/KernelWriterAssembly.py
```diff
@@ -3420,6 +3420,12 @@ def graTileAssignment(self, kernel, tP):
       dummy       = self.vgprPool.checkOut(1, "dummy", self.preventVgprOverflowDuringNewTile)
       kStr += vectorStaticRemainder(dummy, dividendReg, "Serial", self.kernel["WavefrontSize"], tmpVgpr, tmpSgpr)
 
+    splitRead = kernel["SplitGlobalRead"]
+    # Split global read reorders reading rows within lanes of a wavefront
+    # If the wavefront is reading all from a single row, then disable split global read for this tensor
+    if divisor > self.kernel["WavefrontSize"]:
```
nakajee, line 3426: We do not actually need "self." here....
Kernel is passed as a parameter.

Tensile/Common.py
```diff
@@ -396,6 +396,11 @@ def getArchitectureName(gfxName):
     "WaveSeparateGlobalReadA":    [ 0, 1 ],
     "WaveSeparateGlobalReadB":    [ 0, 1 ],
 
+    # Splits global read addresses within a wave into a number of smaller groups
+    # The default value of 1 reads the most contiguous elements across lanes,
+    # but higher values may help avoid bank conflicts
+    "SplitGlobalRead":            [1, 2, 4, 8, 16, 32],
```
ramjana, line 402: I believe values 16, 32 are unnecessary choosing those values results into low memory load efficiency.

Tensile/SolutionStructs.py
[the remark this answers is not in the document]
ramjana: SplitGlobalRead feature is most useful for directToLds and should be supported for coalescing dimension of input tensor

Tensile/KernelWriterAssembly.py
```diff
@@ -3452,6 +3458,28 @@ def graTileAssignment(self, kernel, tP):
...
+      groupDivisor = waveSize // splitRead
+      groupOffset = waveSize // divisor
+      newDivisor = divisor // splitRead
+
+      kStr += vectorStaticRemainder(tmpVgpr, splitIndex, dividendReg, groupDivisor, tmpVgpr, tmpSgpr, "Split index")
+      kStr += vectorStaticDivideAndRemainder(qReg, rReg, splitIndex, newDivisor, tmpVgpr, tmpSgpr)
+
+      kStr += vectorStaticDivideAndRemainder(splitGroup, splitIndex, dividendReg, waveSize, tmpVgpr, tmpSgpr)
+
+      kStr += inst("v_mul_u32_u24", vgpr(splitGroup), groupOffset, vgpr(splitGroup), "Calculate wave group offset")
```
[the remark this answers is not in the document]
AlexBrownAMD: Added a check for this. I need to do a bit more testing, may be able to skip more in the case that groupOffset becomes 1

Tensile/KernelWriterAssembly.py
AlexBrownAMD: Fixed

Tensile/Common.py
AlexBrownAMD: Yea, mainly had them for testing. I've put in a change to remove those values since they'd be impractical.

Tensile/SolutionStructs.py
AlexBrownAMD: Aha yes you're right. I misread a section of the tensor code and thought they were hard-wired to the same value. Working on a fix now

AlexBrownAMD: Fixed

nakajee (approved):
Thanks for your update. Looks good.

PR #1509, opened 2022-05-17
updating HHS tests - LWPTENSILE-29

babakpst:
ComputeDataType is added to all HHS yaml files in Tensile/Tests so that Tensile does not throw the data type inconsistency warning.


AlexBrownAMD (approved):

benjaminulmer (approved):
Based on my (quick) check it doesn't look like you missed any files. gfx1030 timeout is not related to these changes. LGTM

PR #1514, opened 2022-05-20
Use shortened symlink in CI to prevent character limit

eidenyoshida:


AlexBrownAMD:
Looks like the rhel8 build has a new problem with the path. Also the cs9 build says py36 is missing.

AlexBrownAMD:
I think it's failing cause normal current working directory is build instead of project root. This patch should fix it:

[cipath.txt](https://github.com/ROCmSoftwarePlatform/Tensile/files/8807202/cipath.txt)
.

benjaminulmer (dismissed):
I'm not sure what the status of this PR is. If it's needed, we can probably get approval and merge since it's only jenkins changes. Until such a time, I'm blocking until after feature complete merge.

PR #1518, opened 2022-05-25
Config

tywerbicki:
Added a new example config file with the intention that it be referenced in the Tensile wiki for new users to familiarize themselves with Tensile.

The current config file that is referenced as an example in the wiki does not run as is. It contains deprecated parameters, poorly specified parameters, and requires over an hour to execute on a 6900xt once fixed. It also is not commented. This new config file hopes to address these shortcomings.

Following approval, the Tensile wiki will have to be updated accordingly.


Tensile/Configs/rocblas_sgemm_example.yaml
```diff
@@ -0,0 +1,80 @@
...
+      TransposeA: False
+      TransposeB: False
+      UseBeta: True
+      Batched: True
+
+    - # BenchmarkProblemSizeGroup - standard
+      InitialSolutionParameters:
+      # Empty for simplicity. 
+      # Please see https://github.com/ROCmSoftwarePlatform/Tensile/wiki/Benchmark-Protocol
+      # for a detailed explanation.
```
benjaminulmer, line 42: InitialSolutionParameters is no longer supported (see BenchmarkStructs.py). Also, I wouldn't reference that page of the wiki. We should update it, but in its current state it's horribly out of date.

Tensile/Configs/rocblas_sgemm_example.yaml
```diff
@@ -0,0 +1,80 @@
...
+      BenchmarkCommonParameters:
+      # Specify parameters that are to be considered "determined"
+      # prior to kernel forking.
+        - ProblemSizes:
+          - Exact: [1024, 1024, 1, 512]
```
benjaminulmer, line 47: ProblemSizes in anything but BenchmarkFinalParameters no longer do anything

Tensile/Configs/rocblas_sgemm_example.yaml
```diff
@@ -0,0 +1,80 @@
...
+      Batched: True
+
+    - # BenchmarkProblemSizeGroup - standard
+      InitialSolutionParameters:
+      # Empty for simplicity. 
+      # Please see https://github.com/ROCmSoftwarePlatform/Tensile/wiki/Benchmark-Protocol
+      # for a detailed explanation.
+      BenchmarkCommonParameters:
+      # Specify parameters that are to be considered "determined"
+      # prior to kernel forking.
```
benjaminulmer, line 45: "determined" is maybe a bit misleading. I would say something like 
"Parameters with a single value that are the same for each solution"

Tensile/Configs/rocblas_sgemm_example.yaml
tywerbicki: Removed InitialSolutionParameters and reference to wiki.

Tensile/Configs/rocblas_sgemm_example.yaml
tywerbicki: Removed.

Tensile/Configs/rocblas_sgemm_example.yaml
tywerbicki: Adjusted.

Tensile/Configs/rocblas_sgemm_example.yaml
```diff
@@ -0,0 +1,80 @@
...
+        - ProblemSizes:
+          - Exact: [1024, 1024, 1, 512]
+        - KernelLanguage: ["Assembly"]
+        - EdgeType: ["ShiftPtr"]
+        - LoopTail: [True]
+      ForkParameters:
+      # The following parameters are forked on because they often have an
+      # outsized impact on the performance of a kernel for a given problem 
+      # and a given macro tile.
```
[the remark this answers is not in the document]
tywerbicki: Adjusted.

Tensile/Configs/rocblas_sgemm_example.yaml
```diff
@@ -0,0 +1,72 @@
...
+          - [16, 16, 1]
+          - [8, 8, 4]
+        - WorkGroupMapping: [1, 4]
+        - TransposeLDS: [0, 1]
+        - LdsBlockSizePerPad: [-1, 0, 64]
+        - DepthU: [4, 8]
+        - StoreVectorWidth: [-1, 1, 2, 4]
+      BenchmarkFinalParameters:
+        - ProblemSizes:
+          # Specify the dimensions of the problems that the forked kernels
```
benjaminulmer, line 63: We're not very good about being precise about kernels vs solutions, but it would be nice to try and be consistent going forward. In which case, this should be "solutions".

benjaminulmer (approved):
LGTM. One final nitpick but I won't block merging on it.

benjaminulmer:
When ready, please squash-merge (and maybe make the name a bit more descriptive; something like "Add example config")

Tensile/Configs/rocblas_sgemm_example.yaml
tywerbicki: Thanks for the clarification. Adjusted.

PR #1522, opened 2022-05-30
Switch default MACInstruction to FMA

AlexBrownAMD:
FMA matches MAC performance and integrates additional flags.


nakajee:
Why don't we specify default parameter for MACInstruction?
    {"MACInstruction":            [ '' ]},
      -> set FMA as default

nakajee:
Thanks for your update.
I do not have any further comment on your change.
Please make sure this works for all arch and CI tests pass.

yoichiyoshida (approved):

AlexBrownAMD:
1 test failure on CI. Ran the tests locally a few times with no errors. Appears to be an intermittent CI issue to be monitored

PR #1523, opened 2022-05-31
GridBased kernel selection

aazz44ss:
This solution selection need rocblas to point to custom logic file consist of grid sizes.
If incoming size falls inside grid boundary (32768 in this case), searching toward larger M,N, and nearest K. Otherwise, searching toward nearest boundary and nearest K.

### DIFF
--- a/Tensile/Source/lib/include/Tensile/Distance.hpp
+++ b/Tensile/Source/lib/include/Tensile/Distance.hpp
@@ -331,6 +331,58 @@ namespace Tensile
             }
         };
 
+        template <typename Key>
+        struct GridBasedDistance : public Distance<Key>
+        {
+            enum
+            {
+                HasIndex = false,
+                HasValue = false
+            };
+
+            static std::string Type()
+            {
+                return "GridBased";
+            }
+            virtual std::string type() const override
+            {
+                return Type();
+            }
+
+            inline double operator()(Key const& p1, Key const& p2) const
+            {
+                double distance = 0.0;
+
+                double M = p2[0];
+                double N = p2[1];
+                double K = p2[2];
+
+                // This is hard coding workaround solution //
+                // If incoming_size falls inside grid boundary (32768 in this case), searching toward larger M,N.
+                // If incoming_N > 32768, searching toward larger M and nearest boundary.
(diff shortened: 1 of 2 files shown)


nakajee:
I am not a reviewer of this PR, but I have some comments.

(1)100 * stepM * stepN can cause integer overflow (like M=N=1, p1[0]=p1[1]=32768).
   (because * is left-associative and double operation is done after this part of multiplications)
  I suggest to use double type for stepM and stepN.
(2)We can set stepM = 1 if p1[0] > 32768 and stepN = 1 if p1[1] > 32768.
  Then, we can unify else if, else if and else part.
  Also, distance += (std::abs(K - p1[2]) / (K + p1[2] * 8)); is same for all parts.
  We can move this after if-else to make the code simpler.

aazz44ss:
> I am not a reviewer of this PR, but I have some comments.
> 
> (1)100 * stepM * stepN can cause integer overflow (like M=N=1, p1[0]=p1[1]=32768). (because * is left-associative and double operation is done after this part of multiplications) I suggest to use double type for stepM and stepN. (2)We can set stepM = 1 if p1[0] > 32768 and stepN = 1 if p1[1] > 32768. Then, we can unify else if, else if and else part. Also, distance += (std::abs(K - p1[2]) / (K + p1[2] * 8)); is same for all parts. We can move this after if-else to make the code simpler.

@nakajee Thanks

Tensile/Source/lib/include/Tensile/Distance.hpp
```diff
@@ -331,6 +331,58 @@ namespace Tensile
...
+                return Type();
+            }
+
+            inline double operator()(Key const& p1, Key const& p2) const
+            {
+                double distance = 0.0;
+
+                double M = p2[0];
+                double N = p2[1];
+                double K = p2[2];
```
benjaminulmer, line 358: Eventually we'll need to figure out a way to not have this hardcoded, but for now this is fine.

benjaminulmer (approved):
I still need to sit down and wrap my head around the calculations you're doing, but the code looks good to me. Please squash-merge when you're ready.

PR #1524, opened 2022-06-02
env variable can be windows or linux path style

TorreZuk:


cgmb:
Won't we just run into this problem again in the next library up the stack?

cgmb:
I suppose I wonder why the `ROCM_PATH` can't just be set with forward slashes in the first place?

benjaminulmer (approved):
Discussed with Cory offline. Please squash-merge when ready.

AlexBrownAMD (approved):

yoichiyoshida (approved):

TorreZuk:
I will wait for CI to come back.  yes @cgmb we could document all our env variables should use linux style paths if we want, but even then if a user copies and pastes a correct windows style path into one of our standard env variables I would prefer not to fail to build (with strange errors).   We should try to make things robust so both work.   I would suggest all paths read from environment variables should use the file( TO_CMAKE_PATH helper to be safe.

TorreZuk:
@benjaminulmer and @yoichiyoshida is it worth re-running CI as it looks like all your PRs are failing extended tests, what is broken on your side?

AlexBrownAMD:
Looks like the same test is failing on this and other PRs. Looking into the cause now. Merging this PR should be ok cause the error is not related.

AlexBrownAMD:
Looks like the error is related to current build being run on CI, error does not happen on older builds.

PR #1526, opened 2022-06-02
Solution selection selection

nielenventer:
Currently in Tensile there is a mechanism to select libraries benchmarked using device efficiency or using per CU efficiency, at runtime. This change makes this selection more generic so we can use it to test alternative solution selection strategies e.g. the ML based approach proposed by Bill, or Henry's distance metric.

The "Experimental" option is added as a catch-all option for any experiments we want to try (or allow other users to try).

The "DeviceEfficiency" and "CUEfficiency" options are preserved for the sake of compatibility, although ideally I think we'd like to deprecate them for the sake of simplifying the user experience. If one or the other are better for a use-case, we should make that decision and the selection should happen under the hood.


benjaminulmer:
I'll review in full once the rocBLAS API proposed changes are "approved". Although this mostly just renames things, so I expect it is all fine.

One thing I'll point out now is that by changing the name of the GlobalParameter, we are breaking any tuning configs that have the old name of the parameter. None of the configs in the repo have that parameter, and afaik I'm the only person who has ever done tuning with the parameter, so this very well may be an OK price to pay instead of having to continue support for the old name. Still, it's worth being aware of.

Tensile/Source/client/include/LogReporter.hpp
```diff
@@ -96,10 +96,10 @@ namespace Tensile
                     level = args["log-level"].as<LogLevel>();
                 std::cout << "Log level: " << level << std::endl;
 
-                PerformanceMetric metric = args["performance-metric"].as<PerformanceMetric>();
-                // Default to 'DeviceEfficiency' benchmarking if CUEfficiency not specified
+                SolutionSelectionMethod metric = args["solution-selection-method"].as<SolutionSelectionMethod>();
```
benjaminulmer, line 99: proposed: keep the arg as `performance-metric`

Tensile/Source/client/include/ResultFileReporter.hpp
```diff
@@ -72,7 +72,7 @@ namespace Tensile
             bool              m_invalidSolution = false;
             bool              m_extraCol;
             bool              m_mergeSameProblems;
-            PerformanceMetric m_performanceMetric;
+            SolutionSelectionMethod m_solutionSelectionMethod;
```
benjaminulmer, line 75: proposed: keep the name as `m_performanceMetric`

benjaminulmer (commented):
Renaming from PerformanceMetric to SolutionSelectionMetric makes sense in the context of solution selection, but for benchmarking this is a bit weird since it truly is a performance metric. Solution selection is more "user" facing, so having the type name updated seems appropriate. However, in situations related to benchmarking where we don't _have_ to update the names of thing, my proposal is to keep them as was before. This would also fix the backwards compatibility issue I mentioned above. 

I tried to identify all these areas with comments, but I may have missed some. Hopefully this all makes sense.

nielenventer:
Ya that makes sense. They are slightly different, so not quite a round peg in a square hole, but still a bit of a squeeze.

From Tensile's point of view it makes sense to group them, since mechanically they just select the library, but keeping them separate conceptually where appropriate makes sense.

nielenventer:
As discussed, scope of change was reduced to just adding new performance metric, along with env variable to toggle it.

Renaming the performance metric to make it more generic will be tackled in a later change.

benjaminulmer (approved):
LGTM.

It'd be nice if someone outside of the solution selection team could approve as well (this PR is a lot shorter now).

PR #1527, opened 2022-06-07
win32 clang 15.0.0 branches may only take -inputs

TorreZuk:
* this may likely be reverted once clang branches align better
* for now version can not be used to detect argument support on windows

### DIFF
--- a/Tensile/TensileCreateLibrary.py
+++ b/Tensile/TensileCreateLibrary.py
@@ -207,7 +207,8 @@ def buildSourceCodeObjectFile(CxxCompiler, outputPath, kernelFile):
       # for hipclang 5.2 and above, clang offload bundler changes the way input/output files are specified
       inflag = "-inputs"
       outflag = "-outputs"
-      if (hipccMaj == 5 and hipccMin >= 2) or hipccMaj >= 6:
+      # clang 15.0.0 still wants inputs on windows so the hipcc version condition doesn't work
+      if os.name != "nt" and ((hipccMaj == 5 and hipccMin >= 2) or hipccMaj >= 6):
         inflag = "-input"
         outflag = "-output"
 


Tensile/TensileCreateLibrary.py
```diff
@@ -207,7 +207,8 @@ def buildSourceCodeObjectFile(CxxCompiler, outputPath, kernelFile):
       # for hipclang 5.2 and above, clang offload bundler changes the way input/output files are specified
       inflag = "-inputs"
       outflag = "-outputs"
-      if (hipccMaj == 5 and hipccMin >= 2) or hipccMaj >= 6:
+      # clang 15.0.0 still wants inputs on windows so the hipcc version condition doesn't work
+      if os.name != "nt" and ((hipccMaj == 5 and hipccMin >= 2) or hipccMaj >= 6):
```
cgmb, line 211: This is the sort of thing that would normally be checked at configuration-time with something akin to [`CheckCXXCompilerFlag`](https://cmake.org/cmake/help/latest/module/CheckCXXCompilerFlag.html) rather than depending on the version number. The Windows fork is just one example of where the version check is insufficient. Another would be the upstream clang compiler. The upstream releases of clang-13 and clang-14 are sufficient to build HIP language programs, but don't correspond to any particular ROCm version.

I know it's not the thing you're changing, but I also wanted to point out that this comparison can be simplified using tuples:
```
(hipccMaj, hipccMin) >= (5, 2)
```

TorreZuk:
So ./Tensile/Source/lib/include/Tensile/Distance.hpp  and ./Tensile/Source/lib/include/Tensile/Serialization/MatchingLibrary.hpp appear to fail Static analysis (not my change), @aazz44ss will you fix that failure from your PR #1523 ?
As to this PR @cgmb I will leave it up to @yoichiyoshida if he wants to later refactor things into cmake configure from python runtime detection and tuple comparison, my PR I'll leave out suggested edits as expect I'll revert by commit to remove soon once we get a toolchain update in the private branch.

benjaminulmer (approved):
Extended failures are unrelated to this change. I've created #1530 to fix the static analysis failures.

I'm fine with this change as a quick fix. It may be worth revisiting at some point to have a more robust solution.

PR #1536, opened 2022-06-10
DirectToLds for hgemm configurations

AlexBrownAMD:
Modify parameter check for DirectToLds for hgemm and add extended test suite with working configurations

### DIFF
--- a/Tensile/SolutionStructs.py
+++ b/Tensile/SolutionStructs.py
@@ -2442,13 +2442,6 @@ class Solution(collections.abc.Mapping):
       reject(state, "can't use DirectToLds for NumThreads % WavefrontSize != 0")
       return False
 
-    # GLVW*BPe only for precision(s) < 4 (bpe)
-    #if (state["ProblemType"]["TLU%c"%tc] == True and numBytes < 4): 
-    # if (numBytes < 4): 
-    #   if state["GlobalLoadVectorWidth%c"%tc] * numBytes != 4:
-    #     reject(state, "can't use DirectToLds for bpe < 4 and GlobalLoadVectorWidth * numBytes != 4")
-    #     return False
-
     if state["ProblemType"]["TLU%c"%tc] == state["UnrollMajorLDS%c" % tc]:
       reject(state, "can't use DirectToLds for TLU%c == UnrollMajorLDS%c"%(tc, tc))
       return False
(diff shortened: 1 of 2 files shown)


Tensile/SolutionStructs.py
```diff
@@ -2420,9 +2420,16 @@ def isDirectToLdsDoable(state, tc):
...
     # use for all precisions with TransposeLDS=1
 
-    if state["ProblemType"]["DataType"].isHalf() and state["AssertSummationElementMultiple"] % (2 * state["GlobalLoadVectorWidth%c"%tc])  != 0:
-      reject(state, "can't use DirectToLds for FP16 with AssertSummationElementMultiple %u" % state["AssertSummationElementMultiple"])
-      return False
+    if state["ProblemType"]["DataType"].isHalf():
+      if state["AssertSummationElementMultiple"] % (2 * state["GlobalLoadVectorWidth%c"%tc])  != 0:
+        reject(state, "can't use DirectToLds for FP16 with AssertSummationElementMultiple %u" % state["AssertSummationElementMultiple"])
+        return False
+      if state["ProblemType"]["TransposeA"] != True or state["ProblemType"]["TransposeB"] != False:
```
nakajee, line 2430: Probably, we need to improve this part.
Actually, I changed isDirectToLdsDoable to reject instead of just changing DirectToLds = 0.
This is because I did not want to run same combinations by converting DTL to 0.
However, I realized that this prevented enabling DirectToLds for only A or B.
In this case, we should be able to apply DirectToLds for A if TransposeA=false (regardless of B).

I think it is better to change back this part not to reject code here.

AlexBrownAMD: For hgemm, DirectToLds only works for TN, all other transpose combos currently fail validation. So the check for specific transpose types is accurate in this case.

nakajee: OK. We cannot combine like DirectToLdsA= True with DirectToLds=B.
For HGEMM only?
For other types, I think we should be able to do that if we changed the reject logic in this function.

nakajee: I understand we do not support other types for now, but technically we can support them in the future.

nakajee (approved):
Looks good.

PR #1539, opened 2022-06-14
Update copyright githook

benjaminulmer:
Old regex wasn't working properly for the updated header. This updates the regex to be the same as the one being used by rocBLAS. 

Did a quick test and this one seems to be working.


AlexBrownAMD (approved):
Ran a couple quick tests with single year and year range and it works.

PR #1512, opened 2022-05-20
[windows] Remove boost dependency in client build.

hcman2:
In windows platform, the boost library filename is modified.
So remove boost.

PR #1541, opened 2022-06-21
Revert "[windows] Remove boost dependency in client build. (#1512)"

benjaminulmer:
This reverts commit 4689a9d3ab47f000a09f084e34cd1eb2531529b3.

@AlexBrownAMD brought up that the changes in #1512 might create issues with Palamida. There are two possible solutions:
1. Revert the changes (this PR)
2. Ensure the changes are compliant. At a minimum, the copyright header in program_options.hpp would need to be updated.

Does anyone know if there's a quick and easy way to get a scan done?


benjaminulmer:
@TorreZuk @amcamd 
I don't think reverting this would cause any issues for rocBLAS since all the changes are to the tensile client, not the host library itself.

TorreZuk:
> @TorreZuk @amcamd I don't think reverting this would cause any issues for rocBLAS since all the changes are to the tensile client, not the host library itself.

Agree we shouldn't be affected in rocBLAS

PR #1544, opened 2022-06-23
ROCm 5.3 merge master into develop

benjaminulmer:


TorreZuk (commented):
Fine by me, but do you go to 4.35 in develop for 5.4 ?  @amcamd they did it differently in Tensile than us I recall

benjaminulmer:
Historically it seems we haven't updated the version in develop. Although, I'd be more than willing to switch to the model rocBLAS uses as I find it more intuitive. Maybe Andrew can weigh in on if there's a reason for Tensile not updating the version in develop.

amcamd:
I think it is best to keep with the current system until we know if the proposed manifest based testing allows us to create release branches at feature complete. With the rocBLAS/rocSOLVER/hipBLAS version numbering we need to coordinate the first merge from staging to master after the release branches are created, we do not need any coordination with Tensile.

amcamd (approved):
checked that the minor version goes from 33 to 34

amcamd (approved):
Confirm version is at 4.34.0

PR #1546, opened 2022-06-27
Kernel converter

aferoz21:


Tensile/kernelConverter.py
```diff
@@ -0,0 +1,317 @@
...
+
+    YamlFileData.append("    - # BenchmarkProblemSizeGroup\n") 
+    YamlFileData.append("      InitialSolutionParameters:\n") 
+    YamlFileData.append("      BenchmarkCommonParameters:\n") 
+    
+    YamlFileData.append("      ForkParameters:\n")
+    macInstruction = ""
+    for forkKey, forkValue in currentIndexSolution.items():
+      if forkKey == "MACInstruction":
+        if forkValue == "MAC": #This is a bug in the library logic file
```
nakajee, line 134: Are you sure about "This is a bug"?

aferoz21:
1) Tool with the below usage is generated. 
 python3 ../bin/kernelConverter <lib Logic File .> <solution Index> <output tensile Yaml file>.
2) Validated with SGEMM, DGEMM, HPA HGEMM inputs. generated tensile file ran on tensile client.
3) Flake8 has been ran on the script.

Tensile/kernelConverter.py
```diff
@@ -0,0 +1,317 @@
...
+def calculateMatrixMNThreadTileMacroTileWorkGroupParameters(MIBlock,MIWaveTile,MIWaveGroup):
+
+  MatrixM = int(MIBlock[0]) *int(MIBlock[4])
+  MatrixN = int(MIBlock[1])
+  TT0 = int(MIWaveTile[0]) 
+  TT1 = int(MIWaveTile[1])*int(MIBlock[1])
+  MT0 =  MatrixM*int(MIWaveTile[0])*int(MIWaveGroup[0])
+  MT1 =  MatrixN*int(MIWaveTile[1])*int(MIWaveGroup[1])
+  WG0 =  int(MIBlock[0]) * int(MIBlock[4])*int(MIWaveGroup[0])
+  WG1 = int(MIWaveGroup[0]) * int(MIWaveGroup[1]) *64 // int(WG0)
```
nakajee, line 301: Don't we need to use WavefrontSize instead of 64?

Tensile/kernelConverter.py
nakajee: Hi @benjaminulmer 
Do you know why we have MACInstruction: MAC in rocblas logic file?

nakajee:
Can we have TransposeA,B in output yaml file?
The logic file name tells which transpose is used.

nakajee:
> Can we have TransposeA,B in output yaml file?

It looks like it is because the value was same as default (NT).

Tensile/kernelConverter.py
```diff
@@ -0,0 +1,317 @@
...
+    YamlFileData.append("#########################################################################################\n")
+    YamlFileData.append("LibraryLogic:\n")
+    YamlFileData.append("    ScheduleName: %s\n"%scheduleName)
+    YamlFileData.append("    DeviceNames: %s\n"%deviceNames)
+    YamlFileData.append("    ArchitectureName: \"%s\"\n"%architectureName)
+        
+def CheckMacroTileThreadTileWorkGroupMatches(currentSolution):
+
+  if not currentSolution.keys() & {'MIBlock','MIWaveTile','MIWaveGroup'}:
+    raise RuntimeError("one or more of these fields => MIBlock, MIWaveTile, MIWaveGroup is missing in the library logic file!! ..\n")
```
babakpst: I guess, if these items are not available, then the config file does not have MatrixInstruction and it is only based on ThreadTile and WorkGroup. For example SolutionIndex: 15, in ../asm_full/arcturus_Cijk_Ailk_Bjlk_SB.yaml. If that is the case, all you need is to add TT and WG. Can you please check this out?

Tensile/bin/kernelConverter
```diff
@@ -0,0 +1,13 @@
...
+    import os.path
+    import sys
+    parentdir = os.path.normpath(os.path.join(os.path.dirname(os.path.realpath(__file__)), "..", ".."))
+    sys.path.append(parentdir)
+
+    from Tensile import kernelConverter
+
+# script run from commandline
+if __name__ == "__main__":
+    kernelConverter.main()
```
babakpst, line 13: We should change this file to an executable.

babakpst (changes requested):

babakpst:
@nakajee The plan is to only keep non-default parameters, except for some basic parameters. @aferoz21 Can you add TransposeA, TransposeB, UseBeta, and Batched to the yaml, even if they are the same as default?

Tensile/kernelConverter.py
aferoz21: yes ,MI* parameters are missing for if the matrix instruction is empty or EnableMatrixInstruction is false
EnableMatrixInstruction: false
MatrixInstruction: []

benjaminulmer (changes requested):
Good work on this tool. Converting library logic to tuning configs is something many people have tried to do, and if we can create a tool that successfully does that it would be a big help.

Some more general comments in addition to what's bellow: 
- The name should be changed to Tensile\<something\>
  - "kernelConverter" is not a good name in my opinion; solution parameters---not kernels---are being converted to a different format. TensileTuningConfig is a bit more accurate and would follow the naming convention set by TensileClientConfig, but I'm open to other suggestions
- Please follow the convention of reserving names beginning with a Capital letter for types; variables and functions should begin with a lowercase letter
- Please remove code that is commented out

Some other points for discussion:
- I usually prefer creating the yaml representation in memory and relying on yaml.dump as opposed to creating the string manually, as it is more resistant to bugs and formatting errors
- I need to look at some of the logic more carefully, but I wonder if some of the special checks you do for some parameters could be handled by first creating the solution object from the library logic file and then using that to fill in what should go into the tuning config. This might even help with the current bug where non-mfma solutions cause errors.

Tensile/kernelConverter.py
benjaminulmer: @nakajee My answer to Babak previously on this matter:
> I don't have a specific answer, but there are a lot of ways parameter values can make it into rocBLAS that aren't "valid" values as specified in Common.py. Usually what happens is they get set that way in SolutionStructs for some reason or another

> and that might have happened in an older version of Tensile where that no longer happens anymore

Tensile/kernelConverter.py
aferoz21: some kernel does not have WavefrontSize parameter,  in that case , do we have to default to 64?

nakajee: Like other parameters, we should be able to use the default value if it is not explicitly defined.

Tensile/TensileLibLogicToYaml.py
```diff
@@ -0,0 +1,319 @@
...
+from .Common import print1,HR,defaultBenchmarkCommonParameters,defaultProblemType
+from .DataType import DataType
+             
+import argparse
+import os
+import sys
+import re
+
+#accumulates the yaml file data to be printed in tensile input yaml file
+def TensileLibLogiceToYaml(userArgs):
```
babakpst, line 35: Logice -> Logic

Tensile/TensileLibLogicToYaml.py
```diff
@@ -0,0 +1,319 @@
...
+    forkParametersYamlData = []
+    forkParametersYamlData.append("    - # BenchmarkProblemSizeGroup\n") 
+    forkParametersYamlData.append("      InitialSolutionParameters:\n") 
+    forkParametersYamlData.append("      BenchmarkCommonParameters:\n") 
+    forkParametersYamlData.append("      ForkParameters:\n")
+    
+    macInstruction = ""
+    for forkKey, forkValue in currentIndexSolution.items():
+      if forkKey == "MACInstruction":
+        if forkValue == "MAC": #This is a bug in the library logic file
```
nakajee, line 186: We do not need to append maxInstruction, right?
It looks like it is removed later.

Tensile/TensileLibLogicToYaml.py
```diff
@@ -0,0 +1,319 @@
...
+    
+    forkParametersYamlData.append("    - # BenchmarkProblemSizeGroup\n") 
+    forkParametersYamlData.append("      InitialSolutionParameters:\n") 
+    forkParametersYamlData.append("      BenchmarkCommonParameters:\n") 
+    
+    forkParametersYamlData.append("      ForkParameters:\n")
+    macInstruction = ""
+    for forkKey, forkValue in currentIndexSolution.items():
+      if forkKey == "MACInstruction":
+        if forkValue == "MAC": #This is a bug in the library logic file
```
nakajee, line 150: You can simplify this part of code.
You just need to change forkValue to "MAD" if it is MAC

if forkValue == "MAC"
  forkValue = "MAD"

Tensile/TensileLibLogicToYaml.py
```diff
@@ -0,0 +1,319 @@
...
+        # Find the matching index for fork key name from list of dictionaries => defaultBenchmarkCommonParameters
+        index =  next((i for i,d in enumerate(defaultBenchmarkCommonParameters) if forkKey in d), None)
+        if index != None:
+          #Strip the leading and trailing [] brackets, trim the leading and trailing spaces
+          DefaultValue = str(defaultBenchmarkCommonParameters[index][forkKey])[1:-1].strip()
+          
+          # remove the "" from the string
+          DefaultValue = re.sub("[\"\']", "", DefaultValue)
+          if str(forkValue) != DefaultValue:
+            if forkKey == "ThreadTile" or forkKey == "WorkGroup":
```
nakajee, line 169: Can we make this more generic?
Like, check if the type of forkKey is list or not, or something similar way...
Maybe like this.
  if isInstance(forkValue, list) == False:
    forkValue = [forkValue] # convert to list

Tensile/TensileLibLogicToYaml.py
aferoz21: Thank you. if we are using MI instruction we don't need MAC instruction earlier. I removed that logic. earlier we thought both MI instruction and non MI instruction could come here.

Tensile/TensileLibLogicToYaml.py
babakpst: This may work, but I prefer we fix the library logics to be compatible with the current state of Tensile, rather than tweak this code to compensate for the issue with the MAC. That would be a separate ticket.

babakpst (changes requested):
The values for ThreadTile and WorkGroup should be on the next line, similar to MI. Try 

../Tensile/bin/TensileLibLogicToYaml .../asm_full/arcturus_Cijk_Ailk_Bjlk_SB.yaml 15 sgemm15.yaml

Then, the output in the yaml file is:
```yaml
- ThreadTile: [4, 4]
- WorkGroup: [16, 8, 2]
```
which should be:
```yaml
- ThreadTile: 
           - [4, 4]
- WorkGroup: 
           - [16, 8, 2]
```

nakajee:
> ```yaml
> - ThreadTile: [4, 4]
> - WorkGroup: [16, 8, 2]
> ```

It looks like this causes an error....
Could you please confirm the output can work without an error?
Please cover possible scenarios and make sure the output works.

aferoz21:
> > > ```yaml
> > > - ThreadTile: [4, 4]
> > > - WorkGroup: [16, 8, 2]
> > > ```
> > 
> > 
> > It looks like this causes an error.... Could you please confirm the output can work without an error? Please cover possible scenarios and make sure the output works.
> 
> It looks like using []can work.
> 
> * ThreadTile: [[4, 4]]
> * WorkGroup: [[16, 8, 2]]
> 
> I am OK to put it in the same line as long as it works. You do not need separate process even if the value is list type. We always need to add each value in [ ].

aferoz21:
Accidently hit the closed. 
Since matrix instruction uses the other format, currently for thread tile and work group I am printing it in next line as like matrix instruction.

aferoz21:
> > > ```yaml
> > > - ThreadTile: [4, 4]
> > > - WorkGroup: [16, 8, 2]
> > > ```
> > 
> > 
> > It looks like this causes an error.... Could you please confirm the output can work without an error? Please cover possible scenarios and make sure the output works.

Similar to matrix instruction , ThreadTile and Workgroup is printed in the next line. add this test case in my golden suites. if any one have other test cases I can add it my list

babakpst (approved):
LGTM. Thanks, Feroz.

benjaminulmer (approved):
Didn't have a chance to review again in detail, but it seems my major concerns have been addressed. 

When you merge please squash-merge and set the commit message to match the new name for the app

PR #1550, opened 2022-07-15
use --help to check ClangOffloadBundler input option

TonyYHsieh:
window's hipcc/clang version is different to linux.
So use ClangOffloadBundler  --help information to decide to use --input or --inputs.


TorreZuk (approved):
If this works fine.  But we finally got promotion on the hipcc fixes which brings the later llvm as well.  It just changed last night windows CI compute-main-osdb build 125 clang offload now has --input.   So assuming we don't have to revert hip back this is just adding robustness, vs. just drop the os.name != "nt" now.

As a design pattern this kind of stuff should also likely be done once and not every invocation, as tensile is the bottleneck in building rocblas so prefer not to slow it down.

TonyYHsieh:
close as compute-main-osdb will make change later

PR #1551, opened 2022-07-18
Revert "win32 clang 15.0.0 branches may only take -inputs (#1527)"

TorreZuk:
* This reverts commit 50f32c62266154704142da6b077b5d5114069a94 which was a workaround
* hipcc version wasn't tightly linked to clang-offload-version but should work now on windows CI as updated hip


benjaminulmer (approved):
CI is doing some weird stuff, but it shouldn't be related to this change. LGTM

PR #1553, opened 2022-07-25
[dual mac] Support dual mac instruction on new platform.

hcman2:
New platform supports dual instruction format to improve performance.
Modify Tensile to generate this instructions.

### DIFF
--- a/Tensile/Components/MAC_F32.py
+++ b/Tensile/Components/MAC_F32.py
@@ -24,6 +24,7 @@
 
 from ..Component import Component, MAC
 from ..DataType import DataType
+import queue
 
 class MAC_F32_Plain(MAC):
     """
@@ -50,6 +51,10 @@ class MAC_F32_Plain(MAC):
         if not writer.asmCaps[instruction]:
             raise RuntimeError("{} instruction specified but not supported on {}".format(instruction, kernel["ISA"]))
 
+        dualMacEnable = 0
+        if writer.asmCaps["v_dual_fmac_f32"] and kernel["WavefrontSize"] == 32:
+            dualMacEnable = 1
+
         kStr = self.commentHeader()
 
         vars = {}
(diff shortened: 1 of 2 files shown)


TonyYHsieh (approved):
LGTM

Tensile/Components/MAC_F32.py
```diff
@@ -66,32 +71,86 @@ def __call__(self, writer, m, innerUnroll):
...
+                    for iui in range(0, innerUnroll):
+                        vars["idx0"] = idx0
+                        vars["idx1"] = idx1
+                        vars["a"] = idx0 if writer.tPB["tile01Idx"] else idx1
+                        vars["b"] = idx1 if writer.tPB["tile01Idx"] else idx0
+                        vars["iui"] = iui
+
+                        vars["cStr"] = "v[vgprValuC + {idx0} + {idx1}*{ThreadTile0}]".format_map(vars)
+                        vars["aStr"] = "v[vgprValuA_X{m}_I{iui} + {a}]".format_map(vars)
+                        vars["bStr"] = "v[vgprValuB_X{m}_I{iui} + {b}]".format_map(vars)
```
AlexBrownAMD, line 87: Could be reorganized to reduce duplicated code

Tensile/Components/MAC_F32.py
```diff
@@ -50,6 +51,10 @@ def __call__(self, writer, m, innerUnroll):
         if not writer.asmCaps[instruction]:
             raise RuntimeError("{} instruction specified but not supported on {}".format(instruction, kernel["ISA"]))
 
+        dualMacEnable = 0
+        if writer.asmCaps["v_dual_fmac_f32"] and kernel["WavefrontSize"] == 32:
+            dualMacEnable = 1
+
```
AlexBrownAMD, line 57: Is there ever a case where the instruction is available, but we don't want to use it? This change could invalidate tuning results because there is no option to generate kernels the way they were when tuning was done.

AlexBrownAMD (changes requested):
Do we have any test cases that run on CI that specifically test this feature?

PR #1554, opened 2022-07-26
Add debug flag for printing library logic index of winning solution

benjaminulmer:
Mostly for development purposes.

### DIFF
--- a/Tensile/Source/lib/include/Tensile/MasterSolutionLibrary.hpp
+++ b/Tensile/Source/lib/include/Tensile/MasterSolutionLibrary.hpp
@@ -117,7 +117,8 @@ namespace Tensile
                                                             Hardware const&  hardware,
                                                             double* fitness = nullptr) const
         {
-            const int solution_index = Debug::Instance().getSolutionIndex();
+            const int                   solution_index = Debug::Instance().getSolutionIndex();
+            std::shared_ptr<MySolution> rv;
 
             if(solution_index >= 0)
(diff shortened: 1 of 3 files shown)


Tensile/Source/lib/include/Tensile/MasterSolutionLibrary.hpp
```diff
@@ -134,15 +135,21 @@ namespace Tensile
...
                 }
             }
             else
-                return library->findBestSolution(problem, hardware, fitness);
-        }
+                rv = library->findBestSolution(problem, hardware, fitness);
 
+            if(Debug::Instance().printLibraryLogicIndex() && rv)
+            {
+                std::cout << "Library logic solution index of winning solution: "
```
nielenventer, line 148: nit: I'd vote to have a separate message for no solution too, more info is probably better if we're debugging

benjaminulmer: Sure. The user would probably find out pretty quick anyways if no solution is found, but having a message for logging purposes could be nice. How's a simple "No solution found" sound?

nielenventer (approved):

babakpst (approved):
I built rocblas with this branch. It drops the correct solution index from the lib logic.

benjaminulmer:
Precheckin gfx908 fails are also showing up in develop, so they're not related to this change.

PR #1555, opened 2022-07-26
Fix benchmark timings when using warmups and/or validation

AlexBrownAMD:
Fix some bugs that caused incorrect benchmark timing results. GPU timing is impacted by the overhead of using hip events, on small kernels, the timing is noticeable. CPU timing is more precise, but when using warmups, the warmup run time was included in the benchmark time, but only averaged over the actual number of benchmark runs.
Also, when doing both benchmarking and validation the times would be very inaccurate. To prevent validation from impacting benchmark times, validation is now done after the first benchmark run instead of the last (for best results there should be at least 2 warmups). OpenMP was also defaulting to 1 thread, which was somehow impacting other code. Set num threads to 64 and enabled the option for dynamic adjustment based on hardware.

### DIFF
--- a/Tensile/Source/client/main.cpp
+++ b/Tensile/Source/client/main.cpp
@@ -602,36 +602,22 @@ int main(int argc, const char* argv[])
                             TimingEvents warmupStartEvents(warmupInvocations, eventCount);
                             TimingEvents warmupStopEvents(warmupInvocations, eventCount);
 
-                            if(warmupInvocations > 0)
+                            for(int i = 0; i < warmupInvocations; i++)
                             {
-                                // Do validation after first warmup
                                 listeners.preWarmup();
                                 if(gpuTimer)
                                     HIP_CHECK_EXC(adapter.launchKernels(kernels,
                                                                         stream,
-                                                                        warmupStartEvents[0],
-                                                                        warmupStopEvents[0]));
+                                                                        warmupStartEvents[i],
+                                                                        warmupStopEvents[i]));
                                 else
                                     HIP_CHECK_EXC(
                                         adapter.launchKernels(kernels, stream, nullptr, nullptr));
                                 listeners.postWarmup();
-                                listeners.validateWarmups(
-                                    inputs, warmupStartEvents, warmupStopEvents);
-
-                                // Remainder of warmups
-                                for(int i = 1; i < warmupInvocations; i++)
-                                {
-                                    listeners.preWarmup();
-                                    if(gpuTimer)
-                                        HIP_CHECK_EXC(adapter.launchKernels(kernels,
-                                                                            stream,
-                                                                            warmupStartEvents[i],
-                                                                            warmupStopEvents[i]));
-                                    else
-                                        HIP_CHECK_EXC(adapter.launchKernels(
-                                            kernels, stream, nullptr, nullptr));
-                                    listeners.postWarmup();
-                                }
+                                // Do validation after first warmup
+                                if(i == 0)
+                                    listeners.validateWarmups(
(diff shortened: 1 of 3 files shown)


Tensile/Source/client/source/Reference.cpp
```diff
@@ -173,6 +175,8 @@ namespace Tensile
                 }
             }
 
+            omp_set_dynamic(1);
```
nakajee, line 178: I suggest to put some comment here to understand the intention.

Tensile/Source/client/main.cpp
```diff
@@ -602,21 +602,36 @@ int main(int argc, const char* argv[])
...
-                                                                        warmupStartEvents[i],
-                                                                        warmupStopEvents[i]));
+                                                                        warmupStartEvents[0],
+                                                                        warmupStopEvents[0]));
                                 else
                                     HIP_CHECK_EXC(
                                         adapter.launchKernels(kernels, stream, nullptr, nullptr));
                                 listeners.postWarmup();
-                            }
+                                listeners.validateWarmups(inputs, warmupStartEvents, warmupStopEvents);
```
nakajee, line 618: You do not need to duplicate the code.
You can just put if (i==0) for validateWarmups.

nakajee:
Please fix the formatting error...
===========

605c605

<                             if(warmupInvocations > 0)

---

>                             if (warmupInvocations > 0)

618,619c618

<                                 listeners.validateWarmups(

<                                     inputs, warmupStartEvents, warmupStopEvents);

---

>                                 listeners.validateWarmups(inputs, warmupStartEvents, warmupStopEvents);

631,632c630,631

<                                         HIP_CHECK_EXC(adapter.launchKernels(

<                                             kernels, stream, nullptr, nullptr));

---

>                                         HIP_CHECK_EXC(

>                                             adapter.launchKernels(kernels, stream, nullptr, nullptr));

nakajee (approved):
Looks good. 
I do not have any further comment. Seems like no comments from others.

benjaminulmer (approved):
I don't really know enough about OpenMP to comment on those changes. Otherwise LGTM

Tensile/Source/client/source/Reference.cpp
```diff
@@ -175,9 +177,12 @@ namespace Tensile
                 }
             }
 
-            // Enable dynamic thread adjustment and set num threads to avoid defaulting to 1
+            // Set num threads and enable proc_bind to prevent main thread from migrating
             // Defaults cause validation to affect benchmark timing
-            omp_set_dynamic(1);
+            // Dynamic thread mode causes HostLibraryTests to fail
+            const char* env_bind = std::getenv("OMP_PROC_BIND");
```
nakajee, line 183: Do we ignore the case OMP_PROC_BIND is already defined and the value is not true?

AlexBrownAMD: If it's already defined, then we keep the original value. This only sets it if user didn't request a specific OMP_PROC_BIND.

nakajee: I am not sure if we need to care such case or not, but the issue might not be solved if OMP_PROC_BIND is defined as false.
Is that correct?

AlexBrownAMD: That's correct. I figured if someone manually set OMP_PROC_BIND to false then that is the behaviour they want. More likely they might set OMP_PROC_BIND=close or spread to keep threads from migrating, but affect distribution.

nakajee: Thanks for confirmation. That should be OK.

PR #1556, opened 2022-07-27
Add CleanupBuildFiles global param and ClientPath Tensile param

benjaminulmer:
Two small additions to help with Tuning/benchmarking automation and quality of life. 

If `CleanupBuildFiles` is true, TensileCreateLibrary will delete the build_tmp directory after use. I made this a global param since it affects all front ends that depend on TensileCreateLibrary, which is most of them.

The `ClientPath` param for Tensile allows for full control over the path to the benchmarking client. The current way we're dealing with building the client is sorta a mess and is due for a complete overhaul imo, so this is more of an interim solution. I'd like to do a follow up that merges this new `ClientPath` param with the existing `ClientBuildPath` and `PrebuiltClient` global params.

### DIFF
--- a/Tensile/BenchmarkProblems.py
+++ b/Tensile/BenchmarkProblems.py
@@ -29,7 +29,6 @@ import time
 
 from copy import deepcopy
 
-from . import ClientExecutable
 from . import SolutionLibrary
 from . import LibraryIO
 from . import Utils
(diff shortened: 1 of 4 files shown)


wbgilmartin (approved):
Greate

yenong-amd (changes requested):

yenong-amd (approved):

PR #1559, opened 2022-08-02
Adding all gemm data types to the efficiency script

babakpst:
With this PR, the efficiency calculator tracks/gets the gemm data types from Tensile DataType. This change allows calculating the efficiency for all data types, including C/Z/I8 which were missing previously.


nakajee (approved):
Looks good.

PR #1560, opened 2022-08-04
It is redundant to call syncThreads if the number of threads in a block is equal to the size of wavefront.

amgddm:
This modification remove software overhead if blocksize is equal to wavefrontsize. The LocalSplitU is used in optimizing GEMM performance for M=1. The best bloksize is equal to 64 (wavefrontsize) on MI100 and MI200. This modification can improve GEMM performance on MI200.  However,  performance changes are not observed on MI100.

### DIFF
--- a/Tensile/KernelWriterSource.py
+++ b/Tensile/KernelWriterSource.py
@@ -2757,7 +2757,7 @@ class KernelWriterSource(KernelWriter):
                 self.tileChar0, self.tileChar1, \
                 s, self.tileChar1, j, self.tileChar0, self.tileChar1, vc, i, s, \
                 self.tileChar0, j, self.tileChar0, self.endLine)
-    kStr += self.indent + self.syncStr + self.endLine
+    kStr += self.syncThreads(kernel);
     """
 
     kStr += "    /* print Local state */" + self.endLine
@@ -3311,7 +3311,13 @@ class KernelWriterSource(KernelWriter):
   # SyncThreads
   ##############################################################################
   def syncThreads(self, kernel, comment=""):
-    return self.indent + self.syncStr + " //" + comment + self.endLine
+    if globalParameters["CurrentISA"] == (9,0,8):
+      return self.indent + self.syncStr + " //" + comment + self.endLine
+    else:
+      if kernel["SubGroup0"]*kernel["SubGroup1"]*kernel["LocalSplitU"] != kernel["WavefrontSize"]:
+        return self.indent + self.syncStr + " //" + comment + self.endLine
+      else:
+        return self.indent + self.endLine
 
   ##############################################################################
   # MapAcctoArch


Tensile/KernelWriterSource.py
```diff
@@ -3311,7 +3311,13 @@ def wait(self, kernel, tPA, tPB, globalRead, localWrite, localRead, comment):
   # SyncThreads
   ##############################################################################
   def syncThreads(self, kernel, comment=""):
-    return self.indent + self.syncStr + " //" + comment + self.endLine
+    if globalParameters["CurrentISA"] == (9,0,8):
```
nakajee, line 3314: Why only 908 is special?

amgddm: @nakajee No performance improvement is observed on MI100. Actually, the GEMM performance degrades a little bit.

nakajee: Have you tried all architectures other than MI100,200?
Plus, is this logic really applicable for all syncThreads use cases?

nakajee: Shouldn't it be just for localSplitU?

amgddm: @nakajee, I don't have any other GPU type other than MI100 and MI200. This modification is correct and efficient theoretically. Because there is no software overhead. But it didn't show advantage on MI100 for some reasons.

amgddm: > Shouldn't it be just for localSplitU?

@nakajee. No, it has nothing to do with LocalSplitU. As long as the number of threads in a block is equal to the WaveFront size, any synThreads is not necessary since the synchronization is guaranteed by the hardware.

nakajee: > > Shouldn't it be just for localSplitU?
> 
> @nakajee. No, it has nothing to do with LocalSplitU. As long as the number of threads in a block is equal to the WaveFront size, any synThreads is not necessary since the synchronization is guaranteed by the hardware.

If the synchronization is guaranteed by HW, we should not need to check if it is MI100 or not.

amgddm: > 

@nakajee The reason why the MI100 is excluded with this change is not because it is not correct on MI100 and is because the performance is worse with this change (I think hipcc compile has an issue).

nakajee: I know.
Even though, we do not need such illogical code.
Have you checked the generated source code (and asm code as well)?
We need better understanding why MI100 does not perform better with this change.

nakajee: If we have confident on this change from logical point of view, we should apply to all architectures.
If not, we should apply this to the architectures which we verified (it means MI200 only).
It should not be just excluding MI100.

nakajee: (without verifying other architectures)

Tensile/KernelWriterSource.py
```diff
@@ -3311,7 +3311,13 @@ def wait(self, kernel, tPA, tPB, globalRead, localWrite, localRead, comment):
...
+      return self.indent + self.syncStr + " //" + comment + self.endLine
+    else:
+      if kernel["SubGroup0"]*kernel["SubGroup1"]*kernel["LocalSplitU"] != kernel["WavefrontSize"]:
```
nakajee, line 3317: I think it is better to follow the same way as KernelWriterAssembly.py
( logically, > should be better than !=)
if kernel["NumThreads"] > self.kernel["WavefrontSize"]

Tensile/KernelWriterSource.py
```diff
@@ -3311,7 +3311,13 @@ def wait(self, kernel, tPA, tPB, globalRead, localWrite, localRead, comment):
...
+        return self.indent + self.syncStr + " //" + comment + self.endLine
+      else:
+        return self.indent + self.endLine
```
nakajee, line 3320: Here, we can also use the same message as KernelWriterAssembly

Tensile/KernelWriterSource.py
nakajee: We can use kernel["WavefrontSize"] instead of self.kernel["WavefrontSize"]
(self. is redundant...)

Tensile/KernelWriterSource.py
nakajee: return "// Skip barrier: NumThreads=%s"%(kernel["NumThreads"]) + \
              comment + self.endLine

Tensile/KernelWriterSource.py
amgddm: > 

@nakajee I am not sure if kernel["NumThreads"] includes the value of LocalSplitU.

amgddm: > > 
> 
> @nakajee I am not sure if kernel["NumThreads"] includes the value of LocalSplitU.

@nakajee I checked the code again, kernel["NumThreads"] includes parameter LocalSplitU.

Tensile/KernelWriterSource.py
nakajee: I prefer not to check if it is MI100 or not because the same function in assembly does not have such condition.

amgddm: > I prefer not to check if it is MI100 or not because the same function in assembly does not have such condition.

@nakajee I don't think anyone compare the performance w/wo synThreads in assembly on MI100. I gave you an example, The memory frequency on MI100 is 1087MHz. The GEMM performance is 404 Gflops with synThreads on MI100 (numThreads=64) for m=1, N=2048, K=9217. However, the performance is 192 GFlops without synThreads on MI100 (numThreads=64). The performance on MI200 is another story. The performance is 363Gflops with synThreads on MI200. The performance is 372.683GFlops without synThreads. The memory clock frequency is 985MHz.

nakajee: Have you check the generated code (source and asm(converted from source))?
It is something wrong.

Another thing is asm generator side already cover this sync optimization (without excluding MI100).
Could you check if that perf drop happens with asm generator?

At least, I won't accept the current if condition without verifying all other architectures.
If you cannot verify other architectures, we should change the output code only for the verified architecture.

Before judging this, I would like to know more detail about the MI100 perf drop.

amgddm: > Have you check the generated code (source and asm(converted from source))? It is something wrong.
> 
> Another thing is asm generator side already cover this sync optimization (without excluding MI100). Could you check if that perf drop happens with asm generator?
> 
> At least, I won't accept the current if condition without verifying all other architectures. If you cannot verify other architectures, we should change the output code only for the verified architecture.
> 
> Before judging this, I would like to know more detail about the MI100 perf drop.

@nakajee @bragadeesh @babakpst @AlexBrownAMD @benjaminulmer @msujon-AMD  I don't know asm generator has covered this issue before. Basically, modification in source and asm did the similar thing. Those changes have nothing to do with any specific architectures. By the way, I have verify the correctness of the results with rocblas. The person checked in the assembly must also know GPGPU very well. I don't know if he has tested the correctness of the code fragment on all architecture. I think he did not. It is not necessary since the GPU architecture guarantee it. It is common practice for anyone knows OpenCL/CUDA/HIP programming.  

I agree that I need to check if assembly part decrease the performance on MI100. But it doesn't affect the fact that hip's performance is worse with the change on MI100. That is the reason I put conditional statement to exclude MI100. Probably it is not true for the assembly on MI100. If the performance drops with assembly, we also need to add   a conditional statement to exclude MI100. My assumption is that no one tested that fragment before since NumThreads = 128, 256 are widely used in most cases. 64 is seldom used in practice. 

I also agree that we need to understand why the GEMM performance drops on MI100 without synchroonizattion. It shouldn't theoretically since no additional software overhead. I guess hipcc compiler generates inefficient code on MI100.

nakajee: > @nakajee @bragadeesh @babakpst @AlexBrownAMD @benjaminulmer @msujon-AMD I don't know asm generator has covered this issue before. Basically, modification in source and asm did the similar thing. Those changes have nothing to do with any specific architectures. By the way, I have verify the correctness of the results with rocblas. The person checked in the assembly must also know GPGPU very well. I don't know if he has tested the correctness of the code fragment on all architecture. I think he did not. It is not necessary since the GPU architecture guarantee it. It is common practice for anyone knows OpenCL/CUDA/HIP programming.

I checked the history of syncThreads (for both KernelWriterAssembly and KernelWriterSource).
That if condition (numThreads > 64) was added in 2017 (much older than MI100).
As you say, probably, no one verified your scenario before.

> I agree that I need to check if assembly part decrease the performance on MI100. But it doesn't affect the fact that hip's performance is worse with the change on MI100. That is the reason I put conditional statement to exclude MI100. Probably it is not true for the assembly on MI100. If the performance drops with assembly, we also need to add a conditional statement to exclude MI100. My assumption is that no one tested that fragment before since NumThreads = 128, 256 are widely used in most cases. 64 is seldom used in practice.

My point here is we should use the same condition for both asm generator and source generator if performance drops a lot.

> I also agree that we need to understand why the GEMM performance drops on MI100 without synchroonizattion. It shouldn't theoretically since no additional software overhead. I guess hipcc compiler generates inefficient code on MI100.

Could you provide some more clues instead of just guessing?

Another aspect is the accuracy of the performance result.
It might be better to integrate the following fix to make sure your benchmark result is more accurate.
  Fix benchmark timings when using warmups and/or validation #1555
This will be merged into develop soon.

nakajee: One more thing is that we do not want to see unexpected performance drop on old architectures.
If we do not want to verify old architectures, we do not need to change the behavior of old architectures.
Based on the current result on MI100/200, the improvement on MI200 is small, but the drop on MI100 is huge.

amgddm: > One more thing is that we do not want to see unexpected performance drop on old architectures. If we do not want to verify old architectures, we do not need to change the behavior of old architectures. Based on the current result on MI100/200, the improvement on MI200 is small, but the drop on MI100 is huge.

Yes, it is better to perform experiments on old architectures to make sure the performance doesn't drop on old architecture.

amgddm: > One more thing is that we do not want to see unexpected performance drop on old architectures. If we do not want to verify old architectures, we do not need to change the behavior of old architectures. Based on the current result on MI100/200, the improvement on MI200 is small, but the drop on MI100 is huge.

The performance improvement is small, which is expected since the additional overhead is small.

Tensile/KernelWriterSource.py
amgddm: @nakajee updated code. The poor performance on MI100 is caused from old compiler. A newer compiler like rocm 5.2 doesn't have this issue. I removed conditional statement checking against (9,0,8) (MI100.

nakajee: OK. Please also check other suggestions I put.

(1) please use the condition kernel["NumThreads"] > self.kernel["WavefrontSize"]
  use > instead of !=
  use NumThreads
  
  (2)we can also use the same message as KernelWriterAssembly

amgddm: @nakajee Done.

amgddm: @nakajee use the same comments in KernelWriteAssembly.py.

Tensile/KernelWriterSource.py
```diff
@@ -3311,7 +3311,11 @@ def wait(self, kernel, tPA, tPB, globalRead, localWrite, localRead, comment):
   # SyncThreads
   ##############################################################################
   def syncThreads(self, kernel, comment=""):
-    return self.indent + self.syncStr + " //" + comment + self.endLine
+    if kernel["NumThreads"] > kernel["WavefrontSize"]:
+        return self.indent + self.syncStr + " //" + comment + self.endLine
+    else:
+        return self.indent + "// Skip barrier: NumThreads=%s"%(kernel["NumThreads"]) + \
+              self.indent + self.endLine
```
nakajee, line 3318: Do we need self.indent again here?

amgddm: Yes. Otherwise, the comment line has different indent. @nakajee

nakajee: I know, but second self.indent seems not necessary before self.endLine.
We do not need any indent for endLine, right?

amgddm: @nakajee right. I forgot removing it. Done.

nakajee (approved):
Thanks for your update.
Please make sure that CI test does not fail due to your change.

PR #1561, opened 2022-08-04
Make CI build HostLibraryTests in release mode temporarily

AlexBrownAMD:
Workaround for an issue causing long build times on certain files until CI is updated to a version with a fix

### DIFF
--- a/.jenkins/common.groovy
+++ b/.jenkins/common.groovy
@@ -32,7 +32,9 @@ def runCompileCommand(platform, project, jobName, boolean debug=false)
     String compiler = 'hipcc'
     String pythonVersion = 'py36'
     String cov = "V3"
-    String buildType = debug ? 'Debug' : 'RelWithDebInfo'
+    // Do release build of HostLibraryTests on CI until it is upgraded to rocm 5.3 to
+    // avoid bug causing long build times of certain files.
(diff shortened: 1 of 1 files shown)


.jenkins/common.groovy
```diff
@@ -32,7 +32,9 @@ def runCompileCommand(platform, project, jobName, boolean debug=false)
     String compiler = 'hipcc'
     String pythonVersion = 'py36'
     String cov = "V3"
-    String buildType = debug ? 'Debug' : 'RelWithDebInfo'
+    // Do release build of HostLibraryTests on CI until it is upgraded to rocm 5.3 to
+    // avoid bug causing long build times of certain files.
+    String buildType = 'Release' // debug ? 'Debug' : 'RelWithDebInfo'
```
nakajee, line 37: Just confirmation.
Both Debug and RelWithDebInfo cause the long build time issue?

benjaminulmer (approved):
The precheckin gfx908 compile fails looks like it's an issue with rocm-smi on the node it ran on. The compile time on gfx906 was only 17 minutes, so it looks like this did fix the issue.

.jenkins/common.groovy
AlexBrownAMD: Yea, any build that generates debug info hangs.

PR #1563, opened 2022-08-05
Updated decision trees

nielenventer:
Updates dtrees:
- No explicit return node (child idx indicates return)
- Clearer tree property names
- MLFeature properties (return float)


Tensile/Source/lib/include/Tensile/ContractionProblemProperties.hpp
```diff
@@ -27,6 +27,7 @@
 #pragma once
 
 #include <Tensile/ContractionProblem.hpp>
+#include <Tensile/ContractionSolution.hpp>
```
benjaminulmer, line 30: am I just missing why this include needed to be added?

benjaminulmer (commented):
Overall looks good. My one piece of feedback would be it seems there's some overlap in functionality between the calculations you do in the new functions in ContractionSolution (numTiles0, numTiles1, and tileGranularity) and the existing computeGranularities. Did you look at all into leveraging the existing code here?

Tensile/Source/lib/include/Tensile/ContractionProblemProperties.hpp
nielenventer: good catch, that was a relic of when the feature properties were in this file

nielenventer:
Made a mess by PRing my dev instead of just the trees branch, updated PR here:
https://github.com/ROCmSoftwarePlatform/Tensile/pull/1568

Tensile/Source/lib/include/Tensile/ContractionProblemProperties.hpp
nielenventer: R.e. overlap, I removed the tile calc functions (`numTiles<x>`) completely, they were too specific to MLFeature, and actually it seemed clearer to just have no function. `tileGranularity` was changed to `computeGranularity` and used in a few more places.

PR #1566, opened 2022-08-16
Option to output library logic in dict format

benjaminulmer:
#1534 added support for a new dict format library logic. This PR adds the ability for Tensile to output library logics in the new format. This is part of an effort to eventually move the rocBLAS library logic files to this new format.

`globalParameters["DictLibraryLogic"] = True` is included for CI to test the new format; the default will be set to False before merging. If/when the library files in rocBLAS are updated to this new format the default will be set to True again.

### DIFF
--- a/Tensile/LibraryIO.py
+++ b/Tensile/LibraryIO.py
@@ -67,7 +67,7 @@ def writeYAML(filename, data, **kwargs):
         kwargs["default_flow_style"] = None
 
     with open(filename, "w") as f:
-        yaml.dump(data, f, **kwargs)
+        yaml.dump(data, f, sort_keys=False, **kwargs)
 
 
 def writeMsgPack(filename, data):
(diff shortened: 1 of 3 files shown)


yenong-amd (approved):

Tensile/LibraryIO.py
```diff
@@ -43,6 +43,12 @@
     print("Message pack python library not detected. Must use YAML backend instead.")
 
 
+def updateProblemDatatypes(problemType):
```
nielenventer, line 46: nitpick, but not really a fan of this function, since it doesn't get reused and has a very opaque name it's just hiding what's happening. What about having something like

```
for dtype_key in ["DataType", "DestDataType", "ComputeDataType"]:
    solutionState["ProblemType"][dtype_key] = \
                    solutionState["ProblemType"][dtype_key].value
```

Not in a function but just at line 92. That way you're still grouping them, but it's clear what's happening

benjaminulmer: Maybe you just missed it, but this is called in five places across the various functions. I agree it's a bit opaque, but I thought it was the lesser of the two evils compared to the code duplication.

nielenventer: Oh I totally did miss that sorry

nielenventer (approved):

PR #1567, opened 2022-08-16
"No solution" error message

AlexBrownAMD:
Add an error message to client when no solution is found
Update example used on main wiki page to remove LibraryClient step (fails unless run on target hardware)


benjaminulmer (approved):
lgtm.
I'll try to take a look at autofilling the architecture name (as per the discussion earlier today) in a future PR.

yoichiyoshida (approved):

PR #1568, opened 2022-08-19
Updated decision trees for solution selection

nielenventer:
Updates dtrees:

    No explicit return node (child idx indicates return)
    Clearer tree property names
    MLFeature properties (return float)


benjaminulmer (commented):
Nice, much less baggage in ContractionSolution now. Thanks for cleaning things up.

I might take a closer look on Monday, but LGTM so far besides the static analysis failures (and maybe a more descriptive PR title? :P )

benjaminulmer (approved):

PR #1569, opened 2022-08-22
Lazy loading hotfix

JeremyAdamHart:
Failures were detected when using multiple GPUs and using lazy loading. The source of these errors were narrowed down to a race condition that occurs when accessing/modifying codeObjectFilename in the Solution object when multiple threads are accessing the same solution. I've introduced a new class to wrap that variable and protect all instances where it is being assigned to or read from. The failure hasn't been observed with these changes, so I'm confident it fixes the issue.

It was also found that m_codeObjectDirectory could result in a race condition, I've moved the assignment and reading of that variable into a mutex protected area.

### DIFF
--- a/Tensile/Source/lib/include/Tensile/ContractionSolution.hpp
+++ b/Tensile/Source/lib/include/Tensile/ContractionSolution.hpp
@@ -37,6 +37,7 @@
 #include <Tensile/ContractionProblem_fwd.hpp>
 #include <Tensile/DataTypes.hpp>
 #include <Tensile/Predicates.hpp>
+#include <Tensile/Utils.hpp>
 
 namespace Tensile
 {
@@ -286,11 +287,11 @@ namespace Tensile
             double max       = 1000.0;
         };
 
-        int         index = 0;
-        std::string kernelName;
-        std::string codeObjectFilename;
-        bool        debugKernel   = false;
-        bool        kernelArgsLog = false;
+        int                          index = 0;
+        std::string                  kernelName;
+        ThreadSafeValue<std::string> codeObjectFilename;
+        bool                         debugKernel   = false;
+        bool                         kernelArgsLog = false;
 
         std::shared_ptr<Predicates::Predicate<Problem>> problemPredicate
             = std::make_shared<Predicates::True<Problem>>();
--- a/Tensile/Source/lib/include/Tensile/Utils.hpp
+++ b/Tensile/Source/lib/include/Tensile/Utils.hpp
@@ -37,6 +37,59 @@
 namespace Tensile
 {
 
+    // Type wrapper that can be copied or assigned to in a threadsafe manner. Value cannot be modified
+    template<typename T>
+    struct ThreadSafeValue
+    {
+    private:
+        mutable std::mutex access;
+        T                  value;
+    public:
+        ThreadSafeValue()
+        {
+
+        }
+
+        ThreadSafeValue(const ThreadSafeValue<T>& other)
+        {
+            std::lock_guard<std::mutex> lock (other.access);
+            value = other.value;
+        }
+
+        ThreadSafeValue(const T& other): value(other)
+        {
+
+        }
+
+        ThreadSafeValue<T>& operator=(const ThreadSafeValue<T>& other)
+        {
+            std::lock_guard<std::mutex> otherLock (other.access);
+            std::lock_guard<std::mutex> selfLock (access);
+            value = other.value;
+
+            return *this;
+        }
+
+        ThreadSafeValue<T>& operator=(const T& other)
+        {
+            std::lock_guard<std::mutex> lock (access);
+            value = other;
+
+            return *this;
+        }
+
+        T copy() const
+        {
+            std::lock_guard<std::mutex> lock(access);
+            return value;
+        }
+
+        T operator*() const
+        {
+            return copy();
+        }
+    };
+
     /**
  * \ingroup Tensile
  * \addtogroup Utilities
--- a/Tensile/Source/lib/include/Tensile/Utils.hpp
+++ b/Tensile/Source/lib/include/Tensile/Utils.hpp
@@ -38,12 +38,13 @@ namespace Tensile
 {
 
     // Type wrapper that can be copied or assigned to in a threadsafe manner. Value cannot be modified
+    // Intended for using value semantics with non-trivially copyable data
     template<typename T>
-    struct ThreadSafeValue
+    class ThreadSafeValue
     {
     private:
-        mutable std::mutex access;
-        T                  value;
+        mutable std::mutex m_access;
+        T                  m_value;
     public:
         ThreadSafeValue()
         {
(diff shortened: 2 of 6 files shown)


benjaminulmer (approved):
The full threadsafe wrapper around the filename is a bit heavy handed in my opinion, but it would protect against future mistakes so I understand the motivation. We can potentially revisit in the future.

Bleeding edge CI is having some issue, but none of the changes in this PR should really depend on the ROCm version so regular precheckin is fine by me.  

Since this is a hotfix we'll need approval from @bragadeesh

JeremyAdamHart:
@benjaminulmer I originally tried adding a mutex to ContractionSolution, but we are copying it in several places so I saw several options:
- Write a copy and assignment operator for the ContractionSolution class
- See if I can reorganize the code to eliminate the copies (might require a move constructor)
- Make a wrapper for the problem variable

It might be a little heavy handed, but since this is the only thing in that class being accessed in parallel, it has the smallest impact on the code itself, and makes it more certain that it can't cause issues even if you refer to it elsewhere in the future. That's my reasoning at least

bragadeesh (approved):

Tensile/Source/lib/include/Tensile/Utils.hpp
```diff
@@ -37,6 +37,59 @@
...
+
+        ThreadSafeValue(const T& other): value(other)
+        {
+
+        }
+
+        ThreadSafeValue<T>& operator=(const ThreadSafeValue<T>& other)
+        {
+            std::lock_guard<std::mutex> otherLock (other.access);
+            std::lock_guard<std::mutex> selfLock (access);
```
cgmb, line 67: You need to guard against self-assignment. This will deadlock if `this == &other`.

Tensile/Source/lib/include/Tensile/ContractionSolution.hpp
```diff
@@ -286,11 +287,11 @@ namespace Tensile
...
         };
 
-        int         index = 0;
-        std::string kernelName;
-        std::string codeObjectFilename;
-        bool        debugKernel   = false;
-        bool        kernelArgsLog = false;
+        int                          index = 0;
+        std::string                  kernelName;
+        ThreadSafeValue<std::string> codeObjectFilename;
```
cgmb, line 292: What's the difference between this and `std::atomic<std::string>`?

cgmb (changes requested):

Tensile/Source/lib/include/Tensile/ContractionSolution.hpp
cgmb: Oh. `ThreadSafeValue<std::string>` is copyable while `std::atomic<std::string>` is not. I'd suggest explicitly contrasting the two in the documentation.

Tensile/Source/lib/include/Tensile/Utils.hpp
```diff
@@ -37,6 +37,59 @@
...
+            value = other;
+
+            return *this;
+        }
+
+        T copy() const
+        {
+            std::lock_guard<std::mutex> lock(access);
+            return value;
+        }
```
cgmb, line 85: The equivalent function in the standard library is called [`load`](https://en.cppreference.com/w/cpp/atomic/atomic/load).

Tensile/Source/lib/include/Tensile/Utils.hpp
```diff
@@ -37,6 +37,59 @@
...
+    private:
+        mutable std::mutex access;
+        T                  value;
+    public:
+        ThreadSafeValue()
+        {
+
+        }
+
+        ThreadSafeValue(const ThreadSafeValue<T>& other)
```
cgmb, line 53: You've redefined the copy constructor and the assignment operator, but have not defined a destructor. This breaks the [rule of three](https://en.cppreference.com/w/cpp/language/rule_of_three).

cgmb:
If the semantics of `std::unique_ptr<std::atomic<std::string>>` are sufficient, that might be worth considering as an alternative to `ThreadSafeValue<std::string>`. It's not copyable, but it is movable.

Tensile/Source/lib/include/Tensile/ContractionSolution.hpp
JeremyAdamHart: You need it to be a trivially copyable type to use atomic. The alternative would be using an atomic pointer and adding copy/assignment/destructors/etc to delete it properly

TorreZuk (commented):
Did you reproduce the crash to codeOjbectFilename access to get the full stack?   I would suggest you make a ticket to continue the refactor as just wrapping it in this helper is not clear of what initialization ordering requirements are being violated.  If the value is changed by another thread at the end of a generate call to me is what I call "surprise".    I repeat my advice this isn't really a struct and use the m_ prefix to make it clearer where accessors would be better code.  A single mutex managing resources may be clear than a helper wrapper on single variables.  I only see fixes to avoid a crash in this PR whereas I would assume some design problem, i.e. wouldn't the lock be best to acquire and release only after the code object is loaded to match the string?  Two threads can queue up to load the same codeObject, so saying guard the resource allocation process rather than the variables.

Tensile/Source/lib/include/Tensile/Utils.hpp
JeremyAdamHart: That makes more sense, I'll change the name to load

cgmillette (commented):
Recalling our previous discussion, this related code [here](https://github.com/ROCmSoftwarePlatform/Tensile/blob/dcb5142dc4a4145114ea018c1b3c70c9682f322f/Tensile/Source/lib/include/Tensile/PlaceholderLibrary.hpp#L118) throws a few red flags.

- local member 'library' is written under mutex protection in loadPlaceholderLibrary() 
but is read in several places without mutex protection.

- [this code](https://github.com/ROCmSoftwarePlatform/Tensile/blob/dcb5142dc4a4145114ea018c1b3c70c9682f322f/Tensile/Source/lib/include/Tensile/PlaceholderLibrary.hpp#L143
) local variable 'solutions' shadows struct member 'solutions', which becomes ambiguous. Access to this is also not guarded by its mutex, if this is what is intended.

- Is there only ever one Placeholder library created? Is this there a managed static instance to this class? If not, then we might consider a static master library mutex to ensure that the same mutex is protecting a shared unique instance of the master library.

I agree with Torre that this patch might address the symptoms of an apparent race condition, but it feels like there could be a deeper design issue with flow control in loading the library/solutions.

JeremyAdamHart:
@TorreZuk The crashes were messy so we weren't able to examine the stack unfortunately. We did work out a couple different ways that it could occur though, one situation involved a race condition within placeholderLibrary when assigning to codeObjectFilename, the other was between placeholderLibrary and the generateXCall function. The second happening less frequently. The ideal solution in my mind would be to have codeObjectFilename set during the deserealization process of TensileLibrary.dat, but that is more of an involved process and requires more consideration. It's a change I plan on trying to make on the develop branch, but my primary goal for this PR was to have a fix that deals with the issues and has as small of a chance of introducing new problems as possible.

JeremyAdamHart:
> If the semantics of `std::unique_ptr<std::atomic<std::string>>` are sufficient, that might be worth considering as an alternative to `ThreadSafeValue<std::string>`. It's not copyable, but it is movable.

@cgmb Unfortunately that still gets a template error as far as I can tell. It looks like there's a specialization for shared_ptr in c++20 that can allow std::atomic<std::shared_ptr<T>>, but I'm not sure whether or not T still needs to be a trivially copyable value

benjaminulmer:
Full rocBLAS build and test passed using [cf8a6d0](https://github.com/ROCmSoftwarePlatform/Tensile/pull/1569/commits/cf8a6d08c8957191aae52eaec91c862612bb1340). Jeremy did some sanity checks on the latest commit; I don't think the changes warrant another full build and test. 

I agree with Torre that this is a bit of a bandaid and we should revisit in the future. However, I do believe this properly addresses the issue, at least with regards to the library structures we actually generate and how we interface with them in rocBLAS. All these issues will disappear if we can figure out how to set this properly at deserialization time, so that should be where we focus our efforts in my opinion. 

If/when precheckin passes and Cory (and others) approve the changes we can go ahead with the merge and follow the rest of the Tensile/rocBLAS hotfix process (ping me if you have questions).

cgmb:
> > If the semantics of `std::unique_ptr<std::atomic<std::string>>` are sufficient, that might be worth considering as an alternative to `ThreadSafeValue<std::string>`. It's not copyable, but it is movable.
> 
> @cgmb Unfortunately that still gets a template error as far as I can tell.

Ah. Too bad. I'd thought the mutex-based fallbacks for `std::atomic` could handle arbitrary types, but it seems I was mistaken.

The underlying concern was basically that the semantics of a type like `ThreadSafeValue<std::string>` are very complex. It is extraordinarily difficult to do this sort of fine-grained parallelism correctly, and doubly so when you're defining your own custom abstractions. It would be nice over the long term to find a pathway towards a simpler solution.

cgmb (approved):
I don't have enough knowledge of Tensile to be a real judge of this change overall, so my review was focused on very narrow issues. The self-assignment in `ThreadSafeValue` was my primary concern and that has been fixed.

PR #1571, opened 2022-08-23
Make two-tier solution selection batch aware

benjaminulmer:
Solution selection is not currently batch aware due to runtime considerations of the distance matching algorithm. The primary tier equality matching in two-tier solution selection, however, is more limited meaning additional keys can be added with very little cost to the runtime of matching.

Adding batch count to the primary tier will allow us to tune problems with the same M, N, and K at different batch counts (e.g. SWDEV-343413) by storing the tuning results into the primary tier.

### DIFF
--- a/Tensile/SolutionLibrary.py
+++ b/Tensile/SolutionLibrary.py
@@ -82,9 +82,10 @@ class MatchingLibrary:
         propertyKeys = {
             2: Properties.Property("FreeSizeA", index=0),
             3: Properties.Property("FreeSizeB", index=0),
-            #0: Properties.Property("BatchSize", index=0),
             1: Properties.Property("BoundSize", index=0)
         }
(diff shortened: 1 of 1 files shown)


babakpst (commented):
Currently, we keep batched and not-batched kernels in the same lib logic. How does this PR change our lib logic? Based on the description, I inferred that we should separately tune for the batched sizes. Is that correct?

benjaminulmer:
@babakpst This PR won't have any effect on our existing lib logic files. For now, the only situation where the primary tier will need to be used is if we need tuning results for GEMMs with the _same_ M, N, and K but _different_ batch count values; those cases go into a separate library logic file that will make up the primary tier.

babakpst (approved):
Looks good. Thanks.

yenong-amd (approved):

PR #1575, opened 2022-08-25
hotfix (prototype): using hipModuleLoadData instead of hipModuleLoad

benjaminulmer:
Prototype of a fix for SWDEV-348341

### Cause of the crash
`hipModuleLoad` doesn't release file descriptors after loading. Since the new lazy-loading splits rocBLAS kernels into many more code object files than before, there are enough modules that if rocBLAS loads all the gfx90a ones on multiple GPUs the process will reach the limit for open file descriptors.

This behaviour of `hipModuleLoad` is intended and won't be changed (see SWDEV-353036 and SWDEV-304151). Thus, a workaround is up to us.

### Potential solutions
**1) Reduce the number of code object files lazy-loading splits kernels into**
Medium dev effort and only side steps the issue---the issue could still show up again in the future if the number of GPUs increases or file descriptors are needed for other purposes (e.g. other libraries loading modules, application needing files).

**2) Manage modules and unload modules using some heuristics when necessary**
Very high dev effort: not feasible for a quick fix. Might be worth exploring in the future.

**3) Use `hipModuleLoadData`**
This is the solution prototyped in this PR. This API also has an undocumented gotcha where the original memory buffer must outlive the module; the code to meet this requirement is also included in this PR.

### Feedback requested
We've tested this fix and it resolves the issue in SWDEV-348341, so it seems this is functionally sound. If possible, I'd like to know the impact this has on rocBLAS library initialization times (@JeremyAdamHart any help here would be appreciated). I'd also like to know if anyone has ideas for a more elegant way to accomplish what I've roughed out. Finally, I want to know if there are any unforeseen consequences that might result from these changes.

### DIFF
--- a/Tensile/Source/lib/source/hip/HipSolutionAdapter.cpp
+++ b/Tensile/Source/lib/source/hip/HipSolutionAdapter.cpp
@@ -28,6 +28,7 @@
 #include <hip/hip_runtime.h>
 
 #include <cstddef>
+#include <fstream>
 
 #include <Tensile/Debug.hpp>
 #include <Tensile/EmbeddedData.hpp>
@@ -83,7 +84,20 @@ namespace Tensile
         {
             hipModule_t module;
 
-            HIP_CHECK_RETURN(hipModuleLoad(&module, path.c_str()));
+            // load file into buffer
+            std::ifstream coFile(path, std::ifstream::binary);
+            coFile.seekg(0, coFile.end);
+            auto length = coFile.tellg();
+            coFile.seekg(0, coFile.beg);
+
+            auto buff = std::make_unique<char[]>(length);
+            coFile.read(buff.get(), length);
+
+            // use moduleLoadData
+            HIP_CHECK_RETURN(hipModuleLoadData(&module, (void*)buff.get()));
+
+            // keep buffer around
+            m_moduleBuffers.push_back(std::move(buff));
 
             if(m_debug)
                 std::cout << "loaded code object " << path << std::endl;
--- a/Tensile/Source/lib/include/Tensile/hip/HipSolutionAdapter.hpp
+++ b/Tensile/Source/lib/include/Tensile/hip/HipSolutionAdapter.hpp
@@ -87,6 +87,7 @@ namespace Tensile
             std::mutex m_access;
 
             std::vector<hipModule_t>                       m_modules;
+            std::vector<std::unique_ptr<char[]>>           m_moduleBuffers;
             std::unordered_map<std::string, hipFunction_t> m_kernels;
             bool                                           m_debug           = false;
             bool                                           m_debugSkipLaunch = false;


TorreZuk:
I just commented to Ruban I made a simple multi-thread test on hayabusa 16 GPU and got this error (rocblas bench calling gemm_ex on each thread using a different GPU).  If it helps I can push my test code into rocblas branch.
rocBLAS info: Time taken to complete rocBLAS library initialization is 0 milliseconds.
OMP: Error #179: Function Can't open SHM2 failed:
OMP: System error #24: Too many open files
Sounds like the same problem.

benjaminulmer:
@TorreZuk Definitely sounds like the same issue to me. Would you mind rebuilding with this Tensile patch and confirming if it fixes it or not? Alternatively, you could push the code and I can run the test.

TorreZuk:
> @TorreZuk Definitely sounds like the same issue to me. Would you mind rebuilding with this Tensile patch and confirming if it fixes it or not? Alternatively, you could push the code and I can run the test.

I can clean my test code up enough to push to a PR branch off current develop shortly.  I can try your tensile commit as well.

TorreZuk:
My rocblas PR 1411 you can use to test against, I will also try against this hotfix but probably need the other hotfix as well.  Note I initialize in each thread and was seeing crashes with subtle variations in initializations but the current one reports the too many files open when testing on 15 GPUs.

benjaminulmer:
Thanks, I'll do some testing today. This branch has both hotfixes, so hopefully building with it fixes all the issues.

Tensile/Source/lib/source/hip/HipSolutionAdapter.cpp
```diff
@@ -83,7 +84,20 @@ namespace Tensile
         {
             hipModule_t module;
 
-            HIP_CHECK_RETURN(hipModuleLoad(&module, path.c_str()));
+            // load file into buffer
+            std::ifstream coFile(path, std::ifstream::binary);
```
JeremyAdamHart, line 88: You're missing a test to check if the ifstream is valid, this causes a crash in the Tensile tuning client

JeremyAdamHart (changes requested):
I think this is a good way forward, but I also want to double check that kernel debugging is still possible with this. If there ends up being issues with that, maybe we should have this version as a build option and still allow use of hipModuleLoad

Tensile/Source/lib/source/hip/HipSolutionAdapter.cpp
benjaminulmer: I see, the client is only supposed to error only if _all_ code objects fail to load (although it looks like there's a bug in the implementation of this, but that's beside the point).

How does something like this sound?
```suggestion
            std::ifstream coFile(path, std::ifstream::binary);
            if (coFile) {
                // do the stuff
            }
            else
            {
                return hipErrorFileNotFound;
            }
```
I think this would mimic the previous behaviour

JeremyAdamHart: That's exactly what I did to test it your changes, looks good to me!

benjaminulmer:
> I think this is a good way forward, but I also want to double check that kernel debugging is still possible with this. If there ends up being issues with that, maybe we should have this version as a build option and still allow use of hipModuleLoad

Based on my understanding debugging should still work, it just might be a bit less convenient (not sure in what ways exactly...). In either case though, debugging Tensile kernels isn't a likely use case for users, so I'm not too worried about potentially breaking that in the release branch.

I'd like to turn this into a build option of some kind when we bring this back into develop. Doing so in the hotfix itself though felt out of scope and would delay getting the fix in place.

TorreZuk (commented):
As currently multi-thread usage can have regresssions where applications will now run out of file handles think this should get into develop soon as will see failure reports.

AlexBrownAMD:
Just catching up on the history/related tickets on this PR. Have you checked if rocgdb still works with this change? Definitely important that we don't lose debug support.

Tensile/Source/lib/source/hip/HipSolutionAdapter.cpp
```diff
@@ -81,9 +82,27 @@ namespace Tensile
...
+            // running out of descriptors/handles. Use hipModuleLoadData as a workaround
+            if(coFile)
+            {
+                coFile.seekg(0, coFile.end);
+                auto length = coFile.tellg();
+                coFile.seekg(0, coFile.beg);
 
-            HIP_CHECK_RETURN(hipModuleLoad(&module, path.c_str()));
+                buffer = std::make_unique<char[]>(length);
+                coFile.read(buffer.get(), length);
```
cgmb, line 98: What happens if any of these operations fail?

benjaminulmer: If my understanding of `std::ifstream` is correct, these methods won't throw unless you set the `exceptions()` mask to tell it to. So if something goes wrong here and the buffer doesn't get filled with the correct data, we should still make it to the `hipModuleLoadData` call which will then return whatever hip error is relevant.

benjaminulmer:
@TorreZuk 
> As currently multi-thread usage can have regresssions where applications will now run out of file handles think this should get into develop soon as will see failure reports.

Yes I didn't mean to imply that the build flag should gate the initial sync with develop, just that it's something we should develop afterwards.

@AlexBrownAMD 
> Just catching up on the history/related tickets on this PR. Have you checked if rocgdb still works with this change? Definitely important that we don't lose debug support.

I worked with Alex offline on this and we were able to confirm rocgdb works with the change.

AlexBrownAMD (approved):

JeremyAdamHart (approved):
The code object file load time doesn't seem to be having a noticeable impact, lgtm

bragadeesh (approved):

PR #1580, opened 2022-08-31
Fix CI issue on gfx1030 temporarily

amgddm:
Comment out the conditional statement in function syncThreads. The WavefrontSize is 32 on gfx1030.

In SolutionStructs.py, the following statement cause CI issue on gfx1030 because the
WavefrontSize is 32 on gfx1030.


benjaminulmer (commented):
It seems the plan is to revisit this more properly in the future?
If so, could you add a TODO comment with a brief explanation in the meantime? I don't like commented out code being merged, but at least with a TODO comment I can stomach it.

benjaminulmer (approved):
Approved pending CI and rocBLAS tests on gfx1030 passing

PR #1581, opened 2022-09-01
use _putenv_s to replace setenv in windows

TonyYHsieh:
To support Tensile run on Windows
use _putenv_s to replace setenv in windows
use _putenv to replace unsetenv in windows


TorreZuk (approved):
Sure but in setting up CI I would still ask them to put this in the environment variables they set so it is clearly exposed

PR #1589, opened 2022-09-12
ROCM 5.4 merge master into develop

benjaminulmer:


amcamd (approved):
Checked version bump from 4.34 to 4.35

TorreZuk (approved):
LGTM, same old question why develop wouldn't move forward right now to not match master/release branch, but this is fine.

PR #1590, opened 2022-09-14
Clean up LLVM dependency code

AlexBrownAMD:
Remove searches for specific versions of llvm. Tensile can be built using LLVM bundled with ROCm as long as other dependencies are met. I'll update the wiki with new instructions. Separate LLVM installations still work, and cmake seems to prefer finding stand-alone LLVM libs, with the ROCm one being a fallback.

Also added a tag to fix some pytest warnings.

### DIFF
--- a/HostLibraryTests/CMakeLists.txt
+++ b/HostLibraryTests/CMakeLists.txt
@@ -139,13 +139,7 @@ set(test_sources ${test_sources}
 )
 
 if(TENSILE_USE_LLVM)
-    find_package(LLVM 13.0 QUIET CONFIG)
-    if(NOT LLVM_FOUND)
-        find_package(LLVM 12.0 QUIET CONFIG)
-        if(NOT LLVM_FOUND)
-            find_package(LLVM REQUIRED CONFIG)
-        endif()
-    endif()
+    find_package(LLVM REQUIRED CONFIG)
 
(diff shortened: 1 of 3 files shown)


nakajee:
I still see 1 CI test failed.
Is that expected?

nakajee:
Is there any reason why the specific llvm version is searched before?
This looks simpler and better solution, but I would like to know why we did not implement in this way before.

AlexBrownAMD:
The failures I see so far on CI are also present on other PRs / mainline builds - likely due to current ROCm version on CI (just a few left to triage).

I think the specific version searches were there to try and get it to pick the LLVM bundled with ROCm over the recomended llvm6.0 dependency, but they haven't been updated as our ROCm build shifts to newer versions of LLVM. But in these cases if either 12 or 13 were present on the system, it would still pick those over the bundled LLVM.

I don't know of any problems in using a local LLVM instead of the one bundled with ROCm, so this change doesn't restrict it from using external libraries. It mostly just cleans up our build process since any version of LLVM, including the one bundled with ROCm, should work.

nakajee:
Thanks for your comment.
Waiting for comments from others (especially people who implemented or reviewed this part before).

benjaminulmer (approved):
Glad we can finally clean this up. LGTM

PR #1591, opened 2022-09-15
Report errors caused by subprocesses

AlexBrownAMD:
And optionally print all subprocess output in verbose mode

### DIFF
--- a/Tensile/TensileCreateLibrary.py
+++ b/Tensile/TensileCreateLibrary.py
@@ -276,14 +276,11 @@ def buildSourceCodeObjectFile(CxxCompiler, outputPath, kernelFile):
               if globalParameters["PrintCodeCommands"]:
                 print(' '.join(bundlerArgs))
               # change to use  check_output to force windows cmd block util command finish
-              try:
-                out = subprocess.check_output(bundlerArgs, stderr=subprocess.STDOUT)
-                print2(out)
-              except subprocess.CalledProcessError as err:
-                print(err.output)
-                raise
+              out = subprocess.check_output(bundlerArgs, stderr=subprocess.STDOUT)
+              print2(out)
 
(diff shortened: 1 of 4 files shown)


Tensile/TensileCreateLibrary.py
```diff
@@ -261,7 +276,12 @@ def buildSourceCodeObjectFile(CxxCompiler, outputPath, kernelFile):
               if globalParameters["PrintCodeCommands"]:
                 print(' '.join(bundlerArgs))
               # change to use  check_output to force windows cmd block util command finish
-              subprocess.check_output(bundlerArgs, stderr=subprocess.STDOUT)
+              try:
+                out = subprocess.check_output(bundlerArgs, stderr=subprocess.STDOUT)
```
nakajee, line 280: This part is already in try section.
Do we need to use nested try here?

AlexBrownAMD: Just posted an update to clean that up. Thanks!

nakajee: I found 1 more subprocess.check_output call at line 238 (before change) in TensileCreateLibrary.py.
Do we need to add print for the output of this call?

AlexBrownAMD: That one should also be covered by my most recent change

nakajee: I understand that the exception is captured.
I am just talking about print2 you added for all other check_output calls.
Like this.
  print2(out)

benjaminulmer (approved):
Still not perfect compared to the `check_call` we used to have, but it's an appreciated improvement.
All the appropriate calls have been updated as far as I can tell.

cgmillette (approved):
LGTM

PR #1594, opened 2022-09-16
Fixes for tests failing on CI

AlexBrownAMD:
Mark tests failing due to rocm 5.3 as xfail for now.

Fix tests failing due to CpuThreads=0. CpuThreads=0 used to mean disable threading, but the value of CpuThreads was passed directly to the make command. Current versions of make do not support setting "-j0" (number of threads must be a positive number). This changes the behaviour of CpuThreads to mean positive number is number of threads requested, non-positive number means use max threading. If a user really wants to disable makefile threading they should set threads to 1. Also removed instances where CpuThreads was manually set to 0 in CI test cases.

### DIFF
--- a/Tensile/Parallel.py
+++ b/Tensile/Parallel.py
@@ -29,15 +29,15 @@ import sys
 
 def CPUThreadCount(enable=True):
   from .Common import globalParameters
-  if not enable or globalParameters["CpuThreads"] == 0:
-    return 0
+  if not enable:
+    return 1
   else:
     if os.name == "nt":
       cpu_count = os.cpu_count()
     else:
       cpu_count = len(os.sched_getaffinity(0))
     cpuThreads = globalParameters["CpuThreads"]
-    if cpuThreads < 0:
+    if cpuThreads < 1:
         return cpu_count*abs(cpuThreads)
     return min(cpu_count, cpuThreads)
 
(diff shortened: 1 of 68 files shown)


Tensile/Parallel.py
```diff
@@ -29,15 +29,15 @@
...
+    return 1
   else:
     if os.name == "nt":
       cpu_count = os.cpu_count()
     else:
       cpu_count = len(os.sched_getaffinity(0))
     cpuThreads = globalParameters["CpuThreads"]
-    if cpuThreads < 0:
+    if cpuThreads < 1:
         return cpu_count*abs(cpuThreads)
```
nakajee, line 41: I think this returns 0 if cpuThreads=0.
If we would like to make 0 behave like -1, it is better to change to "return cpu_count".

AlexBrownAMD: Good catch, just posted a fix. Thanks!

nakajee (approved):
Thanks for your update.
Looks good.

benjaminulmer (commented):
`CpuThreads` changes are fine by me.
I spun up another run of extended tests; hopefully it clears.

benjaminulmer:
Is `4xi8gemm_hpa_hip_nn` the validation fail you mentioned this morning @AlexBrownAMD ? Or is this something else?

AlexBrownAMD:
> Is `4xi8gemm_hpa_hip_nn` the validation fail you mentioned this morning @AlexBrownAMD ? Or is this something else?

Looks like this one is also caused by 5.3, but this test failure doesn't happen consistently every build. I'm marking it as skip instead of xfail for now, and will file a ticket for it to be investigated further.

benjaminulmer (approved):
lgtm. Thanks for putting the work in on this one

PR #1511, opened 2022-05-19
Enabling gfx1031

littlewu2508:
This commit enables benchmarking and compiling kernels for gfx1031 target

PR #1595, opened 2022-09-17
Add gfx900:xnack-, gfx1032, gfx1034, gfx1035

cgmb:
This change enables Tensile to be built for a few more architectures. None of the architectures being added are relevant to the official binaries distributed by AMD, but they're useful for users building from source. I'd have added gfx1013, gfx1033, and gfx1036 as well, but I'm not sure what names put in the architecture map.

The addition of gfx1031 in #1511 really was a really helpful guide in making this change.

* The miscommunication relating to whether gfx900 should be gfx900 or gfx900:xnack- was never properly resolved and so many of the libraries that depend on rocBLAS use gfx900:xnack- in their default configuration.

  This change allows both gfx900 and gfx900:xnack- to be accepted by Tensile, thereby allowing users that are building from source to choose whatever they prefer.

* This change allows users to build rocBLAS from source for Navi23, Navi24 and Rembrandt.

### DIFF
--- a/Tensile/Common.py
+++ b/Tensile/Common.py
@@ -271,7 +271,7 @@ defaultGlobalParameters = deepcopy(globalParameters)
 
 # Translate GPU targets to filter filenames in Tensile_LOGIC directory
 architectureMap = {
-  'all':'_','gfx000':'none', 'gfx803':'r9nano', 'gfx900':'vega10',
+  'all':'_','gfx000':'none', 'gfx803':'r9nano', 'gfx900':'vega10', 'gfx900:xnack-':'vega10',
   'gfx906':'vega20', 'gfx906:xnack+':'vega20', 'gfx906:xnack-':'vega20',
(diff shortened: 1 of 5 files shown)


benjaminulmer (approved):
I don't see any potential problems with adding moar arches. I don't have access to the right machines to test, but the changes look fine to me.

cgmb:
On a gfx1030 machine, you can kind of test it by using `export HSA_OVERRIDE_GFX_VERSION=10.3.2`, `export HSA_OVERRIDE_GFX_VERSION=10.3.4` or `export HSA_OVERRIDE_GFX_VERSION=10.3.5` to test gfx1032, gfx1034 and gfx1035, respectively.

(It raises the question: if these architectures are so compatible that you can run code compiled for one on the other, perhaps Tensile should be doing something smarter than treating them as entirely distinct? That would probably be ideal, but it's not obvious how that should work.)

PR #1596, opened 2022-09-26
Tensile_4_33_Rocm_SMI_1_4_1_GPU dynamic frequency, power, temperature information logging within tensile

aferoz21:
Logging of GPU dynamic information frequency, power, temperature. This eventually will allow is to calculate the GEMM kernel efficiency.

Tensile reports have been modified to log the median frequency, power, temperature like below. ROCm API have been used to extract the current frequency, power temperature and array during kernel execution and median is calculated and reported in the below format.   

**From** 
run,problem-progress,solution-progress,operation,problem-sizes,solution,validation,time-us,gflops,empty,total-gran,tiles-per-cu,num-cus,tile0-gran,tile1-gran,cu-gran,wave-gran,mem-read-bytes,mem-write-bytes,temp-edge,clock-sys,clock-soc,clock-mem,fan-rpm,hardware-samples,enqueue-time
0,0/0,0/10,Contraction_l_Ailk_Bjlk_Cijk_Dijk,"(1024,1024,1,512)",Cijk_Ailk_Bjlk_DB_MT128x128x16_MI16x16x4x1_SN_K1_WGM4,NO_CHECK,305.731,3512.05,,0.581818,0.581818,110,1,1,0.581818,1,75497472,8388608,27,1350,1090,1600,nan,1,2022-09-26 15:56:17.  

**To**
run,problem-progress,solution-progress,operation,problem-sizes,solution,validation,time-us,gflops,empty,total-gran,tiles-per-cu,num-cus,tile0-gran,tile1-gran,cu-gran,wave-gran,mem-read-bytes,mem-write-bytes,temp-edge,clock-sys,clock-soc,clock-mem,fan-rpm,hardware-samples,**gfx-frequency(median),power(median),hotspot-temperature(median)**,enqueue-time
0,0/0,0/10,Contraction_l_Ailk_Bjlk_Cijk_Dijk,"(1024,1024,1,512)",Cijk_Ailk_Bjlk_DB_MT128x128x16_MI16x16x4x1_SN_K1_WGM4,NO_CHECK,85.8809,12502.7,,0.581818,0.581818,110,1,1,0.581818,1,75497472,8388608,26.8824,1350,1090,1600,nan,**17,1263,36,**243,2022-09-26 15:28:30.


Tensile/Source/client/source/HardwareMonitor.cpp
```diff
@@ -386,9 +390,178 @@ namespace Tensile
...
+            if(m_dataPoints == 0)
+                throw std::runtime_error("No data points collected!");
+                
+            if (m_freqValues.size() != 0)
+            {
+                if(m_freqValues[0] == std::numeric_limits<uint16_t>::max())
+                    return std::numeric_limits<double>::quiet_NaN();
+
+                 size_t FrequencyMidValueIndex = m_freqValues.size() / 2;
+                 std::nth_element(m_freqValues.begin(), m_freqValues.begin()+FrequencyMidValueIndex, m_freqValues.end());
```
nakajee, line 424: Is m_freqValues sorted?

Tensile/Source/client/source/HardwareMonitor.cpp
```diff
@@ -386,9 +390,178 @@ namespace Tensile
                 else
                     m_fanValues[i] += newValue;
             }
+            
+            rsmi_gpu_metrics_t gpuMetrics;
+            auto status = rsmi_dev_gpu_metrics_info_get(m_smiDeviceIndex, &gpuMetrics);
+            if(status != RSMI_STATUS_SUCCESS)
+            {
+                m_freqValues[0]        = std::numeric_limits<uint16_t>::max();
```
nakajee, line 398: The size of m_freqValues can be 0 here, right?
Is that safe to assign value to m_freqValues[0] here?

babakpst: Good catch, If rsmi_dev_gpu_metrics_info_get returns false in the first attempt, the m_freqValues is of size 0. We can check the size of arrays first.

Tensile/Source/client/source/HardwareMonitor.cpp
babakpst: The nth_element is in place and does not return anything.

Tensile/Source/client/source/HardwareMonitor.cpp
```diff
@@ -386,9 +390,178 @@ namespace Tensile
...
+                m_powerValues[0]       = std::numeric_limits<uint16_t>::max();
+                m_temperatureValues[0] = std::numeric_limits<uint16_t>::max();
+            }
+            else
+            {
+                m_freqValues.push_back(gpuMetrics.average_gfxclk_frequency);
```
babakpst, line 404: What is the return type of gpuMetrics.XXX?

Tensile/Source/client/source/HardwareMonitor.cpp
```diff
@@ -386,9 +390,178 @@ namespace Tensile
...
+
+                 size_t FrequencyMidValueIndex = m_freqValues.size() / 2;
+                 std::nth_element(m_freqValues.begin(), m_freqValues.begin()+FrequencyMidValueIndex, m_freqValues.end());
+                 return static_cast<double> (m_freqValues[FrequencyMidValueIndex]);
+            }
+            
+            throw std::runtime_error("Can't read Gfx value that wasn't requested: ");
+        }
+        
+        double HardwareMonitor::getAverageGfxFrequency()
```
babakpst, line 431: I cannot find a call to this function. Are we using this? Anyhow, it is ok to keep it.

Tensile/Source/client/source/HardwareMonitor.cpp
```diff
@@ -386,9 +390,178 @@ namespace Tensile
...
+                if(m_temperatureValues[0] == std::numeric_limits<uint16_t>::max())
+                    return std::numeric_limits<double>::quiet_NaN();
+                    
+                return static_cast<double> (std::accumulate(m_temperatureValues.begin(), m_temperatureValues.end(), 0.0) / m_temperatureValues.size()) ;
+            }
+            
+            throw std::runtime_error("Can't read temperature value that wasn't requested: ");
+        }
+        
+        void HardwareMonitor::logMinMaxMedianAverage()
```
babakpst, line 525: Are we using this func?

babakpst (commented):
Good work. This adds some critical parameters for tuning.

Tensile/Source/client/source/HardwareMonitor.cpp
```diff
@@ -386,9 +390,178 @@ namespace Tensile
...
+
+            if(m_dataPoints == 0)
+                throw std::runtime_error("No data points collected!");
+
+            if (m_powerValues.size() != 0)
+            {
+                if(m_powerValues[0] == std::numeric_limits<uint16_t>::max())
+                    return std::numeric_limits<double>::quiet_NaN();
+                    
+                return static_cast<double> (std::accumulate(m_powerValues.begin(), m_powerValues.end(), 0.0) / m_powerValues.size()) ;
```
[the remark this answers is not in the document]
yoichiyoshida: If the starting value 0.0 for the third parameter of std::accumulate is considered as double (which I think it is) then it should be accumulating to double. Can be more explict and static cast 0 to double to be clearer to the reader, but not necessary. Probably don't need the static_cast in the front though?

Tensile/Source/client/include/ResultReporter.hpp
```diff
@@ -124,6 +124,9 @@ namespace Tensile
             const std::string DeviceIndex         = "device-idx";
             const std::string FanSpeedRPMs        = "fan-rpm";
             const std::string HardwareSampleCount = "hardware-samples";
+            const std::string GfxFrequency = "gfx-frequency(median)"; // GPU clock in Mhz 
+            const std::string Power =  "power(median)"; // power consumption by the GPU
+            const std::string TemperatureHot = "hotspot-temperature(median)"; // GPU temperature
```
yoichiyoshida, line 129: Please follow formatting of the variables above for consistency

Tensile/Source/client/source/HardwareMonitor.cpp
yoichiyoshida: This ties in with my question earlier, where for the other metrics we pre-size the vector so we wouldn't run into the case where the array is empty. Need to either fix it to be consistent with the other metrics or handle the edge case here.

Tensile/Source/client/source/HardwareMonitor.cpp
yoichiyoshida: uint16_t, matches the array

Tensile/Source/client/include/HardwareMonitor.hpp
```diff
@@ -143,6 +152,9 @@ namespace Tensile
 
             std::vector<uint32_t> m_fanMetrics;
             std::vector<int64_t>  m_fanValues;
+            std::vector<uint16_t> m_freqValues;
+            std::vector<uint16_t> m_powerValues;
+            std::vector<uint16_t> m_temperatureValues;
```
yoichiyoshida, line 157: Can we call this something more consistent like m_tempHotspotValues? We already have a m_tempValues and this makes it confusing.

Tensile/Source/client/source/HardwareMonitor.cpp
```diff
@@ -326,7 +326,11 @@ namespace Tensile
                 v = 0;
             for(auto& v : m_fanValues)
                 v = 0;
-
+                
+            m_freqValues.clear();
+            m_powerValues.clear();
+            m_temperatureValues.clear();
```
yoichiyoshida, line 332: is there a reason why we're not pre-setting the size and clearing them like the other things we log, rather than emptying the vector and pushing data in later?

yoichiyoshida (changes requested):
The new metrics aren't being collected collected in the same manner as the existing metrics, is there a reason for that?

Tensile/Source/client/source/HardwareMonitor.cpp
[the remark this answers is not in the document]
aferoz21: 3rd argument to the accumulate function is 0.0 so it stores the result in 8 byte double, hence cannot cause overflow for uint16_t type.  ROCm API  gpuMetrics.XXX return type also uint16_t so API guarantees uint16_t value. 
let me change the casting to before division.

Tensile/Source/client/source/HardwareMonitor.cpp
aferoz21: std:nth_element does partial sorting. it does find a mid position such that all values before this is less than the midpoint value and all values after this will be greater than this mid point value. Efficiency through this way seem to be slightly better than full sorting to find median.  Full sorting is n log(n) Vs O(n) . so I think this could be better. but can revisit later if needed.

Tensile/Source/client/source/HardwareMonitor.cpp
aferoz21: Good catch. it might be an issue, changing to something like this.
...
m_freqValues.resize(1);
m_powerValues.resize(1);
m_temperatureValues.resize(1);
..
void HardwareMonitor::collectOnce() 
{
......
if(status != RSMI_STATUS_SUCCESS)
{
                m_freqValues.at(0)        = std::numeric_limits<uint16_t>::max();
                m_powerValues.at(0)       = std::numeric_limits<uint16_t>::max();
                m_temperatureValues.at(0) = std::numeric_limits<uint16_t>::max();
}
else
{
                // size of m_freqValues,m_powerValues,m_temperatureValues is same.   
                if(m_freqValues[0] != std::numeric_limits<uint16_t>::max())  
                {    
                    m_freqValues.push_back(gpuMetrics.average_gfxclk_frequency);
                    m_powerValues.push_back(gpuMetrics.average_socket_power);
                    m_temperatureValues.push_back(gpuMetrics.temperature_hotspot);
                }
}
...

Tensile/Source/client/source/HardwareMonitor.cpp
aferoz21: yes rsmi_dev_gpu_metrics_info_get returns false then before filling the m_freqValues I am checking if the index 0 has invalid value. I am using m_freqValues .resize(1) and push_back combination.

The nth_element function just rearranges the elements or does partial sorting to get median value.  

 gpuMetrics.XXX return value from the API is uint16_t

Tensile/Source/client/source/HardwareMonitor.cpp
aferoz21: These functions are place holder for future enhancement.  I am going to add tensile switch , if it is enabled I will invoke these functions  getAverageGfxFrequency(), logMinMaxMedianAverage(). 

logMinMaxMedianAverage output will be something like this for each GEMM kernel execution.
ROCm SMI API consolidated frequency,power,temperature data

GFX Value	PPT0_value	Temperature
1245         	253		        36
1245         	108		        36
...
				  Min    Max	      Average	Median
GFX Frequency	  898     1257             1213.17   1256    
Power Value		  108     262                237.733   262     
Temperature		  36      38                   36.7          37

Tensile/Source/client/source/HardwareMonitor.cpp
aferoz21: currently not using this function but will eventually use this in future with new switch added otherwise it will be test time hit.

Tensile/Source/client/include/ResultReporter.hpp
aferoz21: Existing metrics just collect the average information alone.  vector size is 1 for all the values . just accumulates all the values in the same index.
New metrics we are planning to log the detailed report like min, max, average, median and all individual data points during kernel execution hence slightly different. slightly deviating from existing way for minimal overhead. 

yes formatting needs changes.

Tensile/Source/client/include/HardwareMonitor.hpp
aferoz21: >  Need to either fix it to be consistent with the other metrics or handle the edge case
Yes working on handling the edge case.

>Can we call this something more consistent like m_tempHotspotValues? 
yes, you are correct.

Tensile/Source/client/source/HardwareMonitor.cpp
aferoz21: we need all individual data later for detailed logging so keeping all the values. Earlier metrics it accumulates (adds) all the individual data. separating out and logging the individual data is overhead.

Tensile/Source/client/source/HardwareMonitor.cpp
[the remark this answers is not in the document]
nakajee: OK. If the return type of std::accumulate is double, we do not need static_cast<double> here.

Tensile/Source/client/source/HardwareMonitor.cpp
```diff
@@ -386,9 +394,184 @@ namespace Tensile
...
+                    
+                size_t temperatureMidValueIndex = m_tempHotspotValues.size() / 2;
+                std::nth_element(m_tempHotspotValues.begin(), m_tempHotspotValues.begin()+temperatureMidValueIndex, m_tempHotspotValues.end());
+                return static_cast<double> (m_tempHotspotValues[temperatureMidValueIndex]);
+            }
+            
+            throw std::runtime_error("Can't read temperature value that wasn't requested: ");
+        }
+        
+        double HardwareMonitor::getAverageHotSpotTemperature()
```
nakajee, line 516: Can we unify 3 getAverage functions?
I think we just need to pass the reference of vector (and error message if we would like to change).

nakajee: Same for getMedian

Tensile/Source/client/source/HardwareMonitor.cpp
```diff
@@ -386,9 +394,184 @@ namespace Tensile
...
+            std::cout << std::setw(8)<< *std::max_element(m_powerValues.begin(), m_powerValues.end());
+            std::cout <<std::setw(10)<< std::accumulate(m_powerValues.begin(), m_powerValues.end(), double(0)) / m_powerValues.size();
+            std::cout << std::setw(8)<< getMedianPower()<<"\n";
+            
+           // min, max, avg, median Temperature 
+            std::cout <<std::left<<std::setw(15)<<"Temperature\t\t"<< std::setw(8);
+            std::cout << *std::min_element(m_tempHotspotValues.begin(), m_tempHotspotValues.end());
+            std::cout << std::setw(8)<< *std::max_element(m_tempHotspotValues.begin(), m_tempHotspotValues.end());
+            std::cout << std::setw(10)<< std::accumulate(m_tempHotspotValues.begin(), m_tempHotspotValues.end(), double(0)) / m_tempHotspotValues.size();
+            std::cout << std::setw(10)<< getMedianHotSpotTemperature()<<"\n\n";
```
nakajee, line 573: Is it intentional to use setw(10) here? Other locations use setw(8) for median value.

Tensile/Source/client/source/HardwareMonitor.cpp
```diff
@@ -386,9 +394,184 @@ namespace Tensile
...
+                m_freqValues.at(0)        = std::numeric_limits<uint16_t>::max();
+                m_powerValues.at(0)       = std::numeric_limits<uint16_t>::max();
+                m_tempHotspotValues.at(0) = std::numeric_limits<uint16_t>::max();
+            }
+            else
+            {
+                // size of m_freqValues,m_powerValues,m_tempHotspotValues is same.   
+                if(m_freqValues[0] != std::numeric_limits<uint16_t>::max())  
+                {    
+                    m_freqValues.push_back(gpuMetrics.average_gfxclk_frequency);
```
nakajee, line 411: This is not a good solution.
We should remove the first element if we get the valid result.

I think it is better to have another variable to specify if the result is valid or not.
(instead of using [0] to identify if it is valid or not)

nakajee: Or, just consider size()==0 is invalid.

nakajee: I am not sure if this is called multiple times.
If we call this multiple times and make the result invalid if one of the calls is invalid (like valid, valid, invalid -> invalid), it is better to have another variable.
If not, I think we just need to push the result only if the result is valid (do nothing here if the result is invalid).

nakajee:
By the way, please fix the errors in StaticAnalysis.

Tensile/Source/client/include/HardwareMonitor.hpp
```diff
@@ -143,6 +152,9 @@ namespace Tensile
 
             std::vector<uint32_t> m_fanMetrics;
             std::vector<int64_t>  m_fanValues;
+            std::vector<uint16_t> m_freqValues;
```
nakajee, line 155: I read the existing code.
It looks like you implemented these vectors in a different way compared to other existing metrices.
Other existing vectors represent different device/HW, but you use these vectors to store multiple inputs.
It is probably because of median, right?
Why we use median only for new metrices?

The current way is a bit confusing.

aferoz21: Yes. Initially I thought to keep the same old way. but I wanted to print all these information at the end of kernel execution. do you think these information will not be useful later ?
GFX Value				PPT0_value	       Temperature
1262         				190					36
1258         				140					35
1258         				140					35
1258         				140					35
1258         				138					36
1259         				190					36
1259         				190					36
1259         				236					36
1262         				236					36
1262         				236					36
1265         				256					36
1265         				256					36
1265         				256					36
1272         				261					36
1272         				261					36
1301         				261					36
1272         				263					36
1288         				263					36
1288         				263					36
1288         				265					36

				Min	   Max	Average	Median
GFX Frequency	1258    1301    1268.55    1265    
Power Value		  138     265     222.05      256     
Temperature		  35      36      35.85          36

nakajee: We can keep track of max/min value by adding variables for them.
The only thing is median.
If we really need it, I am OK, but I suggest to use different naming for these vector variables and/or put some comment to recognize the new different way.

nakajee: Or, use vector of vector to make the structure align with others.

Tensile/Source/client/source/HardwareMonitor.cpp
aferoz21: Functions unified

Tensile/Source/client/source/HardwareMonitor.cpp
aferoz21: updated setw common for avg and median

Tensile/Source/client/source/HardwareMonitor.cpp
aferoz21: added another variable to keep track invalid GPU index

Tensile/Source/client/include/HardwareMonitor.hpp
aferoz21: Added comments to recognize new different way. I tried to change to current implementation to match existing  but realized it is not needed hence not changing the implementation. Median value is needed and also better to preserve the individual data values. 
Current implementation uses ROCm API (rsmi_dev_gpu_metrics_info_get) where we get all HW type details through single invocation. Existing metrics implementation multiple iteration to store different HW type. if we need to store different HW type in the future, new additonal vector type or vector<vector>>would be appropriate in future.

Tensile/Source/client/source/HardwareMonitor.cpp
```diff
@@ -387,9 +394,158 @@ namespace Tensile
...
+                return std::numeric_limits<double>::quiet_NaN();
+
+            switch(freqPowerTempIndex)
+            {
+            case GPU_FREQUENCY_AVG:
+                if(m_freqValues.size() != 0)
+                {
+                    return (std::accumulate(m_freqValues.begin(), m_freqValues.end(), double(0)))
+                           / m_freqValues.size();
+                }
```
nakajee, line 431: We need break here.

Tensile/Source/client/source/HardwareMonitor.cpp
```diff
@@ -387,9 +394,158 @@ namespace Tensile
...
+                {
+                    m_freqValues.push_back(gpuMetrics.average_gfxclk_frequency);
+                    m_powerValues.push_back(gpuMetrics.average_socket_power);
+                    m_tempHotspotValues.push_back(gpuMetrics.temperature_hotspot);
+                }
+            }
             m_dataPoints++;
         }
 
+        double HardwareMonitor::getAverageMedianGfxFreqPowerTemperature(int freqPowerTempIndex)
```
nakajee, line 415: Can we unify the ave/median calculation for all 3 types?
We just need to pass the reference of vector.

Tensile/Source/client/source/HardwareMonitor.cpp
```diff
@@ -387,9 +394,158 @@ namespace Tensile
...
+            }
+
+            std::cout << "\n";
+            std::cout << "\t\t\t\t\tMin\t\t"
+                      << "Max\t\t"
+                      << "Average\t\t"
+                      << "Median\n";
+
+            // min, max,avg, median frequency
+            std::cout << std::left << std::setw(20) << "GFX Frequency" << std::setw(8)
```
nakajee, line 519: Also, unify this part as well.
We can create a sub function and pass a reference of vector.

Tensile/Source/client/source/HardwareMonitor.cpp
```diff
@@ -327,6 +327,13 @@ namespace Tensile
             for(auto& v : m_fanValues)
                 v = 0;
 
+            m_hasInvalidGpuMetricStatus = false;
+            m_freqValues.resize(1);
```
nakajee, line 331: Do we still need resize(1) here?

Tensile/Source/client/include/HardwareMonitor.hpp
```diff
@@ -143,6 +146,14 @@ namespace Tensile
 
             std::vector<uint32_t> m_fanMetrics;
             std::vector<int64_t>  m_fanValues;
+            // Below 3 vectors stores the individual values of GFx frequency, GPU power, GPU temperature during GEMM kernel execution.
+            // These vectors are implemented slightly different from previous clock/fan metric type implementation which uses add**Monitor functions to              // monitor and store multiple HW types information in the vectors for the same device. (ie)Existing metric implementation represent different            // HW type and uses ROCm API to get multiple HW type(like different sensor, different clock). Hence it uses add**Monitor functions to store              // in multiple vectors. each new HW type requires an invocation of ROCm API.
```
nakajee, line 150: I do not know what is happening here...
Please add new lines properly.

Tensile/Source/client/include/HardwareMonitor.hpp
```diff
@@ -143,6 +146,14 @@ namespace Tensile
...
+            // but below vectors uses ROCm API (rsmi_dev_gpu_metrics_info_get) through single invocation gets all the HW type details hence it dont need
+            // exisitng type of implementation. if we need to store different HW type in the future, new additonal vector type or vector<vector>>would be            // appropriate.
```
nakajee, line 152: Please use spell checker for your code change.
  exisitng -> existing
  dont -> don't or do not
  additonal -> additional

Tensile/Source/client/source/HardwareMonitor.cpp
```diff
@@ -387,9 +391,129 @@ namespace Tensile
...
+            case GPU_TEMPERATURE_MEDIAN:
+                return getAverageOrMedianValue(0, m_tempHotspotValues);
+            default:
+                std::cerr << "Invalid statistical type index.\n";
+                return std::numeric_limits<double>::quiet_NaN();
+            }
+        }
+
+        double
+            HardwareMonitor::getAverageOrMedianValue(bool                   isAvg,
```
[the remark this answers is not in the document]
nakajee: Probably, we can unify getAverageOrMedianValue and getAverageMedianGfxFreqPowerTemperature.
Like 
(1) ave function
  getAverageGfxFreqPowerTemperature(std::vector<uint16_t>& dataValues)
(2) median function
  getMedianGfxFreqPowerTemperature(std::vector<uint16_t>& dataValues)

We can directly calculate the median or avg instead of putting a switch case.

Tensile/Source/client/source/HardwareMonitorListener.cpp
```diff
@@ -99,6 +99,21 @@ namespace Tensile
 
             m_reporter->report(ResultKey::FanSpeedRPMs, m_monitor->getAverageFanSpeed());
             m_reporter->report(ResultKey::HardwareSampleCount, m_monitor->getSamples());
+
+            // Report avg frequency,power,temperature during kernel execution.
+            m_reporter->report(
+                ResultKey::GfxFrequency,
+                m_monitor->getAverageMedianGfxFreqPowerTemperature(GPU_FREQUENCY_MEDIAN));
```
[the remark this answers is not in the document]
nakajee: Any comment about why you use median here rather than average like other existing parameters?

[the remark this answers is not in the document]
nakajee: OK. Can you put some explanation in the code?

nakajee:
Thanks for your update.
I do not have any further comments.
Please make sure all comments from @yoichiyoshida and @babakpst are addressed.

babakpst (approved):
I don't have any further comments. Please merge, once CI tests passed.

benjaminulmer:
Please squash-merge and give the commit a useful title when you do merge

aferoz21:
This PR will be waiting to merge due to a known issue reported below to the ROCm SMI team. This PR causes the CI machine to go to faulty state due to race condition.
https://ontrack-internal.amd.com/browse/SWDEV-361923

aferoz21:
Re-running the PR as the rocm version has been changed to 5.4

babakpst (approved):

yoichiyoshida (approved):

PR #1597, opened 2022-09-27
Remove unused file in cmake interface

JeremyAdamHart:
As far as I can tell, TensileCreateLibrary.cmake isn't actually used anywhere, and TensileCreateLibraryFiles in TensileConfig.cmake is the intended method for creating the library. I've removed this file and the relevant include. The Tensile tuning client, HostLibraryTests and rocBLAS are all verified to work correctly with this file removed without any additional changes.

I've also added some documentation for the steps required to build Tensile with cmake to make the process slightly more transparent. It's a little hidden away in TensileConfig.cmake, if there's a more obvious place to put it let me know and I'll move it there instead.


AlexBrownAMD (approved):
Looks good to me! If you're looking for a slightly more visible place for your documentation notes, maybe it could be added to our wiki?

PR #1598, opened 2022-09-27
Assembly syntax and coverage for DirectToLds has changed

AlexBrownAMD:
When using lds modifier, destination vgpr must now be omitted. The lds modifier only applies to global loads that would have a single register as their destination, so DirectToLds no longer works with dgemm/cgemm/xgemm. Also doesn't work with hgemm since that case currently requires GRVW=4.

Updated solution reject cases and added additional warnings if DTL is somehow disabled when requested, or requested on hardware that doesn't support it.

Updated test cases: re-enabled working test suites, and moved broken tests to disabled folder.


AlexBrownAMD:
I ran some Tensile tests locally before posting, but I still need to run some tests with rocblas

Tensile/KernelWriterAssembly.py
```diff
@@ -2223,6 +2225,16 @@ def defineBufferMemoryMacros(self):
       replace = f'{type_list[t]}' if (self.version[0] < 11) else f'{t}'
       kStr += self.generalMacro('buffer_load_', origin, replace, 'dst', 'voffset', 'base', 'soffset', 'offen', 'ioffset', 'md0', 'md1', 'md2') + self.endLine
 
+    # Extra macro for DirectToLds loads with no destination register
+    type_list = {
+      'b32'       : 'dword',
+      'u16'       : 'ushort'
```
nakajee, line 2231: Do we support u16?

Tensile/KernelWriterAssembly.py
```diff
@@ -10998,7 +11025,7 @@ def globalWriteElements(self, kernel, vectorWidths, elements,
   # bpl = bytes per load op
   ##############################################################################
   def chooseGlobalRead(self, useBuffer, bpl, destVgpr, \
-                       addr0, addr1, soffset, offset, extraFields, hi16=0, comment="load C"):
+                       addr0, addr1, soffset, offset, extraFields, directToLds, dtlUseDestVgpr, hi16=0, comment="load C"):
```
nakajee, line 11028: I think we can unify these 2 new parameters.
We just need 1 parameter like dtlNoDestVgpr to recogize new dtl format.

Tensile/SolutionStructs.py
```diff
@@ -2443,6 +2443,11 @@ def isDirectToLdsDoable(state, tc):
     #TN
     # use for all precisions with TransposeLDS=1
 
+    numRegisters = state["ProblemType"]["DataType"].numRegisters()
+    if numRegisters > 1 or numRegisters * state["GlobalReadVectorWidth"] != 1:
```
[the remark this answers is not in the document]
AlexBrownAMD: Yea, I moved the dgemm and hgemm tests to the disabled folder for now. Not sure if there's any other plans to support them in the future, if not we can just delete them.

Tensile/Common.py
```diff
@@ -1704,47 +1704,52 @@ def locateExe( defaultPath, exeName ): # /opt/rocm/bin, hip-clang
...
-                            or tryAssembler(isaVersion, "buffer_load_b32 v40, v36, s[24:27], s28 offen offset:0 lds")
-  rv["HasAddLshl"]        = tryAssembler(isaVersion, "v_add_lshl_u32 v47, v36, v34, 0x2")
-  rv["HasLshlOr"]         = tryAssembler(isaVersion, "v_lshl_or_b32 v47, v36, 0x2, v34")
-  rv["HasSMulHi"]         = tryAssembler(isaVersion, "s_mul_hi_u32 s47, s36, s34")
-  rv["HasCodeObjectV3"]   = tryAssembler(isaVersion, "", False, "-mcode-object-version=2")
+  # Syntax of DirectToLds loads has changed: destination vgpr should be ommitted
+  # Old syntax should be removed in a future update as it is no longer supported
+  rv["HasDirectToLdsDest"]    = tryAssembler(isaVersion, "buffer_load_dword v40, v36, s[24:27], s28 offen offset:0 lds") \
+                                or tryAssembler(isaVersion, "buffer_load_b32 v40, v36, s[24:27], s28 offen offset:0 lds")
+  rv["HasDirectToLdsNoDest"]  = tryAssembler(isaVersion, "buffer_load_dword v36, s[24:27], s28 offen offset:0 lds") \
```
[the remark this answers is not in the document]
AlexBrownAMD: I think so, I think some architectures have just one name or the other. I'll have to double check references though.

Tensile/KernelWriterAssembly.py
AlexBrownAMD: Half+DTL doesn't work right now, but added for completeness since it was the only other data type in the list of macros that we generate that can also handle lds modifier

Tensile/SolutionStructs.py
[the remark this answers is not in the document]
AlexBrownAMD: Updated the numRegisters check in latest commit

nakajee:
>  Also doesn't work with hgemm since that case currently requires GRVW=4.

I quickly debugged this issue and figured out the cause.
GRVW=2 can be supported for HGEMM.

I will add back hgemm cases when I start PR for the fix.

Tensile/KernelWriterAssembly.py
AlexBrownAMD: Yea, my first pass at this change was going to add the lds field inside the choose function, but that ended up being messier. I've posted an update combining these parameters.

AlexBrownAMD:
> > Also doesn't work with hgemm since that case currently requires GRVW=4.
> 
> I quickly debugged this issue and figured out the cause. GRVW=2 can be supported for HGEMM.
> 
> I will add back hgemm cases when I start PR for the fix.

Thanks Koji

AlexBrownAMD:
I ran rocblas test using these changes. The build works and the test cases that were previously flagged now pass.

Tensile/SolutionStructs.py
```diff
@@ -2443,6 +2443,11 @@ def isDirectToLdsDoable(state, tc):
     #TN
     # use for all precisions with TransposeLDS=1
 
+    numRegisters = state["ProblemType"]["DataType"].numRegisters()
+    if numRegisters * state["GlobalReadVectorWidth"] != 1:
```
nakajee, line 2447: Probably, it is better to use state["GlobalLoadVectorWidth%c"%tc].
GlobalReadVectorWidth might be different from GlobalLoadVectorWidthA/B.

AlexBrownAMD: Updated in latest commit

Tensile/Common.py
```diff
@@ -1704,47 +1704,52 @@ def locateExe( defaultPath, exeName ): # /opt/rocm/bin, hip-clang
...
+  rv["HasExplicitCO"]         = tryAssembler(isaVersion, "v_add_co_u32 v0,vcc,v0,1")
+  rv["HasExplicitNC"]         = tryAssembler(isaVersion, "v_add_nc_u32 v0,v0,1")
 
-  rv["HasDirectToLds"]    = tryAssembler(isaVersion, "buffer_load_dword v40, v36, s[24:27], s28 offen offset:0 lds") \
-                            or tryAssembler(isaVersion, "buffer_load_b32 v40, v36, s[24:27], s28 offen offset:0 lds")
-  rv["HasAddLshl"]        = tryAssembler(isaVersion, "v_add_lshl_u32 v47, v36, v34, 0x2")
-  rv["HasLshlOr"]         = tryAssembler(isaVersion, "v_lshl_or_b32 v47, v36, 0x2, v34")
-  rv["HasSMulHi"]         = tryAssembler(isaVersion, "s_mul_hi_u32 s47, s36, s34")
-  rv["HasCodeObjectV3"]   = tryAssembler(isaVersion, "", False, "-mcode-object-version=2")
+  # Syntax of DirectToLds loads has changed: destination vgpr should be ommitted
```
nakajee, line 1711: Please do spell check.
 ommitted -> omitted

AlexBrownAMD: Ran spell check on the whole file

nakajee: Thanks for your update.
Please put some comment about spell check fix in your squash merge comment.

nakajee (approved):

benjaminulmer (approved):
I'm not qualified to speak to the code generator changes, but everything else seems to check out. 

If no one else has feedback, let's squash-merge and then work on cherry-picking this into the 5.4 branch.

PR #1601, opened 2022-10-05
binary search for GridBased algo

aazz44ss:
@benjaminulmer @yenong-amd 
I have modified binary search algorithm for grid based.
binary search smallest greater M first, then binary search smallest greater N, then search nearest K.
feel free to add comments or commits.

### DIFF
--- a/Tensile/Source/lib/include/Tensile/PropertyMatching.hpp
+++ b/Tensile/Source/lib/include/Tensile/PropertyMatching.hpp
@@ -79,11 +79,13 @@ namespace Tensile
             virtual ~MatchingTable() = default;
 
             virtual std::tuple<ReturnValue, double> findBestMatch(Object const& object,
-                                                                  Transform transform) const = 0;
+                                                                  Transform     transform) const
+                = 0;
 
             virtual ReturnValue findBestEvaluationSolution(Object const&   object,
                                                            Hardware const& hardware,
-                                                           Transform       transform) const = 0;
+                                                           Transform       transform) const
+                = 0;
 
             virtual std::vector<Value> matchesInOrder(Object const& object) const = 0;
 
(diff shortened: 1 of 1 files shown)


benjaminulmer:
@aazz44ss The search code should be identical, just moved to a different spot. My initial testing suggests this works, but I'll be doing some more in-depth verification as well.

Tensile/Source/lib/include/Tensile/PropertyMatching.hpp
```diff
@@ -624,5 +624,276 @@ namespace Tensile
...
+            std::tuple<ReturnValue, double> findBestKeyMatch(Key const& key,
+                                                             Transform  transform) const
+            {
+                if(Debug::Instance().printPropertyEvaluation())
+                    return findBestKeyMatch_GridBased<true>(key, transform);
+                else
+                    return findBestKeyMatch_GridBased<false>(key, transform);
+            }
+
+            template <bool T_Debug>
```
nielenventer, line 685: Howcome you have the debug param passed in as a template? It might make the function sig a bit cleaner to just do 

`bool T_Debug = Debug::Instance().printPropertyEvaluation()`

at the start of the function rather

benjaminulmer: This design is used for the other search algorithms, so it's mostly just used here for consistency. I'm guessing it was done this way to make the non-debug path have less branching/checking the debug flag, but I have no idea if it actually has a measurable impact on runtime.

benjaminulmer (approved):
Should get another approval since I helped work on this.

yenong-amd (approved):

benjaminulmer:
@aazz44ss Feel free to squash-merge whenever ready

PR #1603, opened 2022-10-12
modifying the GSU _GA for int8

babakpst:
This is a fix for some of the rocBLAS CI test failures. It covers I8II and 4xi8II.

### DIFF
--- a/Tensile/Common.py
+++ b/Tensile/Common.py
@@ -495,7 +495,7 @@ validParameters = {
     #
     # Table below shows the required kernels and workspace for each gemm function.
     # ------------------------------------------------------------------------------------------
-    # | gemm function          | non-HPA gemm                  | HPA gemm                      |
+    # | gemm function          | non-HPA gemm                  | HPA gemm (excluding int8)     |
     # | GSU Algorithm          | SingleBuffer | MultipleBuffer | SingleBuffer | MultipleBuffer |
     # | pre-kernel             | yes          | no             | yes          | no             |
     # | main kernel            | yes          | yes            | yes          | yes            |
--- a/Tensile/SolutionStructs.py
+++ b/Tensile/SolutionStructs.py
@@ -2602,7 +2602,7 @@ class Solution(collections.abc.Mapping):
           # For SingleBuffer algorithm, _GA and _WorkspaceSizePerElemC is updated only if the gemm function is HPA. The worskspace is
           # used to convert the final output from ComputeDataType to DestDataType. For non-HPA gemm functions the _GA and _Workspace
           # remain unchanged.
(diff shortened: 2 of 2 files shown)


benjaminulmer:
Thanks for the quick turnaround on this Babak. Did you run the failing rocBLAS tests with this version of Tensile and confirm it passes now?

nakajee:
Could you provide some more detail?
- how the failed case worked before your previous change and how your previous change caused the issue?
- how the fix works?

babakpst:
@benjaminulmer rocblas tests passed with a build from a similar branch. I will re-build rocblas with this commit and re run the failed cases again.

babakpst:
@nakajee The whole point of this (and the previous) PR was to allow singlebuffer GSU with HSS/BSS, which is working fine, but is getting rejected. Currently, the only option for HSS/BSS is MultipleBuffer which requires some memory allocation. HSS/BSS are getting rejected here: (if computeName != state["ProblemType"]["DestDataType"].toName():). To have singleBuffer GSU for HSS/BSS  we need to change the condition to (computeName != state["ProblemType"]["*DataType*"].toName()). Note that for singleBuffer GSU with  /HHS/BBS/BSS/HSS, _GA should be singleBuffer, unless all other types where _GA=None. 

Modifying this condition, changes _GA for 4xi8II (and I8II) from None to singlebuffer. Since we no longer use 4xi8II, I did not test this code for this case. That was the reason for the rocBlas failures.

nakajee:
> @nakajee The whole point of this (and the previous) PR was to allow singlebuffer GSU with HSS/BSS, which is working fine, but is getting rejected. Currently, the only option for HSS/BSS is MultipleBuffer which requires some memory allocation. HSS/BSS are getting rejected here: (if computeName != state["ProblemType"]["DestDataType"].toName():). To have singleBuffer GSU for HSS/BSS we need to change the condition to (computeName != state["ProblemType"]["_DataType_"].toName()). Note that for singleBuffer GSU with /HHS/BBS/BSS/HSS, _GA should be singleBuffer, unless all other types where _GA=None.
> 
> Modifying this condition, changes _GA for 4xi8II (and I8II) from None to singlebuffer. Since we no longer use 4xi8II, I did not test this code for this case. That was the reason for the rocBlas failures. By adding not state["ProblemType"]["DestDataType"].isInt32(), I excluded I8II/4xi8II from _GA changing.

Thanks for your comment.
So, the point is 
- previous change excluded only I8
- the new change excludes both I8 and 4xi8

This part is a bit confusing...
It might be better to put some additional comment in python code....

babakpst:
Correct, Initially, I was only excluding I8, here I excluded 4xi8, as well. I will add a comment once the tests pass.

benjaminulmer (approved):
Don't know what's up with CI, but it should be unrelated.

LGTM

PR #1606, opened 2022-10-24
Loop do-while in source kernel requires assert on size

AlexBrownAMD:
The F16 kernel that used this configuration required M=1. Adding a reject case to avoid this type of kernel being tuned without adding an assert on size, or updating the requirements after checking validation.

### DIFF
--- a/Tensile/SolutionStructs.py
+++ b/Tensile/SolutionStructs.py
@@ -2639,6 +2639,12 @@ class Solution(collections.abc.Mapping):
     if state["WavefrontSize"] == 32 and state["KernelLanguage"] == "Source":
       reject(state, "WavefrontSize=32 not yet supported for source kernels.")
 
+    if state["KernelLanguage"] == "Source" \
+        and state["ProblemType"]["DataType"].isHalf() \
+        and state["LoopDoWhile"] \
+        and not state["AssertSizeEqual"]:
+      reject(state, "hgemm source kernel with do-while loop requires an assert on size and validation")
+
     if state["EnableMatrixInstruction"]:
       if not (state["ProblemType"]["DataType"].isSingle() \
               or state["ProblemType"]["DataType"].isDouble() \


amgddm (approved):

Tensile/SolutionStructs.py
```diff
@@ -2639,6 +2639,12 @@ def assignDerivedParameters(state):
     if state["WavefrontSize"] == 32 and state["KernelLanguage"] == "Source":
       reject(state, "WavefrontSize=32 not yet supported for source kernels.")
 
+    if state["KernelLanguage"] == "Source" \
+        and state["ProblemType"]["DataType"].isHalf() \
+        and state["LoopDoWhile"] \
+        and not state["AssertSizeEqual"]:
```
nakajee, line 2645: Should we specify all dimensions?
I feel it is not enough to just check if AssertSizeEqual is defined or not.
We can specify keys of 0,1,2,3 corresponding to each dimension.

Tensile/SolutionStructs.py
```diff
@@ -2639,6 +2639,12 @@ def assignDerivedParameters(state):
     if state["WavefrontSize"] == 32 and state["KernelLanguage"] == "Source":
       reject(state, "WavefrontSize=32 not yet supported for source kernels.")
 
+    if state["KernelLanguage"] == "Source" \
```
nakajee, line 2642: Is this a temporal check?
I think it is better to add some comment to remove this temporal check later once the issue is fixed.

Tensile/SolutionStructs.py
AlexBrownAMD: This particular bug was on dimension 0 - I could add a check specifically for that one if you think we need it, my main goal was to make sure the tuner was checking limitations and validating if using these parameters. I don't fully know the limitations of the LoopDoWhile option, so I don't yet know what to add beyond that. I've made another ticket to investigate LoopDoWhile errors, and potentially generalize if we want to support the feature.

nakajee: > This particular bug was on dimension 0 - I could add a check specifically for that one if you think we need it, my main goal was to make sure the tuner was checking limitations and validating if using these parameters. I don't fully know the limitations of the LoopDoWhile option, so I don't yet know what to add beyond that. I've made another ticket to investigate LoopDoWhile errors, and potentially generalize if we want to support the feature.

I prefer to specify dimension 0 if we already know the bug is related only to dim 0.

nakajee:
I quickly checked the source code with LoopDoWhile.
I found some logic issues with LoopDoWhile.
(1) numIterL-- needs to be done at the beginning of loop instead of decrementing at the end
    numIterL is referred in the loop and the value in the loop will be different between while case and do-while case if we decrement it at the end of the loop
(2) better not to apply DoWhile to TailLoop
  0 iteration can happen at TailLoop.

I quickly made these fixes as follows. Please try it.
================ 
diff --git a/Tensile/KernelWriterSource.py b/Tensile/KernelWriterSource.py
index f9f628b9..114bc4af 100644
--- a/Tensile/KernelWriterSource.py
+++ b/Tensile/KernelWriterSource.py
@@ -2079,8 +2079,9 @@ class KernelWriterSource(KernelWriter):
     kStr = ""
     loopChar = self.indexChars[ \
         kernel["ProblemType"]["IndicesSummation"][loopIdx]]
-    if kernel["LoopDoWhile"]:
+    if kernel["LoopDoWhile"] and not tailLoop:
       kStr += "%sdo {%s" % (self.indent, self.endLine)
+      kStr += "%snumIter%s--;%s" % (self.indent, loopChar, self.endLine)
       assert(not self.unrollIncIsDepthU)
     else:
       if self.unrollIncIsDepthU and loopIdx==self.unrollIdx and not tailLoop:
@@ -2108,7 +2109,8 @@ class KernelWriterSource(KernelWriter):
   # Close Loop
-  def closeLoop(self, kernel, loopIdx, finalLoop, loopCopies, uDu=0, emitEndLabelOnly=False, oddLabel=False):
+  def closeLoop(self, kernel, loopIdx, finalLoop, uDu=0, emitEndLabelOnly=False, oddLabel=False):
+    tailLoop = loopIdx < 0
     kStr = ""
     if emitEndLabelOnly:
       return kStr
@@ -2129,8 +2131,8 @@ class KernelWriterSource(KernelWriter):
           incAmount = "1"

     self.indent = self.indent[2:]
-    if kernel["LoopDoWhile"]:
-      kStr += "%s} while (--numIter%s > %u);%s" \
+    if kernel["LoopDoWhile"] and not tailLoop:
+      kStr += "%s} while (numIter%s > %u);%s" \
           % (self.indent, loopChar, \
           (1 if kernel["PrefetchGlobalRead"] else 0), self.endLine )
     else:

Tensile/SolutionStructs.py
AlexBrownAMD: Updated

AlexBrownAMD:
Tested out Koji's fix and it is working in a variety of Tensile tests as well as the original rocblas-test prior to updates

PR #1609, opened 2022-10-27
GFX11: use WMMA instruction to optimize GEMM

TonyYHsieh:
GFX11: support WMMA instruction for h/hh/hb/i8
GFX11: HostLibraryTests yamls
GFX11: rearragne FP32(C)/FP64(C) instruction order for gfx11 optimization
GFX11: skip more test cases for gfx11 series
GFX11: use archCaps for instruction renaming condition
GFX11: adjust vgpr bank for A/B/C for optmize
GFX11: seperate vscnt and vmcnt for gfx11
fix flat store correctness issue
fix flat load correctness issue on I8


Tensile/KernelWriterAssembly.py
```diff
@@ -6235,7 +6246,10 @@ def endSummation(self, kernel, label = None, isOptNLL = False):
       #kStr += "s_nop %u\n" % instCycles
       kStr += self.MapAcctoArchRegs(kernel,option=0, isOptNLL=isOptNLL)
       if kernel["MIArchVgpr"] or kernel["StoreCInUnroll"]:
-        kStr += self.MulMIoutAlphaToArch(kernel)
+        if kernel["_GlobalAccumulation"] == 'MultipleBuffer':
```
nakajee, line 6249: I added extra vgpr allocation for MulMIoutAlphaToArch with complex case.
(from line 1519 to line 1523 before change).
If MoveMIoutToArch is called, the extra vgpr allocation is unnecessary.
Would you please add " and  kernel["_GlobalAccumulation"] != 'MultipleBuffer'" at line 1519 (before change) to avoid unnecessary extra vgpr allocation?

hcman2 (approved):
good for me.

TonyYHsieh:
@nakajee I modify change according your comment. May you help to review again?

nakajee (approved):
Sorry for late response.
I did not explicitly approve this PR since I was not a reviewer at first (I marked my comments as resolved...).
I do not have any further comment about the code change.

One more thing I would like to say is that it might be better if we have some test cases for WMMA.
I am not 100% sure but I could not find any test cases for WMMA as long as I checked.
You can add test cases later.

PR #1557, opened 2022-08-02
add skip test for gfx11xx series

TonyYHsieh:
some test cases should not run on gfx11xx.
add skip flag to those testcase.

PR #1610, opened 2022-10-28
hotfix: Assembly syntax and coverage for DirectToLds has changed (#1598)

AlexBrownAMD:
* Assembly syntax and coverage for DirectToLds has changed
* Update DTL check and error message
* Spell-check


nakajee:
It looks like dtl_tsgr_sgemm.yaml includes new changes in #1593.
It will probably fail without code changes in #1593.
Please get the exact version from 5.3 release and add only your change (remove xfails).

nakajee:
Please run it on this branch (release 5.3 + your change).

tox.ini
```diff
@@ -14,6 +15,20 @@ commands =
     python3 ./Tensile/bin/Tensile Tensile/Configs/build_client.yaml {envdir}/client
     py.test -v --basetemp={envtmpdir} --junit-xml={toxinidir}/python_tests.xml --junit-prefix={envname} --color=yes -n 4 --prebuilt-client={envdir}/client/0_Build/client/tensile_client {posargs}
 
+
+[testenv:{py35,py36,py38,py27}-win]
```
nakajee, line 19: It looks like this part is added in #1537 (which is not included in 5.3). 
We do not need it for this change.

AlexBrownAMD:
> Please run dtl_tsgr_sgemm.yaml on this branch (release 5.3 + your change) and make sure it can pass.

Just tried it locally: The test can not run because ThreadSeparateGlobalRead was added after 5.3. The test file should probably not be included with this patch.

Tensile/Tests/extended/direct_to_lds/dtl_sgemm.yaml
```diff
@@ -1,5 +1,5 @@
 TestParameters:
-  marks: [skip-gfx900, skip-gfx906, skip-gfx1010, skip-gfx1011, skip-gfx1012, skip-gfx1030] # not supported by arch
+  marks: [skip-gfx900, skip-gfx906, skip-gfx1010, skip-gfx1011, skip-gfx1012, skip-gfx1030, skip-gfx1100, skip-gfx1101, skip-gfx1102] # not supported by arch
```
nakajee, line 2: This also includes change from other commit (#1557).
It is probably better to add your change on top of 5.3 release code based on your code change log.

tox.ini
AlexBrownAMD: Removed this section

Tensile/Tests/extended/direct_to_lds/dtl_sgemm.yaml
AlexBrownAMD: Extra skips in this test case shouldn't break anything

nakajee: > Extra skips in this test case shouldn't break anything

I understand that. But, I think it is better to make your change as simple as possible.

AlexBrownAMD: Updated

tox.ini
```diff
@@ -1,5 +1,6 @@
 [tox]
-envlist = py35,py36,py27,lint
+envlist = {py35,py36,py38,py27}{,-win},lint
```
nakajee, line 2: {,-win} is also added in #1537. You do not need to add it here.

nakajee:
> > Please run dtl_tsgr_sgemm.yaml on this branch (release 5.3 + your change) and make sure it can pass.
> 
> Just tried it locally: The test can not run because ThreadSeparateGlobalRead was added after 5.3. The test file should probably not be included with this patch.

Thanks for confirmation.
Please also remove dtl_tsgr_dgemm.yaml and dtl_tsgr_hgemm.yaml under disabled folder.

tox.ini
AlexBrownAMD: Updated

nakajee:
The latest change looks OK.
Just confirmation.
Are we going to overwrite release/rocm-rel-5.3 for 5.3.1 (instead of creating a separate branch for 5.3.1)?

benjaminulmer:
> The latest change looks OK. Just confirmation. Are we going to overwrite release/rocm-rel-5.3 for 5.3.1 (instead of creating a separate branch for 5.3.1)?

Yes, we don't create new branches for the dot releases

nakajee (approved):

benjaminulmer (approved):
Whenever we're ready we can go ahead and merge this and get the rocBLAS side started.

bragadeesh (approved):

PR #1611, opened 2022-10-31
New merge script

benjaminulmer:
This is the first pass at an updated version of the merge script. I'm opening this up as a draft PR to serve as a forum to discuss design decisions and request features and functionality. The code still needs some cleanup and polish before it should be considered for merging. 

### Motivation 
The current merge script doesn't integrate with Tensile and therefore cannot use its data structures. This makes the merge script lightweight and portable, but heavily limits the types of operations that can be done within. One example of this issue is with solution equality checking which is overly strict and often results in duplicate solutions in the merged logic files. 

### Features in current version
The base functionality if forked from the current merge script with two major changes
1. Solution equality checking is done using SolutionStructs instead of simple dictionary comparison. 
2. Support for the new dictionary format logic files in addition to the ordered list format.

### Potential features/topics for discussion
1. Option to update logic files during the merge process (could merge this tool with TensileUpdateLibrary)
2. Chance to revisit args and defaults (e.g. should arcturus still have special default force-merge value?)
3. Option for automatically handling Fp16AltImpl files and merging

### DIFF
--- a/Tensile/TensileMergeLibrary.py
+++ b/Tensile/TensileMergeLibrary.py
@@ -1,0 +1,418 @@
+from .SolutionStructs import Solution
+from .Common import printExit, restoreDefaultGlobalParameters, assignGlobalParameters, ensurePath
+from . import LibraryIO
+
+from copy import deepcopy
+from enum import IntEnum
+from typing import NamedTuple
+import yaml
+import os
+import sys
+import argparse
+
+verbosity = 1
+
+class Logic:
+    def __init__(self, sols, sizes):
+        self.sols = sols
+        self.sizes = sizes
+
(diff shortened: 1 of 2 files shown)


babakpst (approved):
@benjaminulmer 
I support this change. In terms of the code, the new script is very similar to our old script, and I don't expect that breaks anything in Tensile/rocBLAS. Currently, merge.py is the only script we have in Utility, and moving it to Tensile/bin (and removing/deleting Utilities) is a good idea.
 
Since this is a standalone script and does not interfere with any other part of the code, I suggest you submit the changes so that I (and other active people in the tunning process) test the script in practice and double-check the output before we deprecate the old one.

PR #1612, opened 2022-11-01
Added asm caching

lringham:
Added asm cap caching and error checking to detect if the asm caps change.
- Passed Linting and Host Library Tests


Tensile/Common.py
```diff
@@ -1720,68 +1737,76 @@ def locateExe( defaultPath, exeName ): # /opt/rocm/bin, hip-clang
 
 def GetAsmCaps(isaVersion):
   """ Determine assembler capabilities by testing short instructions sequences """
-  rv = {}
-  rv["SupportedISA"]          = tryAssembler(isaVersion, "")
-  rv["HasExplicitCO"]         = tryAssembler(isaVersion, "v_add_co_u32 v0,vcc,v0,1")
-  rv["HasExplicitNC"]         = tryAssembler(isaVersion, "v_add_nc_u32 v0,v0,1")
+  if globalParameters["AssemblerPath"] != None:
```
benjaminulmer, line 1740:
```suggestion
  if globalParameters["AssemblerPath"] is not None:
```
It is recommended to use is/is not for None checking

benjaminulmer (approved):
Code looks good besides one nitpick.

We should probably figure out the process/tools for updating the cache and document it somewhere. Did you just generate the asmCaps and then dump to text to generate the cache?

lringham:
> Code looks good besides one nitpick.
> 
> We should probably figure out the process/tools for updating the cache and document it somewhere. Did you just generate the asmCaps and then dump to text to generate the cache?

Yes, that is exactly how I generated the cache. Maybe we need a function that generates and prints the caps?

PR #1613, opened 2022-11-04
Define array localMemory of type COMPUTE_DATA_TYPE in LDS to avoid precision loss.

amgddm:
The shared memory localMemory has two usages if localSplitU is enabled:
1. Storing matrix A and B of type DATA_TYPE from global memory.
2. Storing the intermediate results from registers of type COMPUTE_DATA_TYPE.

In the original implementation, the array localMemory is defined as of type DATA_TYPE. When register variables of type COMPUTE_DATA_TYPE are assigned to localMemory, the precisions are lost. The array localMemory is defined as an array of type COMPUTE_DATA_TYPE to keep the precision of register variables. It is type cast to a pointer to DATA_TYPE when it used a temporary software cache to store global matrix A and B.

### DIFF
--- a/Tensile/KernelWriterSource.py
+++ b/Tensile/KernelWriterSource.py
@@ -1046,7 +1046,7 @@ class KernelWriterSource(KernelWriter):
     # allocate local memory
     kStr += self.endLine
     kStr += "  /* allocate local memory */" + self.endLine
-    kStr += "  %sDATA_TYPE localMemory[LDS_NUM_ELEMENTS];%s" \
+    kStr += "  %sCOMPUTE_DATA_TYPE localMemory[LDS_NUM_ELEMENTS];%s" \
         % (self.sharedDeclStr, self.endLine )
 
 
@@ -2525,7 +2525,7 @@ class KernelWriterSource(KernelWriter):
       for sPerp in range(0, tP["nwpv"]):
         for para in range(0, tP["nrc"]):
           for sPara in range(0, 1): # tP["nwcv"]):
-            kStr += "%slocalWrite%s_%u_%u_%u_%u = (%sDATA_TYPE *)(localMemory + localWriteOffset%s_%u_%u_%u_%u);%s"\
+            kStr += "%slocalWrite%s_%u_%u_%u_%u = (%sDATA_TYPE *)localMemory + localWriteOffset%s_%u_%u_%u_%u;%s"\
                 % (self.indent, tP["tensorChar"], \
                 para, sPara, perp, sPerp, self.sharedPtrStr, tP["tensorChar"], \
                 para, sPara, perp, sPerp, self.endLine)
@@ -2604,7 +2604,7 @@ class KernelWriterSource(KernelWriter):
   ##############################################################################
   def localReadInitPointers(self, kernel, tP):
     kStr = ""
-    kStr += "%slocalRead%s = (%sDATA_TYPE *)(localMemory + localReadOffset%s);%s" \
+    kStr += "%slocalRead%s = (%sDATA_TYPE *)localMemory + localReadOffset%s;%s" \
         % (self.indent, tP["tensorChar"], self.sharedPtrStr, \
         tP["tensorChar"], self.endLine)
     return kStr
@@ -2755,7 +2755,7 @@ class KernelWriterSource(KernelWriter):
   ##############################################################################
   def localSplitULocalWrite(self, kernel):
     kStr = ""
-    kStr += "  %sDATA_TYPE *localLocalSplitU = (%sDATA_TYPE *)(localMemory);%s" \
+    kStr += "  %sCOMPUTE_DATA_TYPE *localLocalSplitU = (%sCOMPUTE_DATA_TYPE *)(localMemory);%s" \
       % (self.sharedPtrStr, self.sharedPtrStr, self.endLine)
     for j in range(0, kernel["ThreadTile1"] // kernel["VectorWidth"]):
       for i in range(0, kernel["ThreadTile0"] // kernel["VectorWidth"]):


Tensile/KernelWriterSource.py
```diff
@@ -1046,7 +1046,7 @@ def allocateResources(self, kernel):
     # allocate local memory
     kStr += self.endLine
     kStr += "  /* allocate local memory */" + self.endLine
-    kStr += "  %sDATA_TYPE localMemory[LDS_NUM_ELEMENTS];%s" \
+    kStr += "  %sCOMPUTE_DATA_TYPE localMemory[LDS_NUM_ELEMENTS];%s" \
```
nakajee, line 1049: Would you please keep LDS allocation size unchanged for LocalSplitU = 1 case?
LDS allocation size will be doubled in HHS case with LocalSplitU = 1 if we use computeDataType here.

I suggest to modify ldsNumElementsReduction part.
Something like
  ldsNumElementsReduction *= sizeof(computeDataType) / sizeof(DataType)

nakajee: pre_checkin/igemm_hpa_hip_nn.yaml seems to be failed due to this change.

b"error: local memory (99712) exceeds limit (65536) in function 'Cijk_Ailk_Bljk_I8II_BH_MT128x128x32_SN_GRVW8_GSU1_K1_LPA3_LPB3_LRVW8_TT8_8_VW8_WG16_16_1'\nerror: local memory (99712) exceeds limit (65536) in function 'Cijk_Ailk_Bljk_I8II_BH_MT128x128x32_SN_GRVW8_GSU3_K1_LPA3_LPB3_LRVW8_TT8_8_VW8_WG16_16_1'\n2 errors generated when compiling for gfx1010.\n"

amgddm: Updated. @nakajee

Tensile/KernelWriterSource.py
```diff
@@ -1046,7 +1046,11 @@ def allocateResources(self, kernel):
     # allocate local memory
     kStr += self.endLine
     kStr += "  /* allocate local memory */" + self.endLine
-    kStr += "  %sDATA_TYPE localMemory[LDS_NUM_ELEMENTS];%s" \
+    if kernel["LocalSplitU"] > 1:
```
nakajee, line 1049: Even though we add if statement here, the LDS size calculation is still inaccurate for LocalSplitU > 1 case.

In SolutionStruct.py, ldsNumElements is calculated in different ways and use the max of these values.

    # lds size is the greater of the two
    ldsNumElements = max(ldsNumElementsAB, ldsNumElementsReduction, ldsNumElementsOccupancy)
     ...
      ldsNumElements = max(ldsNumElements, ldsNumElementsRemapC)

    state["LdsNumElements"] = ldsNumElements

computeDataType is not always necessary even if LocalSplitU > 1.
Would you please use computeDataType only for necessary scenario (which should be ldsNumElementsReduction)?

nakajee: In case of ldsNumElementsRemapC , we do the following adjustment in case size of DestDataType is different from size of DataType
===
      if state["_GlobalAccumulation"]:
        computeBytes = state["ProblemType"]["ComputeDataType"].numBytes()
        multiplier = computeBytes // state["ProblemType"]["DataType"].numBytes()
      elif state["ProblemType"]["DestDataType"].numBytes() > state["ProblemType"]["DataType"].numBytes():
        # Determine ratio of output to input element size.
        # SRVW remaps output so we need to scale up resources.
        multiplier = state["ProblemType"]["DestDataType"].numBytes() // state["ProblemType"]["DataType"].numBytes()
      else:
        multiplier = 1

      ldsNumElementsRemapC *= multiplier
===

We should need similar logic for ldsNumElementsReduction (just need to multiply sizeof(computeDataType) / sizeof(DataType) as I suggested).

amgddm: @nakajee You are right. Instead of changing localsplitU reduction part, we may replace the ldsNumElementsReduction =  state["LocalSplitU"]*state["MacroTile0"]*state["MacroTile1"]
with 
ldsNumElementsReduction =  sizeof(COMPUTE_DATA_TYPE)/sizeof(DATA_TYPE)*state["LocalSplitU"]*state["MacroTile0"]*state["MacroTile1"]
and define localMemory as
__shared__ DATA_TYPE localMemory[LdsNumElements];

amgddm: @nakajee We don't need an additional if statement to deal with LocalSplitU >1 and LocalSplitU == 1

Tensile/KernelWriterSource.py
amgddm: @nakajee Updated.

Tensile/KernelWriterSource.py
amgddm: local@nakajee There are three changes:
1. ldsNumElementsReduction should multiply by sizeof(COMPUTE_DATA_TYPE)/sizeof(DATA_TYPE)
2. localMemory is defined as type DATA_DATA which is the same as the original implementaton.
3. localMemory is typed cast to COMPUTE_DATA_TYPE when it is used to store intermediate results.

Tensile/SolutionStructs.py
```diff
@@ -3606,7 +3606,11 @@ def assignDerivedParameters(state):
       ldsNumElementsAB = ldsNumElementsAlignedA + ldsNumElementsB
 
     # lds buffer size for reduction
-    ldsNumElementsReduction = state["LocalSplitU"]*state["MacroTile0"]*state["MacroTile1"] if state["LocalSplitU"] > 1 else 0
+    bytesPerComElem = state["ProblemType"]["ComputeDataType"].numBytes()
+    bytesPerLoadElem = state["ProblemType"]["DataType"].numBytes()
+    ratio = bytesPerComElem / bytesPerLoadElem
+
+    ldsNumElementsReduction = ratio*state["LocalSplitU"]*state["MacroTile0"]*state["MacroTile1"] if state["LocalSplitU"] > 1 else 0
```
nakajee, line 3613: Please convert this type to integer.
It causes an error in CI test.


  File "/var/jenkins_home/workspace/clang_precheckin_Tensile_PR-1613/uj7lkGv2W/tensile/Tensile/KernelWriterAssembly.py", line 3017, in allocateResources

    kStr += inst("s_mov_b32", "m0", hex(kernel["LdsNumElements"] \

TypeError: 'float' object cannot be interpreted as an integer

nakajee:
The code change looks OK.
Please update the title of this PR to align with the latest change.

nakajee:
Type cast is for localLocalSplitU, not localMemory.

nakajee (approved):
CI tests look OK.
I asked @AlexBrownAMD to confirm 2 fails (actually passed) in extended cases.

PR #1614, opened 2022-11-04
Added command line arg to ignore asm cap cache

lringham:


Tensile/Common.py
```diff
@@ -1800,13 +1802,12 @@ def GetAsmCaps(isaVersion):
 
     derivedAsmCaps["SupportedSource"] = True
 
-    if derivedAsmCaps != CACHED_ASM_CAPS[isaVersion]:
```
AlexBrownAMD, line 1803: Where is the cache stored/defined?

lringham: https://github.com/ROCmSoftwarePlatform/Tensile/blob/e7f2f2a3382903ec912eae5d6ecb5900b9477daa/Tensile/Common.py#L283

AlexBrownAMD (approved):

PR #1617, opened 2022-11-09
Remove xfail on fixed hpa_source tests

AlexBrownAMD:
To be merged after CI rocm version updated


nakajee:
The change looks OK.
Could you confirm why CI test is failing?

AlexBrownAMD:
The test failure was due to a compiler bug introduced in rocm 5.3. The test was marked xfail until a compiler update was ready. The current rocm build on CI has the fix in it, but CI is soon moving to another build that does not have the fix in it (so the test will again be expected to fail). We can merge this as soon as CI returns to a build with the patch (or sooner if we don't mind seeing a known test failure in builds)

nakajee (approved):

PR #1625, opened 2022-11-22
Add dual mac instruction for gfx11.

hcman2:
This is new instructions to boost mac operations in gfx11 platform.


Tensile/Components/MAC_F32.py
```diff
@@ -67,31 +72,86 @@ def __call__(self, writer, m, innerUnroll):
...
+                        if macIdx == kernel["PerformanceSyncLocation"]:
+                            # pop all instructions
+                            while instQ.qsize() > 0:
+                                prevVars = instQ.get()
+                                kStr += "v_fmac_f32 {cStr}, {aStr}, {bStr}{endLine}".format_map(prevVars)
+                                kStr += priority(writer, 1, "Raise priority while processing macs")
+                            kStr += "s_barrier // extra barrier for performance{endLine}".format_map(vars)
+                        macIdx += 1
+            # pop all instructions
+            while instQ.qsize() > 0:
```
nakajee, line 125: Can we have a function for popping all instructions to avoid having duplicated code?

Tensile/Components/MAC_F32.py
```diff
@@ -67,31 +72,86 @@ def __call__(self, writer, m, innerUnroll):
...
+                                instQ.get()
+                            else:
+                                # push instruction
+                                instQ.put(instVars)
+
+                        if macIdx == kernel["PerformanceWaitLocation"]:
+                            # pop all instructions
+                            while instQ.qsize() > 0:
+                                prevVars = instQ.get()
+                                kStr += "{instruction} {cStr}, {aStr}, {bStr}{endLine}".format_map(prevVars)
```
nakajee, line 113: Strictly speaking, we should have a if check (if instruction == "v_fma_f32") here to generate correct code in case of "v_fma_f32".
We can create a sub-function to unify this piece of code.

Tensile/Components/MAC_F32.py
```diff
@@ -67,31 +72,86 @@ def __call__(self, writer, m, innerUnroll):
...
+                                kStr += priority(writer, 1, "Raise priority while processing macs")
+                            kStr += "s_waitcnt lgkmcnt({PerformanceWaitCount}) // extra wait for performance{endLine}".format_map(vars)
+                        if macIdx == kernel["PerformanceSyncLocation"]:
+                            # pop all instructions
+                            while instQ.qsize() > 0:
+                                prevVars = instQ.get()
+                                kStr += "v_fmac_f32 {cStr}, {aStr}, {bStr}{endLine}".format_map(prevVars)
```
nakajee, line 120: This should be instruction instead of using fixed "v_fmac_f32".
If we would like to always use "v_fmac_f32". we should set instruction = "v_fmac_f32" and check asmCap before using it.

Tensile/Components/MAC_F32.py
```diff
@@ -67,31 +72,86 @@ def __call__(self, writer, m, innerUnroll):
...
+                        vars["idx1"] = idx1
+                        vars["a"] = idx0 if writer.tPB["tile01Idx"] else idx1
+                        vars["b"] = idx1 if writer.tPB["tile01Idx"] else idx0
+                        vars["iui"] = iui
+
+                        vars["cStr"] = "v[vgprValuC + {idx0} + {idx1}*{ThreadTile0}]".format_map(vars)
+                        vars["aStr"] = "v[vgprValuA_X{m}_I{iui} + {a}]".format_map(vars)
+                        vars["bStr"] = "v[vgprValuB_X{m}_I{iui} + {b}]".format_map(vars)
+
+                        instVars = {}
```
nakajee, line 90: Can we unify the for loops?
The difference is mac code generation part only.

We can have a if statement inside for loops (like this).

if dualMacEnable == 1:
    (dual mac generation part)
    instVars....
else:
    (existing mac generation)

We can always define instQ and keep it empty for dualMacEnable=0 case.
Then, We can unify PerformanceWaitLocation and PerformanceSyncLocation code.
if PerformanceSyncLocation:
   call pop all instructions function (does nothing for dualMacEnable=0)
   PerformanceWaitCount code

Tensile/Components/MAC_F32.py
```diff
@@ -47,6 +48,10 @@ def __call__(self, writer, m, innerUnroll):
             else:
                 raise RuntimeError("FMA instruction specified but not supported on {}".format(kernel["ISA"]))
 
+        dualMacEnable = 0
+        if writer.asmCaps["v_dual_fmac_f32"] and kernel["WavefrontSize"] == 32:
```
nakajee, line 52: Why WavefrontSize needs to be 32 to generate v_dual_fmac_f32?

hcman2: According to the spec, this instruction format is only legal for wave32.

Tensile/Components/MAC_F32.py
hcman2: Good suggestion. I will try to unify the code.

Tensile/Components/MAC_F32.py
nakajee: > According to the spec, this instruction format is only legal for wave32.

OK. 
I probably need to ask this in a different way.
If it is not legal, I guess asmCap will be false.
I would like to know why we need to check an extra condition besides asmCap.
I feel just checking asmCap should be enough.

hcman2: We set asmCap without checking wave32/64. The gfx11 has both wave32 and wave64 mode. Generally, we will use wave32 mode in gfx11. asmCap will be always true on gfx11 platform.

Tensile/Components/MAC_F32.py
```diff
@@ -80,19 +87,56 @@ def __call__(self, writer, m, innerUnroll):
...
-
-                    kStr += priority(writer, 1, "Raise priority while processing macs")
+                        if instruction == "v_fma_f32":
+                            kStr += "v_fma_f32 {cStr}, {aStr}, {bStr}, {cStr}{endLine}".format_map(vars)
+                        else:
+                            kStr += "{instruction} {cStr}, {aStr}, {bStr}{endLine}".format_map(vars)
 
                     if macIdx == kernel["PerformanceWaitLocation"]:
+                        kStr += self.popAllInstructions(instruction, instQ, priority, writer)
+                        instQ.queue.clear()
```
nakajee, line 118: I think clear() is redundant.
popAllInstructions() ensures the size of instQ is 0.

Tensile/Components/MAC_F32.py
```diff
@@ -80,19 +87,56 @@ def __call__(self, writer, m, innerUnroll):
                     vars["aStr"] = "v[vgprValuA_X{m}_I{iui} + {a}]".format_map(vars)
                     vars["bStr"] = "v[vgprValuB_X{m}_I{iui} + {b}]".format_map(vars)
 
-                    if instruction == "v_fma_f32":
-                        kStr += "v_fma_f32 {cStr}, {aStr}, {bStr}, {cStr}{endLine}".format_map(vars)
+                    instVars = {}
```
nakajee, line 90: You can move instVars related code after if dualMacEnable == 1.
It is unnecessary for dualMacEnable == 0.

Tensile/Components/MAC_F32.py
```diff
@@ -67,9 +72,11 @@ def __call__(self, writer, m, innerUnroll):
         priority = Component.Priority.find(writer)
         macIdx = 0
 
+        instQ = queue.Queue()
+
         for iui in range(0, innerUnroll):
-            for idx1 in range(0, kernel["ThreadTile1"]):
```
nakajee, line 71: No need to change this part.

Tensile/Components/MAC_F32.py
hcman2: Thanks. The queue object is mutable.

nakajee:
Thanks for your update.
I do not have any further comments on the code change.

Do we have some test cases in pre_checkin or extended to verify this feature?

hcman2:
> Thanks for your update. I do not have any further comments on the code change.
> 
> Do we have some test cases in pre_checkin or extended to verify this feature?

hcman2:
Test cases added in pre_checkin/wave32.

nakajee:
Thanks for adding the test cases.
It looks like we need to add the new folder name in pytest.ini to avoid the following warning message.
===
  /var/jenkins_home/workspace/clang_precheckin_Tensile_PR-1625/4dRKDTY1u/tensile/Tensile/Tests/yaml_only/test_config.py:76: PytestUnknownMarkWarning: Unknown pytest.mark.wave32 - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/stable/how-to/mark.html
    return getattr(pytest.mark, name)

nakajee (approved):

PR #1626, opened 2022-11-24
Print failed build commands in better format for debugging and testing

AlexBrownAMD:


Tensile/KernelWriter.py
```diff
@@ -5098,7 +5099,12 @@ def getSourceFileString(self, kernel):
         return (0, self.getKernelSource(kernel))
 
     except subprocess.CalledProcessError as exc:
-      print(exc)
+      if isinstance(exc.cmd, collections.Sequence):
```
nakajee, line 5102: I am not familiar with error handling code.
1) Why we need to check the type of exc.cmd? Can it be a different type?
2) How the output message changed (before and after)?
3) It looks like exc also has output (or stdout, stderr). Don't we need to print it as well?

AlexBrownAMD: subprocess accepts arguments as either a string or a sequence, most places in our code pass a list, but the check ensures both cases are treated correctly.
When the command is a list, the default stringification for the exception prints it to look like a python list like
Command ['cmd', '-arg1', 'val1', ...]
To make it easier to debug, this change reformats the list to the actual command-line string so you can easily copy-paste and re-run the failed command by itself.
Printing stdout+stderr is already taken care of elsewhere in the code.

nakajee: Thanks for your quick response.
Waiting for comments(or approval) from others.

nakajee:
Is the preCheckin fail due to your change?

AlexBrownAMD:
> Is the preCheckin fail due to your change?

I don't think so. Looks like it timed out on 908, which has been happening sporadically on CI for a bit now

nakajee:
OK. It might be better to re-run pre_checkin tests to make sure this is not caused by your change.
I did not see similar failure in other PR.

nakajee:
> OK. It might be better to re-run pre_checkin tests to make sure this is not caused by your change. I did not see similar failure in other PR.

The pre_checkin test seems OK this time.

I would like to have some comment/approval from other reviewers.

nakajee (approved):

PR #1629, opened 2022-11-29
Fix solution index collisions

benjaminulmer:
Fixes a solution index collision that was happening with lazy-loading.

This current version still produces collisions between different architectures, which should probably be fixed at some point, but for now this fixes the main issue of collisions in the same architecture (or at least it does based on my testing).

### DIFF
--- a/Tensile/SolutionLibrary.py
+++ b/Tensile/SolutionLibrary.py
@@ -458,7 +458,7 @@ class MasterSolutionLibrary:
     def merge(self, other, startIndex=0):
         assert self.__class__ == other.__class__
 
-        curIndex = max(startIndex, max(self.solutions.keys()) + 1) if self.solutions else 0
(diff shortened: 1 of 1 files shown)


AlexBrownAMD (approved):

Tensile/SolutionLibrary.py
```diff
@@ -458,7 +458,7 @@ def remapSolutionIndicesStartingFrom(self, curIndex):
     def merge(self, other, startIndex=0):
         assert self.__class__ == other.__class__
 
-        curIndex = max(startIndex, max(self.solutions.keys()) + 1) if self.solutions else 0
+        curIndex = max(startIndex, max(self.solutions.keys()) + 1 if self.solutions else 0)
```
nakajee, line 461: If we make the change in this way, isn't self.solutions.keys()  called even if self.solutions?
Is that OK?

PR #1630, opened 2022-11-30
WMMA: fix h beta vectorwidth4 correctness issue

TonyYHsieh:
one register contains two values in h datatype, so we should divide index by 2 instead of gwvw to get correct vgpr index.


hcman2 (approved):

cmingch (approved):
look good to me

PR #1634, opened 2022-12-02
Script to remove kernels from a logic file that are not used

AlexBrownAMD:
Script to remove kernels from a logic file that are not used by any of the tuned sizes

### DIFF
--- a/tuning/automation/RemoveUnusedKernels.py
+++ b/tuning/automation/RemoveUnusedKernels.py
@@ -1,0 +1,82 @@
+################################################################################
+#
+# Copyright (C) 2016-2022 Advanced Micro Devices, Inc. All rights reserved.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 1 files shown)


babakpst:
@AlexBrownAMD Can you please remind me again why this script is needed? When we update library logics, merge.py will check if there are unused kernels in the libraries. Do we need to separately check unused kernels in any logic file?

babakpst:
I have the same error. I guess this script is related to #1631

AlexBrownAMD:
Yes this is related to #1631 . This logic file contains an error:
https://github.com/ROCmSoftwarePlatform/rocBLAS/blob/develop/library/src/blas3/Tensile/Logic/asm_full/aldebaran/aldebaran_Cijk_Ailk_Bljk_DB_GB.yaml
The first kernel in that file fails to generate. Rocblas build is not reporting any errors for it right now, but with #1631 enabled it will start failing. Luckily this invalid kernel is also not used at all. If the merge script is supposed to remove unused kernels then I'm not sure how this one got through. My script may not be needed in the general repo, but I am currently using to fix errors in the tuned files.
As a side-effect, the script can also fix incorrect indexing, like in this file:
https://github.com/ROCmSoftwarePlatform/rocBLAS/blob/develop/library/src/blas3/Tensile/Logic/asm_full/arcturus/arcturus_Cijk_Ailk_Bjlk_HB.yaml
Here, the values for SolutionIndex are repeated (more than one kernel numbered 0, 1, 2, etc).
I haven't tried running the script on all the logic files yet, so there are probably other cases like the error you posted - I can update the script to be a little more error-resistant if we want to keep it checked in the repo.

nakajee:
I would like to know how we use this in the future.
Are we going to enable some kind of regular tests to remove unused kernel from rocblas logic files?
At this moment, I am not sure if we need to keep this script in Tensile repo.

nakajee:
I also had an impression that removing unused kernels were done by merge script.
Does that mean there is a bug in merge script?
Anyway, I understand that we currently have a lot of unused kernels in rocblas logic files.
It will be good to clean them up with this script.

babakpst (changes requested):
Blocked until after FC merge.

babakpst (changes requested):
Blocked until FC merges are complete.

babakpst:
Closing this PR, as this is not a priority now.

PR #1644, opened 2022-12-13
Add build-id to assembly kernel builds

AlexBrownAMD:


yoichiyoshida (approved):

babakpst (approved):

TorreZuk:
Assuming this may be wanted also as hotfix for 5.5 to match the source kernel build-id.

Tensile/KernelWriterAssembly.py
```diff
@@ -250,6 +250,11 @@ def getCompileArgs(self, sourceFileName, objectFileName, *moreArgs, isa=None, wa
   def getLinkCodeObjectArgs(self, objectFileNames, coFileName, *moreArgs):
     rv = [globalParameters['AssemblerPath'],
           '-target', 'amdgcn-amd-amdhsa']
+    # Add build-id for builds with rocm 5.3+
+    compilerVer = globalParameters['HipClangVersion'].split(".")[:2]
+    compilerVer = [int(c) for c in compilerVer]
+    if len(compilerVer) >= 2 and (compilerVer[0] > 5 or (compilerVer[0] == 5 and compilerVer[1] > 2)):
+      rv += ["-Xlinker", "--build-id"]
```
nakajee, line 257: Do we need to specify this for os.name=="nt" case?
At line 120 of TensileCreateLibrary.py, parameters are set without using this function.
Should we add --build-id option separately?

nakajee:
Why CI tests are failing?

Tensile/KernelWriterAssembly.py
AlexBrownAMD: Updated the nt code to also call this function

PR #1647, opened 2022-12-19
Change the second kernel to improve the performance of multiple-buffer implementaion when GSU is enabled.…

amgddm:
The current multiple-buffer implementation of GSU is not optimal. Because the implementation of the second kernel  for accumulating the GSU partial summation results is simple. A tuned version with loop unroll is implemented.  The tuned version has better performance compared to old multiple-buffer and single-buffer implementation.

The tests are performed on MI250 in auto mode. The GEMM format is HSS. 
[PerfAtomicCASMB.xlsx](https://github.com/ROCmSoftwarePlatform/Tensile/files/10255420/PerfAtomicCASMB.xlsx)

### DIFF
--- a/Tensile/KernelWriterConversion.py
+++ b/Tensile/KernelWriterConversion.py
@@ -236,8 +236,8 @@ class KernelWriterConversion(KernelWriterBase):
       indexChar = self.indexChars[i]
       kStr += " + (size%s - 1) * strideW%s" % (indexChar, indexChar)
     kStr += ";" + self.endLine
-
     kStr += "  " + self.datatype + " accum = 0;%s" % self.endLine
+    kStr += "#pragma unroll%s" % self.endLine
     kStr += "  for (int i=0; i<gsu; i++) {%s" % self.endLine
     kStr += "    accum += W[idxW];%s" % self.endLine
     kStr += "    idxW  += strideW;%s" % self.endLine


nakajee:
Do we have any way to let complier do loop-unroll?

Tensile/KernelWriterConversion.py
```diff
@@ -236,11 +236,19 @@ def kernelBody(self):
...
     kStr += "  " + self.datatype + " accum = 0;%s" % self.endLine
-    kStr += "  for (int i=0; i<gsu; i++) {%s" % self.endLine
+    kStr += "  " + self.datatype + " accum1 = 0;%s" % self.endLine
+    kStr += "  int res_gsu = gsu & 1;%s" % self.endLine
+    kStr += "  int gsu_end = gsu - res_gsu;%s" % self.endLine
+    kStr += "  for (int i=0; i<gsu_end; i+=2) {%s" % self.endLine
     kStr += "    accum += W[idxW];%s" % self.endLine
-    kStr += "    idxW  += strideW;%s" % self.endLine
+    kStr += "    accum1 += W[idxW+strideW];%s" % self.endLine
+    kStr += "    idxW  += 2*strideW;%s" % self.endLine
```
nakajee, line 247: As long as I checked, the compiler generates 2 * strideW inside for loop.
We can move 2 * strideW out of for loop by creating another variable like this.
   int strideW2 = strideW * 2;
   for ....

amgddm: Updated.

Tensile/KernelWriterConversion.py
```diff
@@ -236,11 +236,19 @@ def kernelBody(self):
       indexChar = self.indexChars[i]
       kStr += " + (size%s - 1) * strideW%s" % (indexChar, indexChar)
     kStr += ";" + self.endLine
-
+
```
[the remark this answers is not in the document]
babakpst: Githooks will take care of formatting and trailing spaces.

Tensile/KernelWriterConversion.py
babakpst: Unrolling may help here. I am trying this change for other sizes. However, GSU is a small number when GSU algo is a MB (due to memory limitations). Thus, this for loop might not impose a huge overhead.

Tensile/KernelWriterConversion.py
```diff
@@ -236,11 +236,19 @@ def kernelBody(self):
       indexChar = self.indexChars[i]
       kStr += " + (size%s - 1) * strideW%s" % (indexChar, indexChar)
     kStr += ";" + self.endLine
-
     kStr += "  " + self.datatype + " accum = 0;%s" % self.endLine
-    kStr += "  for (int i=0; i<gsu; i++) {%s" % self.endLine
+    kStr += "  " + self.datatype + " accum1 = 0;%s" % self.endLine
+    kStr += "  int res_gsu = gsu & 1;%s" % self.endLine
+    kStr += "  int gsu_end = gsu - res_gsu;%s" % self.endLine
+    kStr += "  uint64_t strideW2 = strideW << 1;%s" % self.endLine
```
nakajee, line 243: Better to use exactly same way as strideW.
->use self.uint64Str

babakpst (approved):

PR #1658, opened 2023-01-26
Change the range of AssertSummationElementMultiple.

amgddm:


nakajee (commented):
The change looks OK.
Waiting for completing CI tests.

nakajee (approved):
CI test looks OK (908 extended test timeout is not caused by this change)

PR #1663, opened 2023-02-02
CI: Increase test timeout in precheckin and extended

eidenyoshida:


nakajee (approved):
Looks good as a temporary solution.
We need to keep in mind that we need to change the timeout back after reducing total amount of time (or get 908 node faster...)

PR #1665, opened 2023-02-04
Joblib parallel map

nielenventer:


nielenventer:
Tensile uses a large amount of memory when loading and writing kernels and can cause CI to fail if total memory usage gets too high. This is increasingly a problem as library sizes grow. Joblib is a replacement for the multiprocessing calls, which seems to do a better job of keeping the memory usage down.

Plot showing system memory usage during rocBLAS install:
grid

Note that during kernel write phase (about halfway) mem usage peaks at over 100GiB.

Plot showing same thing but with this change:
grid-joblib

Memory usage during the same phase is now flat, at a much lower value.

nielenventer:
NOTE: This adds a new dependency to Tensile.

nielenventer:
I think failing tests from before were unrelated to this change.. Seems to have been resolved with other PRs.

yenong-amd (approved):

PR #1667, opened 2023-02-08
Fix unescaped $PATH variable

lawruble13:
This was causing Tensile to build using the wrong version of Python, as it was using the PATH from the host Groovy environment instead of the local Docker environment.

### DIFF
--- a/.jenkins/common.groovy
+++ b/.jenkins/common.groovy
@@ -61,7 +61,7 @@ def runCompileCommand(platform, project, jobName, boolean debug=false)
             mkdir build
             pushd build
 
-            export PATH=/opt/rocm/bin:$PATH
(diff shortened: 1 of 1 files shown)


lawruble13:
This is blocking #1665.

TorreZuk (approved):
Assume it works if it gets to the shell  :-)

lawruble13:
Is the issue we're seeing in CI the intermittent failure we've observed previously? If so, can I merge this in?

TorreZuk:
If it fixed what you wanted I don't see why not, maybe @eidenyoshida you can quickly give second review to allow merge.

eidenyoshida (approved):
Looks good, thanks

PR #1677, opened 2023-02-24
Turn off merge files to speed up some slow pre_checkin tests

AlexBrownAMD:
igemm_hpa_hip_nn drops from > 12 minutes to < 4 minutes
bfloat16s_hpa_source_tt drops from ~ 2.5 minutes to < 1 minute


yoichiyoshida (approved):

nakajee (approved):
That is great!
I did not know we have such an option.
I am OK as long as these tests pass.

PR #1680, opened 2023-02-28
Speed up extended test runtime

AlexBrownAMD:
Cuts extended test runtime by a few hours by disabling merge files for source kernels with many variations and large file size, and removing some test cases that are nearly duplicates.


Tensile/Tests/extended/direct_to_lds/dtl_sgemm.yaml
```diff
@@ -453,61 +398,6 @@ BenchmarkProblems:
...
-        - KernelLanguage: ["Assembly"]
-        - EdgeType: ["ShiftPtr"]
-        - AtomicAddC: [False] 
-      ForkParameters:
-        - MatrixInstruction:
-          - [16, 16, 1, 4, 1, 2,2, 2,2] # 64x256
-          - [16, 16, 1, 4, 2, 2,2, 2,2] # 128x128
-          - [16, 16, 1, 4, 4, 2,2, 2,2] # 256x64
-
-          - [32, 32, 2, 1, 1, 2,2, 2,2] # 128x128
```
nakajee, line 468: The main purpose of this "MFMA variants" is to test MI32x32.
If we remove this part, no test for MI32x32.
I would suggest to move some of MI32x32 to the above test set.
If we would not like to add any combinations, we can replace the last 3 MI valuse with MI32x32.
Like, replace this
          - [16, 16, 4, 1, 1, 4,4, 4,1]  # 256x64
          - [16, 16, 4, 1, 1, 4,4, 1,4]  # 64x256
          - [16, 16, 4, 1, 1, 4,4, 2,2]  # 128x128
with
          - [32, 32, 2, 1, 1, 2,2, 2,2] # 128x128
          - [32, 32, 2, 1, 1, 2,2, 1,4] # 64x256
          - [32, 32, 2, 1, 1, 2,2, 4,1] # 256x64
Replacing only 1 location should be OK. No need to add MI32x32 to all test sets in this file.

Tensile/Tests/extended/direct_to_lds/dtl_sgemm.yaml
```diff
@@ -703,61 +593,6 @@ BenchmarkProblems:
...
```
nakajee, line 718: Same here and below

yoichiyoshida (approved):

Tensile/Tests/extended/direct_to_lds/dtl_sgemm.yaml
AlexBrownAMD: Good call. Just posted an update with 32x32 instructions added back in. I couldn't replace the 16x16 test cases like you suggested because the 32x32 instructions were rejected with those parameter combinations. So I added back in a simplified version of the section that I removed. The new section has combinations significantly reduced, just enough to make sure the instructions work with DirectToLds.

I took a closer look at the dtl_tsgr_sgemm test that was also changed, and wasn't able to add the 32x32 instructions back in. It looks like in the current version of this test those instructions are not tested with dtl+tsgr  anyway because that feature is not supported so all test cases get rejected. 32x32 with dtl is already tested in dtl_sgemm.yaml. Do we need test cases for 32x32 with tsgr (and no dtl)?

nakajee: Thanks for your update.
I will take a look at dtl+tsgr + MI32x32 later.
I am OK with the current change as long as they do not fail

nakajee (approved):

PR #1681, opened 2023-02-28
joblib dependency as apt-get instead of pip

nielenventer:
A recent change to setuptools has caused issues with some packages. When installing Tensile via the modified dockerfile, this causes an issue with devscripts. This change sidesteps the issue for now.

For detailed discussion see:
- https://github.com/pypa/setuptools/issues/3772
- https://bugs.launchpad.net/ubuntu/+source/devscripts/+bug/1991606

Note that the issue is not related to joblib itself. All other deps in the `pip3 install ... ` section are already installed and skipped. New deps, however, trigger devscripts and cause the issue. Changing joblib to any other (uninstalled) python packages causes the same error (e.g. pandas).

### DIFF
--- a/docker/dockerfile-build-ubuntu-rock
+++ b/docker/dockerfile-build-ubuntu-rock
@@ -46,6 +46,7 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-ins
     python3-pip \
     python3-pytest \
     python3-setuptools \
+    python3-joblib \
     python3-yaml \
     libnuma1 \
     libboost-all-dev \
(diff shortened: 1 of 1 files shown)


TorreZuk (approved):
LGTM but more for Eiden to review and be in the know

TorreZuk:
@amcamd  also should monitor

amcamd:
Thanks for this fix @nielenventer , it should fix [SWDEV-385910](https://ontrack-internal.amd.com/browse/SWDEV-385910)

amcamd (approved):

PR #1684, opened 2023-03-03
Fix offset calculation to prevent overflow if offset is really large

AlexBrownAMD:


amcamd (approved):
LGTM

Tensile/KernelWriterAssembly.py
```diff
@@ -3108,22 +3108,30 @@ def allocateResources(self, kernel):
       kStr += ".if 0\n"
 
     # add offset to buffer
+    tmpOffset = self.sgprPool.checkOutAligned(2, 2, preventOverflow=0)
+
     if not kernel["_GlobalAccumulation"]:
-      kStr += inst("s_lshl_b32", sgpr("OffsetD"), sgpr("OffsetD"), hex(log2(self.bpeCexternal)), "elements offset to bytes offset")
-      kStr += inst("s_add_u32",  sgpr("AddressD+0"), sgpr("AddressD+0"), sgpr("OffsetD"), "add offset to buffer address")
-      kStr += inst("s_addc_u32", sgpr("AddressD+1"), sgpr("AddressD+1"), 0, "add offset to buffer address")
+      kStr += inst("s_mov_b32", sgpr(tmpOffset), sgpr("OffsetD"), "copy to temp b64 to prevent overflow with large offset")
```
nakajee, line 3114: We should initialize tmpOffset+1 to 0 before using 64bit shift.

Tensile/KernelWriterAssembly.py
```diff
@@ -3108,22 +3108,30 @@ def allocateResources(self, kernel):
...
+      kStr += inst("s_lshl_b64", sgpr(tmpOffset, 2), sgpr(tmpOffset, 2), hex(log2(self.bpeCexternal)), "elements offset to bytes offset")
+      kStr += inst("s_add_u32",  sgpr("AddressD+0"), sgpr("AddressD+0"), sgpr(tmpOffset), "add offset to buffer address")
+      kStr += inst("s_addc_u32", sgpr("AddressD+1"), sgpr("AddressD+1"), sgpr(tmpOffset + 1), "add offset to buffer address")
+
+      kStr += inst("s_mov_b32", sgpr(tmpOffset), sgpr("OffsetC"), "copy to temp b64 to prevent overflow with large offset")
```
nakajee, line 3119: Same here. Initialize tmpOffset+1 = 0

Tensile/KernelWriterAssembly.py
```diff
@@ -3108,22 +3108,30 @@ def allocateResources(self, kernel):
...
+      kStr += inst("s_lshl_b64", sgpr(tmpOffset, 2), sgpr(tmpOffset, 2), hex(log2(self.bpeCexternal)), "elements offset to bytes offset")
+      kStr += inst("s_add_u32",  sgpr("AddressC+0"), sgpr("AddressC+0"), sgpr(tmpOffset), "add offset to buffer address")
+      kStr += inst("s_addc_u32", sgpr("AddressC+1"), sgpr("AddressC+1"), sgpr(tmpOffset + 1), "add offset to buffer address")
 
-      kStr += inst("s_lshl_b32", sgpr("OffsetC"), sgpr("OffsetC"), hex(log2(self.bpeCexternal)), "elements offset to bytes offset")
-      kStr += inst("s_add_u32",  sgpr("AddressC+0"), sgpr("AddressC+0"), sgpr("OffsetC"), "add offset to buffer address")
-      kStr += inst("s_addc_u32", sgpr("AddressC+1"), sgpr("AddressC+1"), 0, "add offset to buffer address")
+    kStr += inst("s_mov_b32", sgpr(tmpOffset), sgpr("OffsetA"), "copy to temp b64 to prevent overflow with large offset")
```
nakajee, line 3124: Same here. Initialize tmpOffset+1 = 0

Tensile/KernelWriterAssembly.py
```diff
@@ -3108,22 +3108,30 @@ def allocateResources(self, kernel):
...
+    kStr += inst("s_lshl_b64", sgpr(tmpOffset, 2), sgpr(tmpOffset, 2), hex(log2(self.bpeAB)), "elements offset to bytes offset")
+    kStr += inst("s_add_u32",  sgpr("AddressA+0"), sgpr("AddressA+0"), sgpr(tmpOffset), "add offset to buffer address")
+    kStr += inst("s_addc_u32", sgpr("AddressA+1"), sgpr("AddressA+1"), sgpr(tmpOffset + 1), "add offset to buffer address")
 
-    kStr += inst("s_lshl_b32", sgpr("OffsetA"), sgpr("OffsetA"), hex(log2(self.bpeAB)), "elements offset to bytes offset")
-    kStr += inst("s_add_u32",  sgpr("AddressA+0"), sgpr("AddressA+0"), sgpr("OffsetA"), "add offset to buffer address")
-    kStr += inst("s_addc_u32", sgpr("AddressA+1"), sgpr("AddressA+1"), 0, "add offset to buffer address")
+    kStr += inst("s_mov_b32", sgpr(tmpOffset), sgpr("OffsetB"), "copy to temp b64 to prevent overflow with large offset")
```
nakajee, line 3129: Same here. Initialize tmpOffset+1 = 0

Tensile/KernelWriterAssembly.py
AlexBrownAMD: Thanks! Just posted an update

Tensile/KernelWriterAssembly.py
```diff
@@ -3108,22 +3108,30 @@ def allocateResources(self, kernel):
       kStr += ".if 0\n"
 
     # add offset to buffer
+    tmpOffset = self.sgprPool.checkOutAligned(2, 2, preventOverflow=0)
```
nakajee, line 3111: I am not sure the difference but I do not see checkOut or checkOutAligned is used for sgpr.
Normally, self.getTmpSgpr is used instead.
Any specific reason you used checkOutAligned here instead of self.getTmpSgpr?

AlexBrownAMD: checkOutAligned is in AsmRegisterPool.py:124. Since the two registers are used together for a 64-bit op, I think they need to be 64-bit aligned on gfx90a.

nakajee: It might not work if we call self.getTmpSgpr  twice in the same function.
If I remember correctly, calling getTmpSgpr did not work properly for some reason....
In graWorkGroup, we already have stmp = self.getTmpSgpr(4, 4).idx().
We might need to add 2 to this function instead of calling getTmpSgpr again.

nakajee: I am OK to use checkOutAligned if it is necessary.
I jut would like to know when we should use checkOut or Aligned for sgpr.

nakajee (approved):
Thanks for your update.
Looks good.

Tensile/KernelWriterAssembly.py
AlexBrownAMD: Chatted with Koji about this one offline. Led to us discovering some additional updates that should be done on sgpr register allocation. I've posted an update for the sgpr allocations related to this change and will work on the others later in a separate PR. Thanks!

yoichiyoshida (approved):

AlexBrownAMD:
Reverted the sgpr refactor for now since it changes the max number of simultaneous sgprs. Will complete this change using checkout/checkin mechanism for now to minimize impact to other kernel types and will revisit a refactor in a future PR.

nakajee:
> Reverted the sgpr refactor for now since it changes the max number of simultaneous sgprs. Will complete this change using checkout/checkin mechanism for now to minimize impact to other kernel types and will revisit a refactor in a future PR.

OK. We can refactor sgpr allocation code later.
For now, we need to make sure all tests pass with this change.

PR #1687, opened 2023-03-07
Hotfix: Fix offset calculation to prevent overflow if offset is really large

AlexBrownAMD:


TorreZuk (approved):
Same difference as develop I assume. LGTM

yoichiyoshida (approved):

nakajee (approved):
I am OK with the current change.
It might be better to change commit message if we include dual mac revert code.

By the way, the extended test fail is caused by my previous change.
(dtv_dgemm.yaml causes hang...)
At that time, gfx90a extended test was not enabled and I could not catch the fail when merging the code.

I have confirmed the failed scenario was not included in any of rocblas logic files and would not cause any issue with rocblas.

TorreZuk:
@AlexBrownAMD now I saw in the earlier fix generateBetaOnlyCall didn't seem to promote the offsets to 64bits?  Is that kernel setup not being updated, or was it missed.

AlexBrownAMD:
@TorreZuk Thanks for finding that. I've put together a change to update offsets for both PostGSU and BetaOnly kernels to 64-bit. WOrking on 64-bit strides now as well so I'll probably post both of those changes together.

TorreZuk:
Thanks @AlexBrownAMD.  I put @amcamd in the loop as reviewer as he will likely be consuming the end result of this.  My rocblas PR  I will leave alone but will be either added to or closed I expect.

AlexBrownAMD:
Updated hotfix PR with changes for 64-bit offset, post kernel updates, and max sgpr reduction (fix rocblas kernels)

AlexBrownAMD:
Updated hotfix PR with custom kernel updates, and sgpr allocation bug fix.

PR #1688, opened 2023-03-08
Update kernels to have 64-bit offset arguments

AlexBrownAMD:
Also fix error in client code that allocated double the required GPU memory when using offset.
Added a new test case with a 64-bit offset

### DIFF
--- a/Tensile/Components/Signature.py
+++ b/Tensile/Components/Signature.py
@@ -174,6 +174,11 @@ class SignatureDefault(Signature):
         kStr += self.addArgument(                               'A',     '8', offset, "global_buffer", srcValueType, "generic"); offset += 8
         kStr += self.addArgument(                               'B',     '8', offset, "global_buffer", srcValueType, "generic"); offset += 8
 
+        kStr += self.addArgument("OffsetD", '8', offset, "by_value", "u64"); offset += 8
+        kStr += self.addArgument("OffsetC", '8', offset, "by_value", "u64"); offset += 8
+        kStr += self.addArgument("OffsetA", '8', offset, "by_value", "u64"); offset += 8
+        kStr += self.addArgument("OffsetB", '8', offset, "by_value", "u64"); offset += 8
+
         useSize = max(4, cptByte)
         kStr += self.addArgument(                             "alpha", useSize, offset,      "by_value", cptValueType); offset += useSize
         if kernel["ProblemType"]["UseBeta"]:
@@ -237,11 +242,6 @@ class SignatureDefault(Signature):
         kStr += self.addArgument(                   "NumFullBlocks",     '4', offset,      "by_value",        "u32"); offset += 4
         kStr += self.addArgument(                   "WgmRemainder1",     '4', offset,      "by_value",        "u32"); offset += 4
         kStr += self.addArgument(        "MagicNumberWgmRemainder1",     '4', offset,      "by_value",        "u32"); offset += 4
-
-        kStr += self.addArgument("OffsetD", '4', offset, "by_value", "u32"); offset += 4
-        kStr += self.addArgument("OffsetC", '4', offset, "by_value", "u32"); offset += 4
-        kStr += self.addArgument("OffsetA", '4', offset, "by_value", "u32"); offset += 4
-        kStr += self.addArgument("OffsetB", '4', offset, "by_value", "u32"); offset += 4
 
         kStr += self.addArgument(                         "padding",     '4', offset,      "by_value",        "u32"); offset += 4
         kStr += "    .group_segment_fixed_size:   %u%s" % ( group_segment_size, writer.endLine ) #XXXXXX
--- a/Tensile/Source/client/source/DataInitialization.cpp
+++ b/Tensile/Source/client/source/DataInitialization.cpp
@@ -379,11 +379,6 @@ namespace Tensile
                 m_maxBatch = std::max(m_maxBatch, numOfBatch);
             }
 
-            m_aMaxElements += m_aBufferOffset;
-            m_bMaxElements += m_bBufferOffset;
-            m_cMaxElements += m_cBufferOffset;
-            m_dMaxElements += m_dBufferOffset;
-
             if(m_curBoundsCheck == BoundsCheckMode::NaN)
             {
                 m_aMaxElements += 1024;
--- a/Tensile/Source/lib/source/ContractionSolution.cpp
+++ b/Tensile/Source/lib/source/ContractionSolution.cpp
@@ -392,6 +392,11 @@ namespace Tensile
             rv.args.append<typename TypedInputs::BType const* const*>("batchB", inputs.batchB);
         }
 
+        rv.args.append<uint64_t>("offsetD", d.offset());
+        rv.args.append<uint64_t>("offsetC", c.offset());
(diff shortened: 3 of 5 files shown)


Tensile/Tests/pre_checkin/mfma/dgemm_64bit_offset.yaml
```diff
@@ -0,0 +1,64 @@
...
+  NumElementsToValidate: -1
+  BoundsCheck: True
+  KernelTime: True
+  # PrintSolutionRejectionReason: True
+  BufferOffsetB: 4294967296
+  PristineOnGPU: False
+
+BenchmarkProblems:
+  ########################################
+  # NT - standard
```
nakajee, line 14: Could you please make the comment here accurate?
This seems DGEMM NN.

Tensile/Tests/pre_checkin/mfma/dgemm_64bit_offset.yaml
```diff
@@ -0,0 +1,64 @@
...
+  ########################################
+  - # sgemm TN
```
nakajee, line 16: Same here.

Tensile/Source/client/source/DataInitialization.cpp
```diff
@@ -379,11 +379,6 @@ namespace Tensile
                 m_maxBatch = std::max(m_maxBatch, numOfBatch);
             }
 
-            m_aMaxElements += m_aBufferOffset;
```
nakajee, line 382: Why this part is removed?
If we remove this, should we remove m_aBufferOffset as well?
It looks like m_aBufferOffset used only here.

Tensile/Tests/pre_checkin/mfma/dgemm_64bit_offset.yaml
```diff
@@ -0,0 +1,64 @@
...
+        - KernelLanguage: ["Assembly"]
+        - EdgeType: ["ShiftPtr"]
+        - PrefetchLocalRead: [True]
+      ForkParameters:
+        - AssertMinApproxSize: [0]
+        - DepthU: [8]
+        - ExpandPointerSwap: [0]
+        - GroupLoadStore: [1]
+        - LocalReadVectorWidth: [1]
+        - MACInstruction: ["MAD"]
```
nakajee, line 37: Could you please use FMA here to make thing easier in the future?

nakajee: Or, just remove this line.

Tensile/Source/lib/source/ContractionSolution.cpp
```diff
@@ -392,6 +392,11 @@ namespace Tensile
             rv.args.append<typename TypedInputs::BType const* const*>("batchB", inputs.batchB);
         }
 
+        rv.args.append<uint64_t>("offsetD", d.offset());
+        rv.args.append<uint64_t>("offsetC", c.offset());
+        rv.args.append<uint64_t>("offsetA", a.offset());
+        rv.args.append<uint64_t>("offsetB", b.offset());
```
TorreZuk, line 398: I can't say I've seen the negative offset or stride use case yet, but on our side these are all signed 64bit.  Maybe you just want to use int64_t and do signed arithmetic in the assembly?

Tensile/Tests/pre_checkin/mfma/dgemm_64bit_offset.yaml
AlexBrownAMD: Fixed

Tensile/Tests/pre_checkin/mfma/dgemm_64bit_offset.yaml
AlexBrownAMD: Fixed

Tensile/Tests/pre_checkin/mfma/dgemm_64bit_offset.yaml
AlexBrownAMD: I used "MAD" setting here to match a specific test case from another project that saw an error. This setting shouldn't really affect the test, but just wanted to be sure the requested test was working.

Tensile/Source/client/source/DataInitialization.cpp
AlexBrownAMD: Offsets being added here created a bug where buffers would be allocated that were 2x the size of the offset. Offset is already handled in TensorDescriptor, and unnecessarily doubling the allocation made testing large offset problems impossible due to running out of memory. This change fixes the bug, but as you mention there may be additional clean-up work that could be done. I'll create a separate ticket for that.

Tensile/Source/lib/source/ContractionSolution.cpp
AlexBrownAMD: Thanks for checking that. I updated the arguments to int64_t, but since it is not a current use case, I think I'll defer updating the assembly math to a separate ticket.

Tensile/Components/Signature.py
```diff
@@ -174,6 +174,11 @@ def __call__(self, writer):
         kStr += self.addArgument(                               'A',     '8', offset, "global_buffer", srcValueType, "generic"); offset += 8
         kStr += self.addArgument(                               'B',     '8', offset, "global_buffer", srcValueType, "generic"); offset += 8
 
+        kStr += self.addArgument("OffsetD", '8', offset, "by_value", "u64"); offset += 8
```
nakajee, line 177: Do we need to change here to i64?

Tensile/Source/lib/source/ContractionSolution.cpp
TorreZuk: Yes that is fine we can defer as I pointed out we need to update rocblas side for this to match as we also stored in unsigned in the RocblasContractionProblem struct.   I commented on the commit below which is maybe invisible.  But if we defer this work you may want to keep the size_t for now?

AlexBrownAMD: Changed it back to uint64_t for now and created a ticket to tackle signed offsets after these bug fixes are done

AlexBrownAMD:
The only test failure on initial CI run was the new 64-bit offset test. The test passes locally: It failed because CI nodes limit the amount of host memory. I've replaced the dgemm 64-bit offset test with an sgemm 64-bit offset test. This will cut memory requirement in half while still testing a large offset index.

Tensile/Tests/pre_checkin/mfma/sgemm_64bit_offset.yaml
```diff
@@ -0,0 +1,64 @@
...
+  KernelTime: True
+  # PrintSolutionRejectionReason: True
+  BufferOffsetB: 4294967296
+  PristineOnGPU: False
+
+BenchmarkProblems:
+  ########################################
+  # NN - standard
+  ########################################
+  - # dgemm NN
```
nakajee, line 16: Please also change here to sgemm...

AlexBrownAMD: Fixed

nakajee (approved):
Thanks for your update.
I am OK with current change as long as no CI test fail due to this change.

PR #1691, opened 2023-03-14
Reduce max sgpr usage by improving temporary register re-use

AlexBrownAMD:
The recent change to 64-bit offsets reorganized the order in which some sgprs were allocated. In general, the change to 64-bit offsets did not increase the total number of required registers. But due to some quirks in the way the the register pool does allocations, the change in order meant register re-use could be affected.
A few kernels tuned into rocblas were affected, increasing the number of sgprs required over the maximum. This change updates the way some temp registers are allocated to improve register re-use. This change fixes the rocblas build.
As a future work on this, I'd like to improve the register pool code to fix a few strange cases I discovered while debugging this issue. For now, this change should unblock testing and rocblas build.

### DIFF
--- a/Tensile/KernelWriterAssembly.py
+++ b/Tensile/KernelWriterAssembly.py
@@ -4370,9 +4370,12 @@ class KernelWriterAssembly(KernelWriter):
   def computeLoadSrd(self, kernel, tP, tc, indices, bpe, isPap):
     kStr = ""
 
-    stmp = self.getTmpSgpr(2+2+1).idx()
-    tileStart = stmp+2
-    prePadSgpr = stmp+4
+    stmpRef = self.getTmpSgpr(2)
+    stmp = stmpRef.idx()
+    tileStartRef = self.getTmpSgpr(2)
+    tileStart = tileStartRef.idx()
+    prePadSgprRef = self.getTmpSgpr(1)
+    prePadSgpr = prePadSgprRef.idx()
     wroteTileStart = False
     #---
     # Compute tileStart #elements from the 2D array start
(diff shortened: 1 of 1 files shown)


nakajee (approved):
I am not sure how this can fix the issue.
As long as I checked I do not see any issue by splitting sgpr allocation in this way.
I am OK with this change if this can fix the rocblas test failure.
By the way, one extended test seems failed, but it is added by me and it has nothing to do this change.
I will check the extended test fail.

AlexBrownAMD:
This fixes the issue by reducing distinct allocations in the sgpr pool that are required. Many of our sgpr allocations that are larger than a single register are either 2 or 4 and need to be aligned, often creating gaps in the pool. Odd size allocations like 5 registers are too large to re-use the common 2 or 4 register blocks, causing it to expand the register pool. Additionally because it is an odd number, the next allocation may also waste another register to fix alignment.
I ran a rocblas build locally and it builds successfully with this change.

nakajee:
> This fixes the issue by reducing distinct allocations in the sgpr pool that are required. Many of our sgpr allocations that are larger than a single register are either 2 or 4 and need to be aligned, often creating gaps in the pool. Odd size allocations like 5 registers are too large to re-use the common 2 or 4 register blocks, causing it to expand the register pool. Additionally because it is an odd number, the next allocation may also waste another register to fix alignment. I ran a rocblas build locally and it builds successfully with this change.

Thanks for your explanation.
That is good to know.

By the way, I started a new PR (#1692) for extended test issue fix.
The failed case should be OK with this fix.

PR #1696, opened 2023-03-21
Update custom kernels with 64-bit offsets

AlexBrownAMD:


nakajee (approved):
Looks good if all tests pass

PR #1701, opened 2023-03-24
Move MinKForGSU from globalparameter to BenchmarkCommonParameter to support larger GSU

amgddm:
The parameter MinKForGSU is a global parameter whose default value is 256. This default value works fine for big matrix GEMM like HPL related GEMM. However, it doesn't work well for small matrix GEMM since it prevents GEMM from using larger GlobalSplitU. It checks the validation of a solution by the following statement in file Contractions.py.
           value =  globalparameter['MinKForGSU'] * state['GlobalSplitU']
The parameter MinKForGSU is put into BenchmarkCommonParameter. The condition above is changed to
            value =  state['MinKForGSU'] * state['GlobalSplitU']


Tensile/Common.py
```diff
@@ -250,7 +250,7 @@
 
 # limitation for training
 globalParameters["MaxWorkspaceSize"] = 32 * 1024 * 1024 # max workspace for training (32M)
-globalParameters["MinKForGSU"] = 256 # min K size to use GlobalSplitU algorithm (only for HPA now)
+#globalParameters["MinKForGSU"] = 256 # min K size to use GlobalSplitU algorithm (only for HPA now)
```
nakajee, line 253: We do not need to keep old code.

Tensile/Common.py
```diff
@@ -1292,6 +1292,7 @@ def getArchitectureName(gfxName):
     "MinVgprNumber":                list(range(0,256)),
 
     "MaxVgprNumber":                list(range(0,257)),
+    "MinKForGSU":                   range(1,1024)
```
nakajee, line 1295: Shouldn't this be a list?

nakajee: Please add some comment here to show the usage.
(at least similar comment as the previous global parameter)

nakajee (commented):
I suggest to add some reject conditions.
At least, I suggest to use only default value for GSU=1 case to avoid generating duplicated kernels.
I have not fully considered if we need further reject condition(s) for this parameter.
We might not need such flexible values for this.
Like only power of 2 (32,64,128,256,,,) might be easier to maintain.

Also, please add a test case for this parameter.
You do not need to create completely new test file.
Just add a scenario to enable this in some existing GSU test case.
I am also OK to add a new file if you want.

Tensile/Common.py
babakpst, line 253: Is that only for HPA?
Maybe we need to delete this line.

Tensile/Common.py
babakpst, line 1295: Please let us know if you checked this range in rocBLAS.

nakajee:
What do you mean "to meter" in the title? Typo?

babakpst (changes requested):
I'll put noCI label until the end of FC merge. 

Please let us know if you tested this in rocBLAS.

amgddm:
> What do you mean "to meter" in the title? Typo?

I copied and pasted commit message. Some letters were missing.

nakajee:
I do not know what is wrong with StaticAnalysis check.
Please run "tox -e lint" on your local machine and make sure no error generated.

Tensile/Common.py
amgddm: > Please let us know if you checked this range in rocBLAS.

I agreed with Koji that MinkForGSU should be some power of 2 data instead of a range.

Tensile/Common.py
amgddm: > Is that only for HPA? Maybe we need to delete this line.

Only for HPA.

nakajee: > > Is that only for HPA? Maybe we need to delete this line.
> 
> Only for HPA.

I do not think so.
I do not see any restriction to limit this parameter only for HPA case.
We should be able to use this for SGEMM, DGEMM, etc...

nakajee:
> > What do you mean "to meter" in the title? Typo?
> 
> I copied and pasted commit message. Some letters were missing.

OK. Please make sure the part is correct when doing "Squash and merge".
It might pick up the old version. Please manually change it if it is old.

amgddm:
> I do not know what is wrong with StaticAnalysis check. Please run "tox -e lint" on your local machine and make sure no error generated.

According to the log file, globalParameters is imported but is never used in Contraction.py. That is the reason. It failed if I ran 'tox -e lint'. If I delete the line 'from .Common import globalParameters'. It worked fine.

Tensile/Common.py
```diff
@@ -1292,6 +1291,7 @@ def getArchitectureName(gfxName):
     "MinVgprNumber":                list(range(0,256)),
 
     "MaxVgprNumber":                list(range(0,257)),
+    "MinKForGSU":                   [16,32,64,128,256] # min K size to use GlobalSplitU algorithm (only for HPA now)
```
nakajee, line 1294: Please move this comment to the above line like other parameters.
Also, we do not need "(only for HPA now)" part.
We should be able to use this for non HPA scenarios (SGEMM, DGEMM, etc...).

nakajee:
By the way, are you pushing commits to your fork?
I still do not see your second commit separately (still show only 1 commit).

Tensile/Common.py
babakpst: I am not sure if it is only for HPA. I guess we can use this for all GEMM. We can test this though.

Tensile/Common.py
amgddm: > Please move this comment to the above line like other parameters. Also, we do not need "(only for HPA now)" part. We should be able to use this for non HPA scenarios (SGEMM, DGEMM, etc...).

That comment is copied from original comment. I don't know why it is 'only for HPA now'.  It should work for non HPA scenarios as you commented.

nakajee: > > Please move this comment to the above line like other parameters. Also, we do not need "(only for HPA now)" part. We should be able to use this for non HPA scenarios (SGEMM, DGEMM, etc...).
> 
> That comment is copied from original comment. I don't know why it is 'only for HPA now'. It should work for non HPA scenarios as you commented.

I understand that you copied it from the original code.
However, if it is not true, we should fix it.

amgddm:
> By the way, are you pushing commits to your fork? I still do not see your second commit separately (still show only 1 commit).

Yes. I created a branch based on main repo and then push to my forked repo. I will follow others to create a branch based on forked repo instead of main repo.

Tensile/Common.py
amgddm: > I am not sure if it is only for HPA. I guess we can use this for all GEMM. We can test this though.

As I pointed out in my other posts: I agreed with both you and Koji. That comment was copied and pasted from old comment. From the code, nothing prevents it from being used for other non-HPA GEMM. I will fix it.

nakajee (changes requested):
Remaining items
- add test case
- add reject condition for GSU=1 (+ some more if needed)

nakajee:
> > By the way, are you pushing commits to your fork? I still do not see your second commit separately (still show only 1 commit).
> 
> Yes. I created a branch based on main repo and then push to my forked repo. I will follow others to create a branch based on forked repo instead of main repo.

But, I still do not see your second and third commits separately. It is still single commit including everything....

amgddm:
> > > By the way, are you pushing commits to your fork? I still do not see your second commit separately (still show only 1 commit).
> > 
> > 
> > Yes. I created a branch based on main repo and then push to my forked repo. I will follow others to create a branch based on forked repo instead of main repo.
> 
> But, I still do not see your second and third commits separately. It is still single commit including everything....

I checked out the branch from the forked repo and push to the forked repo. Did you see the another commit separately?

Tensile/SolutionStructs.py
```diff
@@ -2705,6 +2705,9 @@ def assignDerivedParameters(state):
     if state["WavefrontSize"] == 32 and state["KernelLanguage"] == "Source":
       reject(state, "WavefrontSize=32 not yet supported for source kernels.")
 
+    if state["MinKForGSU"] not in [16,32,64,128,256]:
```
nakajee, line 2708: This is not what I said.
We do not need out of range check.
It is already implemented for all parameters.

What I asked is to reject GSU=1 and MiKForGSU!=default value.
We already have similar check for GSU=1 and MultiBuffer.
Please put the new check after other checks for GSU=1.

Tensile/Tests/extended/global_split_u/hgemm_gsu_minkforgsu.yaml
```diff
@@ -0,0 +1,65 @@
...
+        - GlobalSplitUAlgorithm: ["MultipleBuffer"]
+        - GlobalSplitUWorkGroupMappingRoundRobin: [False]
+        - GlobalSplitUSummationAssignmentRoundRobin:  [True]
+        - EdgeType: ["ShiftPtr"]
+        - VectorWidth: [1]
+        - PrefetchGlobalRead: [false]
+        - StaggerU: [0]
+        - StaggerUStride: [0]
+        - StaggerUMapping: [0]
+        - MinKForGSU: [32,33,34,35]
```
nakajee, line 51: The tensile kernels with these values will be exactly same (except for label).
It is not effective to run the test with same kernels multiple times.
Instead, we should try different scenarios.
(1)enable larger GSU with this change (covered by this scenario)
(2) try smallest MinKForGSU (use GSU = K / MinKForGSU)
(3) reject case (K / GSU < MinKForGSU)
(4) reject case 2 with GSU=1 (using not default MinKForGSU + GSU=1)

Tensile/Tests/extended/global_split_u/hgemm_gsu_minkforgsu.yaml
```diff
@@ -0,0 +1,65 @@
...
+  ValidationPrintValids: False
+  ShortNames: False
+  MergeFiles: True
+  PlatformIdx: 0
+  DeviceIdx: 0
+  #  DataInitTypeAB: 0
+  PrintSolutionRejectionReason: True
+
+BenchmarkProblems:
+  - # sgemm NN
```
nakajee, line 21: This is not sgemm. Please change the comment here.

Tensile/Common.py
```diff
@@ -1292,6 +1291,8 @@ def getArchitectureName(gfxName):
     "MinVgprNumber":                list(range(0,256)),
 
     "MaxVgprNumber":                list(range(0,257)),
+    # min K size to use GlobalSplitU algorithm 
+    "MinKForGSU":                   list(range(1,257))
```
nakajee, line 1295: Why you changed this back to range?
Would you provide some explanation about your code change every time you submit new change?

amgddm: Nothing to prevent Tensile from choosing non power of 2 number to be MinKForGSU. If  you take a look at the added test. HGEMM performs better if MinKForGSU of value 35 is selected rather than 32.

nakajee: As I already said, MinKForGSU does not change any kernel code.
This is used just for reject condition.
You are comparing the result with the same kernel.

amgddm: > As I already said, MinKForGSU does not change any kernel code. This is used just for reject condition. You are comparing the result with the same kernel.

Even though MinKForGSU doesn't change kernel code. It, however, affect choosing GSU value. For example, GSU=16 is not a valid value if MinKForGSU=256, K=1024. We cannot limit MinKForGSU to power of 2 numbers, which may prevent us from choosing GSU=12 if it works better.

nakajee: > > As I already said, MinKForGSU does not change any kernel code. This is used just for reject condition. You are comparing the result with the same kernel.
> 
> Even though MinKForGSU doesn't change kernel code. It, however, affect choosing GSU value. For example, GSU=16 is not a valid value if MinKForGSU=256, K=1024. We cannot limit MinKForGSU to power of 2 numbers, which may prevent us from choosing GSU=12 if it works better.

We can enable any GSU values by using  power of 2 MinKForGSU  only.
We can set MinKForGSU=64 to use GSU=12 for K=1024.

nakajee: We would not like to have many duplicated kernels by using  slightly different MinKForGSU values.

amgddm: > We would not like to have many duplicated kernels by using slightly different MinKForGSU values.

I changed the definition of MinKForGSU to a list of [32,64,128,256].  It doesn't affect performance too much anyway.

nakajee:
Remaining items:
- add reject condition for GSU=1
find this piece of code in SolutionStructs.py

\# to eliminate identical/duplicate kernels when GSU=1
if state["GlobalSplitU"] == 1:
\# GlobalSplitUAlgorithm is MultipleBuffer
if state["GlobalSplitUAlgorithm"] == 'MultipleBuffer':
reject(state, " GlobalSplitU=1 and GlobalSplitUAlgorithm='MultipleBuffer'. Rejecting GlobalSplitUAlgorithm='SingleBuffer' to avoid duplicate kernels.")
\# GlobalSplitUAtomicAdd is True
if state["GlobalSplitUAtomicAdd"]:
reject(state, " GlobalSplitU=1 and GlobalSplitUAtomicAdd=True. Rejecting to avoid duplicate kernels.")

you need to add another condition like if MinKForGSU != default (=256), then reject it to avoid duplicated kernel.

- improve the coverage of Test cases with more GSU values
(1)enable larger GSU with this change (covered by this scenario)
(2) try smallest MinKForGSU (use GSU = K / MinKForGSU)
(3) reject case (K / GSU < MinKForGSU)
(4) reject case 2 with GSU=1 (using not default MinKForGSU + GSU=1)

- comment update in the test case
  \# sgemm NN
   -> this is not sgemm

amgddm:
> I suggest to add some reject conditions. At least, I suggest to use only default value for GSU=1 case to avoid generating duplicated kernels. I have not fully considered if we need further reject condition(s) for this parameter. We might not need such flexible values for this. Like only power of 2 (32,64,128,256,,,) might be easier to maintain.
> 
> Also, please add a test case for this parameter. You do not need to create completely new test file. Just add a scenario to enable this in some existing GSU test case. I am also OK to add a new file if you want.

The only purpose of MinKForGSU is to check if GSU is valid by comparing the value of MinKForGSU*GSU and K. 
I set 256 as the default value of MinKForGSU which is the same as previous one.  If MinKForGSU is not set in an existing yaml 
file, the default value (256) is used and existing GSU is still valid.  If both GSU=1 and MF are set in a yaml file, there is an existing reject condition to rule out this combination which has nothing to do with the value of MinKForGSU

amgddm:
> Remaining items:
> 
> * add reject condition for GSU=1
>   find this piece of code in SolutionStructs.py
> 
> # to eliminate identical/duplicate kernels when GSU=1 if state["GlobalSplitU"] == 1: # GlobalSplitUAlgorithm is MultipleBuffer if state["GlobalSplitUAlgorithm"] == 'MultipleBuffer': reject(state, " GlobalSplitU=1 and GlobalSplitUAlgorithm='MultipleBuffer'. Rejecting GlobalSplitUAlgorithm='SingleBuffer' to avoid duplicate kernels.") # GlobalSplitUAtomicAdd is True if state["GlobalSplitUAtomicAdd"]: reject(state, " GlobalSplitU=1 and GlobalSplitUAtomicAdd=True. Rejecting to avoid duplicate kernels.")
> 
> you need to add another condition like if MinKForGSU != default (=256), then reject it to avoid duplicated kernel.
> 
> * improve the coverage of Test cases with more GSU values
>   (1)enable larger GSU with this change (covered by this scenario)
>   (2) try smallest MinKForGSU (use GSU = K / MinKForGSU)
>   (3) reject case (K / GSU < MinKForGSU)
>   (4) reject case 2 with GSU=1 (using not default MinKForGSU + GSU=1)
> * comment update in the test case
>   # sgemm NN
>   -> this is not sgemm

I don't think we need a rejection condition for GSU=1. I explained that in another comment.

nakajee:
> > I suggest to add some reject conditions. At least, I suggest to use only default value for GSU=1 case to avoid generating duplicated kernels. I have not fully considered if we need further reject condition(s) for this parameter. We might not need such flexible values for this. Like only power of 2 (32,64,128,256,,,) might be easier to maintain.
> > Also, please add a test case for this parameter. You do not need to create completely new test file. Just add a scenario to enable this in some existing GSU test case. I am also OK to add a new file if you want.
> 
> The only purpose of MinKForGSU is to check if GSU is valid by comparing the value of MinKForGSU*GSU and K. I set 256 as the default value of MinKForGSU which is the same as previous one. If MinKForGSU is not set in an existing yaml file, the default value (256) is used and existing GSU is still valid. If both GSU=1 and MF are set in a yaml file, there is an existing reject condition to rule out this combination which has nothing to do with the value of MinKForGSU

I am not talking about valid or invalid.
This is just to eliminate duplicated kernels.

We do not need any kernels with GSU1 and different MinKForGSU.

To make it easier, you can set MinKForGSU=256 if GSU=1.
It will not reduce the amount of Tensile test combinations, but at least, Tensile will not generate duplicated kernels with GSU1 and different MinKForGSU.

nakajee:
hgemm_gsu_default_minkforgsu.yaml is not necessary.
We already have GSU test cases. It uses default value.

nakajee:
Why so many CI tests are failing?

nakajee:
I think we need to ensure DepthU>=MinKForGSU.
Lets say, if DepthU=32, MinKForGSU=16, K=64, GSU=4.
K/GSU=64/4=16. This satisfies MinKForGSU=16, but I do not expect this works.
Would you please try this kind of extreme cases?

nakajee:
My comment:
- please fix static analysis fail
- (same as previous comment). hgemm_gsu_default_minkforgsu.yaml is not necessary. We already have GSU test cases which use default value
- Would you add a scenario covering DepthU=32, MinKForGSU=16, some K,GSU which satisfies K/GSU=16?
  I would like to make sure if this works with DepthU<MinKForGSU

It is always difficult to understand what you changed.
Please always provide comment about your change. 
Plus, can't you really avoid using force merge? I cannot trace commit log if you keep using force merge.

nakajee:
Precheckin and extended tests also failed.
Please check why they failed.

nakajee:
> > * Would you add a scenario covering DepthU=32, MinKForGSU=16, some K,GSU which satisfies K/GSU=16?
> >   I would like to make sure if this works with DepthU<MinKForGSU
> 
> Your statements conflict. Do you mean DepthU=16, MinkForGSU=32?

Correction:
  Would you add a scenario covering DepthU=32, MinKForGSU=16, some K,GSU which satisfies K/GSU=16?
   I would like to make sure if this works with DepthU **>** MinKForGSU

nakajee:
> > * (same as previous comment). hgemm_gsu_default_minkforgsu.yaml is not necessary. We already have GSU test cases which use default value
> 
> Remove it.

Did you really remove it?
I still see the file in "File Changed".

amgddm:
> > > * (same as previous comment). hgemm_gsu_default_minkforgsu.yaml is not necessary. We already have GSU test cases which use default value
> > 
> > 
> > Remove it.
> 
> Did you really remove it? I still see the file in "File Changed".

I haven't checked in my code yet.  :)

nakajee:
I do not know why precheckin and extended tests fail in this way.

Expected equality of these values:
  oclProps.maxGridSize[1]
    Which is: 2147483647
  hipProps.maxGridSize[1]
    Which is: 65535
[0;31m[  FAILED  ] [mOclSolutionAdapterTest.HardwareTest (0 ms)

Any idea?

amgddm:
> I do not know why precheckin and extended tests fail in this way.
> 
> Expected equality of these values: oclProps.maxGridSize[1] Which is: 2147483647 hipProps.maxGridSize[1] Which is: 65535 �[0;31m[ FAILED ] �[mOclSolutionAdapterTest.HardwareTest (0 ms)
> 
> Any idea?

I have no idea. I could not find any clues from log. I launched rebuild yesterday and it is still running.

Tensile/Tests/extended/global_split_u/hgemm_gsu_minkforgsu.yaml
```diff
@@ -0,0 +1,65 @@
...
+  CMakeBuildType: Release
+  EnqueuesPerSync: 1
+  SyncsPerBenchmark: 1
+  LibraryPrintDebug: False
+  NumElementsToValidate: 128
+  ValidationMaxToPrint: 16
+  ValidationPrintValids: False
+  ShortNames: False
+  MergeFiles: True
+  PlatformIdx: 0
```
nakajee, line 15: Please remove unrecognized global parameters.
PlatformIdx and DeviceIdx are unrecognized.

Tensile/Common.py
nakajee: > > I am not sure if it is only for HPA. I guess we can use this for all GEMM. We can test this though.
> 
> As I pointed out in my other posts: I agreed with both you and Koji. That comment was copied and pasted from old comment. From the code, nothing prevents it from being used for other non-HPA GEMM. I will fix it.

I just realized that the old comment "only for HPA now" was because of the following condition in Contractions.py
  state['_GlobalAccumulation'] != None

We have already changed the definition of _GlobalAccumulation.
After the change, _GlobalAccumulation is None for HSS, BSS (means Dest=Compute).
It does not make sense we apply minKForGSU for HHS, but not for HSS.
Also, _GlobalAccumulation is not None if GSU algorithm is MultipleBuffer.
I do not see any good reason we apply minKForGSU for MultipleBuffer, but not for SingleBuffer.

Probably, we can apply minKForGSU for all data types.
We can simply check state['GlobalSplitU']>1 instead of checking _GlobalAccumulation.

You can add some test cases for SingleBuffer, non HPA.

nakajee:
Please merge the latest develop branch.
CI test should pass with the latest develop code.

nakajee:
One thing I need to say is that I have added MinKForGSU in Tensile\Tests\extended\direct_to_vgpr\dtv_hgemm.yaml.
Please update this file.

babakpst (approved):

nakajee (approved):
All tests passed.

PR #1704, opened 2023-03-30
Fix BufferStore=0 and Ldc != Ldd case

AlexBrownAMD:
When using global store, D offset was being calculated based on C stride. This change adds correct offset calculations for D in BufferStore=0 case.

### DIFF
--- a/Tensile/KernelWriterAssembly.py
+++ b/Tensile/KernelWriterAssembly.py
@@ -2595,12 +2595,13 @@ class KernelWriterAssembly(KernelWriter):
     # This is used in Buffer addressing modes.
     # Global addressing modes expect the GLOBAL_OFFSET to initialize a full 64-bit address
     for (tc, indices, justOffset32, tP) in [ \
+        ("D", list(range(0, kernel["ProblemType"]["NumIndicesC"])), kernel["BufferStore"], None), \
         ("C", list(range(0, kernel["ProblemType"]["NumIndicesC"])), kernel["BufferStore"], None), \
         ("A", kernel["ProblemType"]["IndexAssignmentsA"], kernel["BufferLoad"], self.tPA), \
         ("B", kernel["ProblemType"]["IndexAssignmentsB"], kernel["BufferLoad"], self.tPB) ]:
 
       # BufferStore does not use this macro so don't generate it:
-      if tc == "C" and kernel["BufferStore"]:
+      if kernel["BufferStore"] and (tc == "C" or tc == "D"):
         continue
 
       kStr += self.comment("Global Offset %s"%tc)
@@ -2616,7 +2617,7 @@ class KernelWriterAssembly(KernelWriter):
       calcDims = [] # dimensions which are participating in the address calc (ignores other summation)
       mirrorSumDims = []
       for i in range(0, numDim):
-        if tc == 'C':
+        if tc == 'C' or tc == 'D':
           useInitialStrides = kernel["ProblemType"]["UseInitialStridesCD"]
           idxChar = self.indexChars[i]
         else:
@@ -2625,8 +2626,8 @@ class KernelWriterAssembly(KernelWriter):
 
         # tile index or unroll vgpr or summation
         # other summation (other than unroll) are included in the GLOBAL_OFFSET macro but not used in address calc
-        if     tc in ('A','C') and indices[i] == kernel["ProblemType"]["Index0"] \
-            or tc in ('B','C') and indices[i] == kernel["ProblemType"]["Index1"] \
+        if     tc in ('A','C','D') and indices[i] == kernel["ProblemType"]["Index0"] \
+            or tc in ('B','C','D') and indices[i] == kernel["ProblemType"]["Index1"] \
             or indices[i] == kernel["ProblemType"]["IndexUnroll"]:
           kStr += " vgprOffset%s:req" % idxChars[i]
           calcDims.append(i)
@@ -10873,7 +10874,7 @@ class KernelWriterAssembly(KernelWriter):
       return kStr
 
     # TODO - mask should be part of AddrCalc state not passed as parm
-    def emitAddressSetupCode(self, kernel, ss, tmpVgpr, tmpS01, edge, beta, atomic, elementIdx, addrVgpr):
+    def emitAddressSetupCode(self, kernel, ss, tc, tmpVgpr, tmpS01, edge, beta, atomic, elementIdx, addrVgpr):
       """
       Generate code to set up the address vgpr
       Input:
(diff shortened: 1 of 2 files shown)


nakajee:
Have you tried LdcEqualsLdd=True?

Tensile/KernelWriterAssembly.py
```diff
@@ -12013,7 +12015,9 @@ def globalWriteBatch(self, kernel, ss, batchIdx, applyAlpha, beta, edge, atomic,
             # CEqualsD case, use SrdD instead of SrdC
             LoadCCodeStr = LoadCCodeStr.replace('SrdC', 'SrdD')
           LoadCCodeMod.addCode(LoadCCodeStr)
-
+          
+      if not kernel["BufferStore"] and not kernel["LdcEqualsLdd"]:
```
nakajee, line 12019: I think we need to use AssertCEqualsD instead of LdcEqualsLdd.
LdcEqualsLdd seems to be deprecated.

Tensile/KernelWriterAssembly.py
```diff
@@ -10963,11 +10965,11 @@ def emitAddressSetupCode(self, kernel, ss, tmpVgpr, tmpS01, edge, beta, atomic,
 
       # calculate global load offset
       # this has to be done after calculating new coord1
-      if not kernel["BufferStore"]:
+      if not kernel["BufferStore"] and (tc == 'D' or (tc == 'C' and beta)):
```
nakajee, line 10968: It looks like GLOBAL_OFFSET C or D code is not generated for ldcEqualLdd and not beta case.
We need to change this part like either change this part of condition or call this function with D.

Tensile/KernelWriterAssembly.py
AlexBrownAMD: Ok, we might need to mark as deprecated and reove

AlexBrownAMD: AssertCEqualsD has a different meaning, that's for when C and D are the same memory buffer. LdcEqualsLdd was just for equal stride values, but if it's not working and no longer needed it should probably just be removed.

nakajee: > Ok, we might need to mark as deprecated and reove

It is already written in the definition of LdcEqualsLdd.

nakajee: > AssertCEqualsD has a different meaning, that's for when C and D are the same memory buffer. LdcEqualsLdd was just for equal stride values, but if it's not working and no longer needed it should probably just be removed.

I understand they are different. CEquelsD implies ldc=ldd and we can do same optimizations.
However, that might not be needed.
We can focus on ldc!=ldd only without considering optimization for ldc=ldd at this moment.

Tensile/KernelWriterAssembly.py
AlexBrownAMD: Updated

Tensile/KernelWriterAssembly.py
```diff
@@ -10963,11 +10965,11 @@ def emitAddressSetupCode(self, kernel, ss, tmpVgpr, tmpS01, edge, beta, atomic,
...
         # global: in-bounds exec mask
         # global offset macro (requires 3 tmpVgpr)
         # final address = C + index*bytes
-        kStr += "GLOBAL_OFFSET_C %u" % addrVgpr
+        kStr += "GLOBAL_OFFSET_%s %u" % (tc, addrVgpr)
```
nakajee, line 10972: It looks like v_mov to tmpVgpr (line 10981,10982 after change) is redundant.
Originally, these v_mov is before setting exec flag and v_add is after exec flag.
With your change, GLOBAL_OFFSET_D is already after setting exec flag.
Then, v_mov to tmpVgpr and v_add (addr += tmpVgpr) are all within exec flag.

I guess the original intention for this part is
   (1) GLOBAL_OFFSET and v_mov to tmpVgpr for all thread
   (2) vadd within exec flag (do not execute vadd for out of range part)

However, it looks like we do not need such consideration.
We can simply add the result of GLOBAL_OFFSET with C/D addr.

nakajee:
Some extended tests are failed.
I quickly checked them.
I found 2 issues in your current code.
(1) bpe for GLOBAL_OFFSET_D macro
  we have a condition to use bpe external for C. We need to add tc=="D" case to use proper bpe for D
(2) Shift Pointer for MFMA code
  it looks like we should not generate this code for D. Generating this part of code twice seems not working.

nakajee:
Have you tried GSU>1 (+ MultiBuffer(?)) scenario?
It looks like WorkGroup2 calculation is still necessary in non-StridedBatch + GSU case.

Tensile/KernelWriterAssembly.py
```diff
@@ -11997,8 +11999,7 @@ def globalWriteBatch(self, kernel, ss, batchIdx, applyAlpha, beta, edge, atomic,
       vc0 = element[3]
 
       kStr += addrCalc.emitAddressSetupCode(kernel, ss, 'C', tmpVgpr, tmpS01, edge, beta, atomic, elementIdx, addrDVgpr)
-
-      if edge:
+      if edge and (kernel["BufferStore"] or beta):
```
nakajee, line 12002: What about edge + BufferStore + not beta case?
edgeProtectCode will not be called, right?

Tensile/KernelWriterAssembly.py
```diff
@@ -12013,6 +12020,11 @@ def globalWriteBatch(self, kernel, ss, batchIdx, applyAlpha, beta, edge, atomic,
             # CEqualsD case, use SrdD instead of SrdC
             LoadCCodeStr = LoadCCodeStr.replace('SrdC', 'SrdD')
           LoadCCodeMod.addCode(LoadCCodeStr)
+          
+      if not kernel["BufferStore"]:
+        kStr += addrCalc.emitAddressSetupCode(kernel, ss, 'D', tmpVgpr, tmpS01, edge, beta, atomic, elementIdx, addrDVgpr)
+        if edge:
```
nakajee, line 12026: We probably do not need to call edgeProtectCode twice for (not BufferStore) + edge + beta case.
It seems not making any difference.

AlexBrownAMD:
> Have you tried GSU>1 (+ MultiBuffer(?)) scenario? It looks like WorkGroup2 calculation is still necessary in non-StridedBatch + GSU case.

Fix GSU MB and added new test cases for all combinations including BufferLoad/Store + GSU MB/SB

nakajee:
> > Have you tried GSU>1 (+ MultiBuffer(?)) scenario? It looks like WorkGroup2 calculation is still necessary in non-StridedBatch + GSU case.
> 
> Fix GSU MB and added new test cases for all combinations including BufferLoad/Store + GSU MB/SB

Thanks for your update.
I did not clearly say, but according to other location of code, the condition seems to be "not kernel["_GlobalAccumulation"]" rather than kernel["_GlobalAccumulation"] == 'MultipleBuffer'.

We have another option _GlobalAccumulation=SingleBuffer.
This is used in case GSUAlgorithm=SingleBuffer + Dest type is different from Compute type (like HHS).
In either way, as long as extra buffer is allocated for GSU, C,D address seems to be provided as StridedBatch.

nakajee:
Please make sure your change works with GSUAlgorithm=SingleBuffer + HHS

AlexBrownAMD:
> Thanks for your update. I did not clearly say, but according to other location of code, the condition seems to be "not kernel["_GlobalAccumulation"]" rather than kernel["_GlobalAccumulation"] == 'MultipleBuffer'.

Thanks for the clarification. Updated the checks and added HHS test cases

> Please make sure your change works with GSUAlgorithm=SingleBuffer + HHS + non StridedBatch

Added a test case for this

PR #1706, opened 2023-04-06
Adding round near to zero mode for fp16alt implementation

yoichiyoshida:
Adding a second rounding mode for fp16alt implementation. Original implementation truncates on internal conversion, adding round nearest to zero option. Tests are added to rocBLAS companion PR

### DIFF
--- a/Tensile/KernelWriterAssembly.py
+++ b/Tensile/KernelWriterAssembly.py
@@ -1507,7 +1507,8 @@ class KernelWriterAssembly(KernelWriter):
     if kernel["ProblemType"]["Fp16AltImpl"]:
       self.G2Lpipe0 = vgprIdx
       self.G2Lpipe1 = self.G2Lpipe0 + 1
-      vgprIdx += 2
+      self.fp16AltTmp = self.G2Lpipe1 + 1
+      vgprIdx += 3
 
     self.startVgprAddressDbg = vgprIdx
     vgprIdx += numVgprAddressDbg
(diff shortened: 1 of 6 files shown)


Tensile/KernelWriterAssembly.py
```diff
@@ -8790,13 +8802,16 @@ def localWriteDo(self, kernel, tP, uDu=0):
...
-                   vgprsrc = vgpr("G2L%s+%u"%(tc, g2lIdx+iter))
-                   vgprsrc += " src0_sel:WORD_1"
-                   localWriteCode.addInst("v_cvt_f32_f16", vgpr("G2Lpipe0"), vgpr("G2L%s+%u"%(tc, g2lIdx+iter)),"")
-                   localWriteCode.addInst("v_cvt_f32_f16", vgpr("G2Lpipe1"), vgprsrc,"")
-                   localWriteCode.addInst("v_pack_b32_f16", vgpr("G2L%s+%u"%(tc, g2lIdx+iter)), vgpr("G2Lpipe0"),vgpr("G2Lpipe1"), "op_sel:[1,1,0]","")
+                  vgprsrc = vgpr("G2L%s+%u"%(tc, g2lIdx+iter))
+                  localWriteCode.addInst("v_cvt_f32_f16", vgpr("G2Lpipe0"), vgprsrc,"")
+                  localWriteCode.addInst("v_cvt_f32_f16", vgpr("G2Lpipe1"), vgprsrc, "src0_sel:WORD_1", "")
+
+                  localWriteCode.addInst("v_add_u32", vgpr("G2Lpipe0"), sgpr("Fp16AltRound"), vgpr("G2Lpipe0"), "")
```
nakajee, line 8811: I am not sure how v_cvt_f32_f16 works if the value is NaN.
Is this safer if we check the value is NaN or not and add rounding only when it is not NaN?

yoichiyoshida: I was debating putting that in, as that would add even more extra instructions on the existing functionality which might cause a performance regression. I'll add it though to be safe.

nakajee:
Thank you for your update.
Do we have any test case to verify this?

babakpst (approved):
Looks good.

yoichiyoshida:
Rejected due to performance regressions on existing fp16 alt truncate kernels. Issuing new PR based on separate kernels for RNZ mode.

PR #1713, opened 2023-04-26
Cleanup some unused and deprecated code

AlexBrownAMD:


Tensile/Common.py
```diff
@@ -205,7 +203,6 @@
 
 globalParameters["CleanupBuildFiles"] = False                     # cleanup build files (e.g. kernel assembly) once no longer needed
 globalParameters["GenerateManifestAndExit"] = False               # Output manifest file with list of expected library objects and exit
-globalParameters["NewClient"] = 2                                 # Old client deprecated: NewClient must be set to 2.
```
nakajee, line 208: I remember that we intentionally kept NewClient in your previous commit.

Comment from zaliu:
@AlexBrownAMD Do we really want to remove the global parameter NewClient? It's used everywhere, not just in the Tensile repo. Can we retain NewClient either as a no-op or issue a warning that we only support the new client?

Is that safe now to remove this global parameter?

nakajee: I would like to make sure if there is any change about this or it was already safe from the previous change.

AlexBrownAMD: It's not used anywhere in code anymore. The only time it's used in code is to check if it's set to "2" and if not, print a warning that it is deprecated. Some legacy yamls in the configs folder still include the option, but those aren't really used anymore either, and having the option there won't break anything. It's been marked as deprecated for over a year, so I think we can safely remove the last references now.

nakajee: OK. I grepped "NewClient" under Tensile.
I see some hit.
  Tensile\Tests\pre_checkin\denorm\mfma\bfloat16_1k_denorm.yaml

A lot under Tensile\Configs\miopen\archives (assuming this will not be used).

nakajee: Not sure about out of Tensile.
But you are sure NewClient is not used, right?

AlexBrownAMD: Yep, all the old client code was removed long ago. The NewClient option doesn't do anything, it's just been kept for a while to print the deprecation warning.

nakajee: OK. Can you remove NewClient from Tensile\Tests\pre_checkin\denorm\mfma\bfloat16_1k_denorm.yaml?

nakajee (approved):
Thanks for your update.
I do not have further comment.
Please make sure all tests pass with this change.
I saw an error with precheckin+gfx908 (but might not be related to your change)

PR #1718, opened 2023-05-06
fix bug in fp16 alt impl rnz for some scheduler modes

yoichiyoshida:
In some scheduling modes, fp16 alt impl rnz cannot use temporary sgprs due to assumptions in the scheduler. Switch to using globally defined sgprs until scheduling mode fixed.


AlexBrownAMD (approved):

nakajee (approved):
I am OK with the code change.
It might be better if we can add some test case for this issue.

PR #1732, opened 2023-06-19
Fix boundary check for mirror dims

AlexBrownAMD:
Fix an issue that caused mirror dims kernels to segfault. Mirrored dimensions were creating vgpr offsets to start at the end, then modifying the buffer srd to step backwards. Some lanes in the tail loop would run off the beginning of the buffer and cause segfault in rocm 5.3 and newer.

This change modifies the way mirrored dimensions are stepped: rather than modifying the srd, the vgpr offset is now updated. This way, the tail loop can easily clamp any lanes that run out of bounds.

This PR also re-enables the mirror_dims test cases that were marked as skip when the bug appeared. Ran all pre_checkin and extended tests to verify other cases are unaffected.

### DIFF
--- a/Tensile/KernelWriterAssembly.py
+++ b/Tensile/KernelWriterAssembly.py
@@ -7291,9 +7291,16 @@ class KernelWriterAssembly(KernelWriter):
         if (tc in ('A', 'B') and kernel["ProblemType"]["IndicesSummation"][self.unrollIdx] in kernel["ProblemType"]["MirrorDims%s"%tc]):
           if kernel["_UseSgprForGRO"]:
             offsetVgpr = "GlobalReadOffset%s+0"%(tc)
+            imod.addInst("_v_add_i32", vgpr(offsetVgpr), vgpr(offsetVgpr), sgpr("GlobalReadIncs%s+%u"%(tc,loopIdx)), "update offset for mirror dims")
           else:
-            offsetVgpr = "GlobalReadOffset%s+%u"%(tc, graIdx)
-          imod.addInst("_v_add_i32", vgpr(offsetVgpr), vgpr(offsetVgpr), sgpr("GlobalReadIncs%s+%u"%(tc,loopIdx)), "update offset for mirror dims")
+            for perp in range(0, tP["nrp"]):
+              for sPerp in range(0, tP["nrpv"]):
+                for para in range(0, tP["nrc"]):
+                  for sPara in range(0, tP["nrcv"]//tP["nrcvpi"]):
+                    i = sPara + (tP["nrcv"] // tP["nrcvpi"]) * (para + tP["nrc"] * (sPerp + tP["nrpv"] * perp))
+                    graIdx = i * self.rpgo if kernel["BufferLoad"] else i * self.rpga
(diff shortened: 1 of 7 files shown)


Tensile/KernelWriterAssembly.py
```diff
@@ -7684,15 +7696,29 @@ def globalReadGuardK(self, kernel, tP, vregSetIdx):
...
+                  # check for pre-padding
+                  prePad = 0
+                  if self.groOffsetInMacroTile:
+                    prePad = self.srdShiftLeft[tc]
+                    if tc == "A":
+                      prePad *= self.tPA["bpe"]
+                    elif tc == "B":
+                      prePad *= self.tPB["bpe"]
+
+                  codeMod.addInst("v_max_i32", vgpr(offsetVgpr), prePad, vgpr(offsetVgpr), "clamp to beginning of buffer")
```
nakajee, line 7720: I think it is better to stick to unsigned calculation.
It is technically possible to have >=2^31 values in vgpr(offsetVgpr) or soffset.
(if lda is very large)
We can do like this
   tmp = offsetVgpr - 4 # remove prepad
   tmp = max_u32(tmp, soffset)
   tmp = tmp - soffset
   tmp = tmp + 4

nakajee:
Please fix static analysis fail.

Tensile/KernelWriterAssembly.py
```diff
@@ -7684,15 +7696,29 @@ def globalReadGuardK(self, kernel, tP, vregSetIdx):
...
+                  if soffset != "0":
+                    codeMod.addInst("_v_sub_u32", vgpr(offsetVgpr), vgpr(offsetVgpr), soffset, "mirror unroll: GRO=GRO-SGRO, soffset=0")
+                    # soffset_prev = soffset
+                    soffset = "0"
+
+                  # check for pre-padding
+                  prePad = 0
+                  if self.groOffsetInMacroTile:
+                    prePad = self.srdShiftLeft[tc]
+                    if tc == "A":
```
nakajee, line 7715: We can simply use self.bpeAB here. No need to put if/elif here.

AlexBrownAMD: Updated

Tensile/KernelWriterAssembly.py
AlexBrownAMD: I don't think mirror dims feature ever worked for offset > 2^31 because sgpr offset was always signed i32. Looking through the code, there's a lot of other code that would also have to change to get it to work for u32. I might file a feature ticket to do that work separately from this change.

nakajee: > I don't think mirror dims feature ever worked for offset > 2^31 because sgpr offset was always signed i32. Looking through the code, there's a lot of other code that would also have to change to get it to work for u32. I might file a feature ticket to do that work separately from this change.

How about vgprGlobalReadOffset?
Even if sgpr offset is within 2^31, v_max_i32 can be wrong if vgprGlobalReadOffset is already >2^31.

nakajee (approved):
As we discussed, I am OK with going with the current version.
We can update the code for unsigned calculation support later.

PR #1733, opened 2023-06-28
Skip experimental libraries when building without lazy loading

yenong-amd:
This PR fixes "Argument list too long" error in TensileCreateLibrary by excluding experimental logic files when building without lazy library loading.


Tensile/TensileCreateLibrary.py
```diff
@@ -1291,6 +1291,10 @@ def splitExtraParameters(par):
   for logicFile in logicFiles:
     print1("#   %s" % logicFile)
 
+  # Skip experimental libraries (if exists) when building without lazy loading
+  if not globalParameters["LazyLibraryLoading"]:
+    logicFiles = [f for f in logicFiles if not ("experimental/dtree/" in f or "experimental/grid/" in f)]
+
```
nielenventer, line 1297: could we just exclude all `experimental/` instead of the two specific ones? I think having experimental as a general reserved name is ok, but I'm less a fan of these special cases hardcoded for dtree and grid.

nielenventer: actually maybe I'll be more nitpicky and request that the experimental folder name be a Tensile config parameter, that defaults to `experimental/` so it isn't hidden way down here.

Tensile/TensileCreateLibrary.py
```diff
@@ -1291,6 +1291,10 @@ def splitExtraParameters(par):
   for logicFile in logicFiles:
     print1("#   %s" % logicFile)
 
+  # Skip experimental libraries (if exists) when building without lazy loading
+  if not globalParameters["LazyLibraryLoading"]:
```
nakajee, line 1295: Shouldn't this be above the print1 at line 1290?
Otherwise, the names are printed but not included.

yenong-amd: Yes. That makes sense.

Tensile/TensileCreateLibrary.py
yenong-amd: Added the ExperimentalLogicDir to globalParameters.

nielenventer (approved):

babakpst:
@yenong-amd Can you please review the failed tests? maybe we need to rerun the tests, because I could not find the cause of the error.
It would also be great if you don't use push-force, because it clears the history, and reviewers cannot track all the changes. Thanks.

babakpst (changes requested):
On hold until after FC 6.

babakpst (approved):
Looks good. I have not tested this feature.

PR #1678, opened 2023-02-27
Improve error message when assembler is not found

cgmb:
This replaces the existing error message of:

    TypeError: expected str, bytes or os.PathLink object, not NoneType

with

    ValueError: No assembler available; set TENSILE_ROCM_ASSEMBLER_PATH to point to ROCm Clang.

PR #1739, opened 2023-07-11
Improve missing clang-offload-bundler message

cgmb:
This replaces the existing error message of:

    TypeError: expected str, bytes or os.PathLink object, not NoneType

with

    ValueError: No bundler available; set TENSILE_ROCM_OFFLOAD_BUNDLER_PATH to point to clang-offload-bundler.


cgmb:
This is basically the same as #1678 but for the offload bundler.

AlexBrownAMD (approved):

PR #1744, opened 2023-07-14
DirectToLds support for larger data types with 32bit global load

nakajee:
* enabled DirectToLds support for larger data types with 32bit global load
* removed old code for DTL + (bpe * GlobalReadVectorWidth > 4)
* added more test cases for DirectToLds


Tensile/SolutionStructs.py
```diff
@@ -2951,6 +2953,8 @@ def assignDerivedParameters(state):
     bufferLoad = state["BufferLoad"] and state["KernelLanguage"] == "Assembly"
     if not bufferLoad:
       state["DirectToLds"] = False
+      state["DirectToLdsA"] = False
+      state["DirectToLdsB"] = False
```
AlexBrownAMD, line 2957: Should this combo be a reject to avoid duplicate kernels?

Tensile/SolutionStructs.py
```diff
@@ -4111,7 +4120,8 @@ def assignDerivedParameters(state):
     # check LocalDotLayout
     ldl = state["LocalDotLayout"]
     if ldl> 1:
-      state["DirectToLds"] = False
+      state["DirectToLdsA"] = False
+      state["DirectToLdsB"] = False
```
AlexBrownAMD, line 4124: Same here?

AlexBrownAMD (commented):
Looks good. Have you tried any rocblas tests with this (it might change some default behaviour)?

Tensile/SolutionStructs.py
nakajee: I quickly tried DirectToLds=[0,1] + BufferLoad=[0,1].
Tensile did not generate duplicated kernels.
BL=0+DTL=1 case, DTL is changed to 0 and this kernel is not generated probably because all kernel parameters are exactly same as BL=0 + DTL=0.

Tensile/SolutionStructs.py
nakajee: I quickly tried LDL>1 + DTL=1, but I could not enable the combination....
It should be same as BL=0 and DTL=1.
Tensile will not generate duplicated kernels if all kernel parameters are the same (DTL1 is converted to DTL0 by this line).

nakajee:
> Looks good. Have you tried any rocblas tests with this (it might change some default behaviour)?

This does not change any existing kernels since I keep DirectToLds unchanged.
I tried TensileCreateLibrary with DGEMM+DTL (which has GlobalLoadVectorWidthB=0.5).
Tensile library was successfully generated with this change.

babakpst (approved):

AlexBrownAMD (approved):

PR #1755, opened 2023-08-16
Changes to Enable ROCm SMI for gfx940

aferoz21:
- Enable ROCm SMI for gfx940. Validated the changes in 21b node.
- Retrieving the mem clock information has a bug for gfx940. so added a safety check in HardwareMonitor.cpp to proceed further. Submitted a ticket to ROCm SMI team as well.

### DIFF
--- a/Tensile/Source/client/source/HardwareMonitor.cpp
+++ b/Tensile/Source/client/source/HardwareMonitor.cpp
@@ -365,16 +365,13 @@ namespace Tensile
                 rsmi_frequencies_t freq;
 
                 auto status = rsmi_dev_gpu_clk_freq_get(m_smiDeviceIndex, m_clockMetrics[i], &freq);
-                if(status != RSMI_STATUS_SUCCESS)
+                if(status != RSMI_STATUS_SUCCESS || freq.current > RSMI_MAX_NUM_FREQUENCIES)
                 {
                     m_clockValues[i] = std::numeric_limits<uint64_t>::max();
                 }
                 else
                 {
-                    if(freq.current < RSMI_MAX_NUM_FREQUENCIES)
-                    {
-                        m_clockValues[i] += freq.frequency[freq.current];
-                    }
+                    m_clockValues[i] += freq.frequency[freq.current];
                 }
             }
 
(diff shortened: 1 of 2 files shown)


Tensile/Source/client/source/HardwareMonitor.cpp
```diff
@@ -371,7 +371,10 @@ namespace Tensile
                 }
                 else
                 {
-                    m_clockValues[i] += freq.frequency[freq.current];
+                    if(freq.current < RSMI_MAX_NUM_FREQUENCIES)
```
nakajee, line 374: Where can I find the definition of RSMI_MAX_NUM_FREQUENCIES?
I am not sure why we need this logic.

aferoz21: it is defined in rocm_smi.h. RSMI_MAX_NUM_FREQUENCIES is 32. 

Excerpt: 
typedef struct {
    /**
     * The number of supported frequencies
     */
    uint32_t num_supported;

    /**
     * The current frequency index
     */
    uint32_t current;

    /**
     * List of frequencies.
     * Only the first num_supported frequencies are valid.
     */
    uint64_t frequency[RSMI_MAX_NUM_FREQUENCIES];
} rsmi_frequencies_t;

nakajee: OK. Thanks for clarification.
I think it is still inaccurate even if you skip adding frequency value here.
This is because m_dataPoints is incremented regardless of this skip.

m_dataPoints is also used for other data.
To make this accurate, either make this case invalid or count the number of this specific case and subtract the number from m_dataPoints when calculating average freq.

aferoz21: > OK. Thanks for clarification. I think it is still inaccurate even if you skip adding frequency value here. This is because m_dataPoints is incremented regardless of this skip.
> 
> m_dataPoints is also used for other data. To make this accurate, either make this case invalid or count the number of this 
>specific case and subtract the number from m_dataPoints when calculating average freq.

- I made this case as invalid to make it more accurate. Earlier it was giving 0 (initialized) for the mem clock as opposed to segmentation fault. Combination of valid and invalid values from the API is not reproducible. So I think incorrect average case will be very hard to reproduce.

babakpst (approved):

nakajee (approved):
Thanks for your update.
Looks good.

PR #1756, opened 2023-08-16
Develop

nielenventer:
Fixes a bug in the override mechanism: when using lazy loading, overrides were ignored (they didn't load when required). Also fixes off-by-one error, Also adds some better debugging logs.

### DIFF
--- a/Tensile/Source/lib/include/Tensile/CachingLibrary.hpp
+++ b/Tensile/Source/lib/include/Tensile/CachingLibrary.hpp
@@ -242,18 +242,14 @@ namespace Tensile
 
                 if(override_debug && solution)
                 {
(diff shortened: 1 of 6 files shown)


nielenventer:
The `getSolutionByIndex` function doesn't load solutions if they weren't already loaded (as indicated by a comment in the function itself).

I just followed the same pattern used here:
https://github.com/ROCmSoftwarePlatform/rocBLAS/blob/develop/library/src/tensile_host.cpp#L975

To load, if the solution isn't found.

nielenventer:
@aferoz21 this was the issue Juan was seeing when trying to use this.

babakpst (approved):

PR #1765, opened 2023-08-24
Enable batch

wbgilmartin:
This add the extra code to enable the decision tree solution selection to handle batch gemms. 

1) In tensile this is mostly coding to access the batch property from withing the dtree logic.
2) I did verify that this does work through by stepping through with a debugger.
3) I produced a model to verify that we get the type of results that are expected. 
4) I have not done a side by side comparison of the new feature compared to the current solution selection through rocBLAS.


nielenventer:
Change contents look good to me, just need to run clang-format so it passes static analysis

babakpst (approved):

yoichiyoshida (approved):

PR #1789, opened 2023-09-27
ROCm 6.0 merge master into develop

babakpst:
Merging master to develop for FC 6.0 stage 2


TorreZuk (commented):
@amcamd should review.  We aren't updating our versions until FC3 so if you bump your numbers now we need to match in our develop for Tensile at least.  Can still defer our version bump until FC3.

amcamd:
In agreement with Torre, for rocBLAS we are bumping version numbers at FC3. For FC2 we are only merging staging into release-staging/rocm-rel-6.0. If you bump version here we will need to bump the version of Tensile here: https://github.com/ROCmSoftwarePlatform/rocBLAS-internal/blob/develop/CMakeLists.txt#L156

babakpst:
@TorreZuk  @amcamd Thanks for your comments. As we discussed offline, this PR is for FC3. I should bump the version now.

nakajee (approved):

PR #1791, opened 2023-09-29
Clear hipErrorNotFound error code since it is an expected part of the search

AlexBrownAMD:
We expect to encounter hipErrorNotFound occasionally while searching for a kernel, and this error case is already handled. We need to call hipGetLastError to reset the stored error code so other apps don't assume an error was generated on one of their hip calls.

### DIFF
--- a/Tensile/Source/lib/source/hip/HipSolutionAdapter.cpp
+++ b/Tensile/Source/lib/source/hip/HipSolutionAdapter.cpp
@@ -230,6 +230,10 @@ namespace Tensile
                 {
                     return err;
                 }
+                else
+                {
+                    (void)hipGetLastError(); // clear hipErrorNotFound
+                }
             }
 
             return err;


nakajee:
Is the error message still printed with this change?

AlexBrownAMD:
We still return hipErrorNotFound if a kernel isn't found in any module, and the client prints an error message.

Also, ran all rocblas gemm tests with this change on gfx90a and all passed.

nakajee (approved):
Thanks for your comment.
That is great!

cgmb:
> We expect to encounter hipErrorNotFound occasionally while searching for a kernel

Why is that? These failures clutter the logs when running with `AMD_LOG_LEVEL=1` set. When I was debugging a problem in rocSOLVER, I saw a series of log messages stating "Cannot find the function: Cijk_Alik_Bljk_SB_..." and my first thought was that there may be a bug in rocBLAS TRSM.

If there's a reasonable way to avoid expected failures, that would be preferable.

AlexBrownAMD:
> Why is that? 

Kernels are divided among a number of modules to improve build/load times. When a kernel is requested, we currently need to search through the available modules until we find the kernel. Each check results in a hipErrorNotFound.

> If there's a reasonable way to avoid expected failures, that would be preferable.

Kernels are grouped by data+transpose type, so if the module name is saved we may be able to first check part of the kernel name for a matching module. It may not completely remove the messages, for cases where we use a fallback/generic source kernel, but might significantly reduce the number of errors generated.

I'll file a feature ticket for someone to look into it when time permits

PR #1797, opened 2023-10-11
Clear hipErrorNotFound error code since it is an expected part of the search (#1791)

AlexBrownAMD:


TorreZuk (approved):
LGTM passed earlier mathCI tests so already used in rocblas develop

babakpst (approved):

PR #1801, opened 2023-10-22
Disable HW monitor for 9fx1101 and gfx1102 products

aferoz21:
ROCm SMI HW monitor has been disabled for 9fx1101 and gfx1102 products.

PR #1802, opened 2023-10-22
PCIID Changes from 32bit to 64bit for ROCm SMI HW monitor

aferoz21:
resolves #SWDEV-419897
PCIID has been changed from 32bit to 64 bit from ROCm version 4.0 onwards. This PCI ID used to get the correct device index.
if the ROCM_VISIBLE_DEVICES=2, then “hip device 0” will match “ROCm SMI device 2”, and the way to discover that is with the PCI ID.

### DIFF
--- a/Tensile/Common.py
+++ b/Tensile/Common.py
@@ -2186,8 +2186,9 @@ def assignGlobalParameters( config ):
     if os.name == "nt":
       globalParameters["CurrentISA"] = (9,0,6)
       printWarning("Failed to detect ISA so forcing (gfx906) on windows")
-  if globalParameters["CurrentISA"] == (9,4,2):
-    printWarning("HardwareMonitor currently disabled for gfx942")
+  if globalParameters["CurrentISA"] == (9,4,2) or globalParameters["CurrentISA"] == (11,0,0) or \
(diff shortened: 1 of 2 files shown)


Tensile/Common.py
```diff
@@ -2207,8 +2207,9 @@ def assignGlobalParameters( config ):
     if os.name == "nt":
       globalParameters["CurrentISA"] = (9,0,6)
       printWarning("Failed to detect ISA so forcing (gfx906) on windows")
-  if globalParameters["CurrentISA"] == (9,4,2):
-    printWarning("HardwareMonitor currently disabled for gfx942")
+  if globalParameters["CurrentISA"] == (9,4,2) or globalParameters["CurrentISA"] == (11,0,0) or \
```
babakpst, line 2210: Do we need this change here? How is it different from #1801?

aferoz21: This is same change as 1801. if this PR CI goes fine, I will discard the 1801 PR. This PR disables ROCM SMI for gfx1100 and also take care of new PCI format changes. PR 1801 is just disable ROCm SMI for gfx1100.

nakajee (approved):
I think code change itself looks OK.
Please make sure all CI tests pass with this change.

PR #1805, opened 2023-10-24
Add code to allow testing stream-k grid multipliers.

AlexBrownAMD:
* Add code to allow testing stream-k grid multipliers.
* Fix race condition that appeared with large grids and small sizes.
* Add debug output to check occupancy.

### DIFF
--- a/Tensile/KernelWriterAssembly.py
+++ b/Tensile/KernelWriterAssembly.py
@@ -12719,12 +12719,12 @@ class KernelWriterAssembly(KernelWriter):
       kStr += PreLoopVmcntCaseStr
 
     # Set flag
-    kStr += inst("s_waitcnt", "lgkmcnt(0)", "wait for data store")
+    kStr += inst("s_waitcnt", "vmcnt(0)", "wait for data store")
     kStr += inst("s_barrier", "store all data before setting flag")
     kStr += inst("s_lshl_b32", sgpr(tmpSgpr), sgpr("StreamKIdx"), log2(4), "flag offset based on CTA index")
     kStr += inst("s_mov_b32", sgpr(tmpSgpr+2), 1, "flag data")
     kStr += inst("s_store_dword", sgpr(tmpSgpr+2), sgpr("AddressFlags", 2), sgpr(tmpSgpr), "glc", "set flag")
-    kStr += inst("s_waitcnt", "lgkmcnt(0)", "wait for flag load") # TODO just for testing
+    kStr += inst("s_waitcnt", "lgkmcnt(0)", "wait for flag") # TODO just for testing
     
     # TODO - if this is the last tile, don't need to jump to next instruction
     # NOTE: in SR kernel, we need long branch since PRNG explodes the line of codes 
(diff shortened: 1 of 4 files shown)


Tensile/Source/lib/source/ContractionSolution.cpp
```diff
@@ -35,6 +35,8 @@
 #include <cstdlib>
 #include <random>
 
+const int TENSILE_STREAMK_GRID 1
```
nakajee, line 38: don't we need ; here?

Tensile/KernelWriterAssembly.py
```diff
@@ -12719,6 +12719,8 @@ def partialsWriteProcedure(self, kernel, vectorWidths, elements, alpha, beta, ed
       kStr += PreLoopVmcntCaseStr
 
     # Set flag
+    kStr += inst("s_waitcnt", "lgkmcnt(0)", "wait for data store")
```
nakajee, line 12722: Shouldn't this be vmcnt?

Tensile/Source/lib/source/ContractionSolution.cpp
AlexBrownAMD: Yea, I had this as a define while testing, started to change it before posting but must have messed up the saved version. I've changed it back to define.

Tensile/KernelWriterAssembly.py
AlexBrownAMD: Yes it should, thanks! Posted a fix

Tensile/Source/lib/source/hip/HipSolutionAdapter.cpp
```diff
@@ -346,6 +346,16 @@ namespace Tensile
                                        &argsSize,
                                        HIP_LAUNCH_PARAM_END};
 
+            if(m_debug)
+            {
+                int numBlocks = 0;
+                int blockSize
+                    = kernel.workGroupSize.x * kernel.workGroupSize.y * kernel.workGroupSize.z;
+                HIP_CHECK_RETURN(hipModuleOccupancyMaxActiveBlocksPerMultiprocessor(
```
nakajee, line 354: I would like to understand this part a bit more.
- how the output look like? Could you share some example?
- Why 4th parameter(dynSharedMemPerBlk ) is 0? I do not quite understand the interface of this function.

AlexBrownAMD: This prints the occupancy of the kernel, I was using it to check occupancy of stream-k vs non-stream-k kernel variants. Output looks like:

Occupancy = 2

dynSharedMemPerBlk is amount of dynamic LDS we plan to allocate. We don't use any dynamic LDS, we know how much shared memory is required when we generate the kernel, so this is set to 0. The function can check .amdhsa_group_segment_fixed_size or .group_segment_fixed_size in kernel header for how much fixed shared memory is required to do the occupancy calculation.

nakajee: Thanks for your explanation. I got it.

nakajee:
Why integration test failed?

nakajee (approved):
All tests passed. Looks good.

PR #1806, opened 2023-10-25
LWPTENSILE-503 adding a new TENSILE_DB debugging value to dump the common kernel parameters

smalekta:
Adding a new debug value (0x80000) to TENSILE_DB which dumps the kernel name and number of common parameters (per Email request)

### DIFF
--- a/Tensile/Source/lib/source/ContractionProblem.cpp
+++ b/Tensile/Source/lib/source/ContractionProblem.cpp
@@ -32,54 +32,10 @@
 
 #include <cctype>
 #include <cstddef>
-#include <iomanip>
 #include <set>
 
 namespace Tensile
 {
-
-    std::ostream& operator<<(std::ostream& stream, const SizeMapping& sizeMapping)
-    {
-        return stream << std::right << std::setw(30)
-                      << "workGroupSize: " << sizeMapping.workGroupSize << std::endl
-                      << std::right << std::setw(30) << "threadTile: " << sizeMapping.threadTile
-                      << std::endl
-                      << std::right << std::setw(30) << "macroTile: " << sizeMapping.macroTile
-                      << std::endl
-
-                      << std::right << std::setw(30) << "staggerU: " << sizeMapping.staggerU
-                      << std::endl
-                      << std::right << std::setw(30) << "depthU: " << sizeMapping.depthU
-                      << std::endl
-                      << std::right << std::setw(30) << "globalSplitU: " << sizeMapping.globalSplitU
-                      << std::endl
-                      << std::right << std::setw(30)
-                      << "staggerStrideShift: " << sizeMapping.staggerStrideShift << std::endl
-                      << std::right << std::setw(30)
-                      << "workGroupMapping: " << sizeMapping.workGroupMapping << std::endl
-
-                      << std::right << std::setw(30)
-                      << "packBatchDims: " << sizeMapping.packBatchDims << std::endl
-                      << std::right << std::setw(30)
-                      << "packSummationDims: " << sizeMapping.packSummationDims << std::endl
-                      << std::right << std::setw(30) << "magicDivAlg: " << sizeMapping.magicDivAlg
-                      << std::endl
-                      << std::right << std::setw(30) << "streamK: " << sizeMapping.streamK
-                      << std::endl
-                      << std::right << std::setw(30)
-                      << "persistentKernel: " << sizeMapping.persistentKernel << std::endl
-                      << std::right << std::setw(30)
-                      << "persistentKernelAlongBatch: " << sizeMapping.persistentKernelAlongBatch
-                      << std::endl
-
-                      << std::right << std::setw(30) << "sourceKernel: " << sizeMapping.sourceKernel
-                      << std::endl
-                      << std::right << std::setw(30)
-                      << "globalAccumulation: " << sizeMapping.globalAccumulation << std::endl
-                      << std::right << std::setw(30)
-                      << "workspaceSizePerElemC: " << sizeMapping.workspaceSizePerElemC
-                      << std::endl;
-    }
 
     std::string ContractionProblem::ZeroPad::description() const
     {
--- a/Tensile/Source/lib/source/ContractionProblem.cpp
+++ b/Tensile/Source/lib/source/ContractionProblem.cpp
@@ -36,7 +36,6 @@
 
 namespace Tensile
 {
-
     std::string ContractionProblem::ZeroPad::description() const
     {
         std::ostringstream rv;
--- a/Tensile/Source/lib/source/ContractionProblem.cpp
+++ b/Tensile/Source/lib/source/ContractionProblem.cpp
@@ -34,8 +34,35 @@
 #include <cstddef>
 #include <set>
 
+#include <iomanip>
+
 namespace Tensile
 {
+
+    std::ostream& operator<<(std::ostream& stream, const SizeMapping& sizeMapping)
+    {
+        return stream << std::right << std::setw(30) << "workGroupSize: " << sizeMapping.workGroupSize << std::endl
+                      << std::right << std::setw(30) << "threadTile: " << sizeMapping.threadTile << std::endl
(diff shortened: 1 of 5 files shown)


Tensile/Source/lib/source/ContractionSolution.cpp
```diff
@@ -1146,6 +1207,22 @@ namespace Tensile
...
+
+            vector4<std::uint32_t> matrixInst;
+            if (this->getMatrixInstructionFromKernelName(matrixInst))
+                std:: cout <<  std::right << std::setw(30) << "MatrixInstruction: " << matrixInst << std::endl;
+
+            std::string GSUAlg;
+            if (this->getGSUAlgorithmFromKernelName(GSUAlg))
+                std:: cout <<  std::right << std::setw(30) << "GSUAlgorithm: " << GSUAlg << std::endl;
+
+            if (this->getMatrixInstructionFromKernelName(matrixInst))
```
nakajee, line 1222: Would you avoid calling same function twice?
Please store the result of the first call.

Tensile/Source/lib/source/ContractionProblem.cpp
```diff
@@ -34,8 +34,35 @@
 #include <cstddef>
 #include <set>
 
+#include <iomanip>
```
nakajee, line 37: No need to keep the empty line above.

Tensile/Source/lib/source/ContractionSolution.cpp
```diff
@@ -1138,6 +1148,60 @@ namespace Tensile
...
+
+        std::string kName = this->KernelName();
+
+        if(std::regex_search(kName, matches, miRegex))
+        {
+            if(matches.size() != KERNEL_NAME_MIINST_REGEX_NUM_MATCHES)
+            {
+                return false;
+            }
+            // the first sub_match element (index 0) corresponds to the entire match
```
babakpst, line 1166: Maybe it is shorter this way:
 bool ret = std::regex_search(kName, matches, miRegex) && matches.size() == KERNEL_NAME_MIINST_REGEX_NUM_MATCHES;
        if(ret)
        {
            // the first sub_match element (index 0) corresponds to the entire match
            matrixInst.x = atoi(matches[1].str().c_str());
            matrixInst.y = atoi(matches[2].str().c_str());
            matrixInst.z = atoi(matches[3].str().c_str());
            matrixInst.w = atoi(matches[4].str().c_str());
        }

        return ret;

Tensile/Source/lib/source/ContractionSolution.cpp
```diff
@@ -35,6 +35,16 @@
...
+#include <regex>
+
+#define KERNEL_NAME_MIINST_REGEX "_MI(\\d+)x(\\d+)x(\\d+)x(\\d+)_"
+#define KERNEL_NAME_MIINST_REGEX_NUM_MATCHES \
+    5 //1 (index 0) for entire match and 4 for the sub-experssion matches
+
+#define KERNEL_NAME_GSUA_REGEX "_GSUA([S|M])B_"
+#define KERNEL_NAME_GSUA_REGEX_NUM_MATCHES \
+    2 //1 (index 0) for entire match and 1 for the sub-experssion matches
+
```
babakpst, line 47: We only used these definitions once in the function. Can we not define them and just use the actual definition in the function?

Tensile/Source/lib/source/ContractionSolution.cpp
```diff
@@ -1138,6 +1148,60 @@ namespace Tensile
...
+        std::string kName = this->KernelName();
+
+        if(std::regex_search(kName, matches, miRegex))
+        {
+            if(matches.size() != KERNEL_NAME_GSUA_REGEX_NUM_MATCHES)
+            {
+                return false;
+            }
+            // the first sub_match element (index 0) corresponds to the entire match
```
babakpst, line 1194: Same as the above comment.

Tensile/Source/lib/source/ContractionSolution.cpp
```diff
@@ -1146,6 +1210,21 @@ namespace Tensile
         if(Debug::Instance().printWinningKernelName())
             std::cout << "Running kernel: " << this->KernelName() << std::endl;
 
+        if(Debug::Instance().printKernelCommonParams())
+        {
+            std::cout << "Kernel (" << this->KernelName() << ") Params:" << std::endl;
+
```
babakpst, line 1216: Since the kernel name is long, if there is no line wrapping (in a text file), Params will be missed. I suggest:

std::court << " Kernel name: " << this->KernelName() << std::endl;
std::court << " Kernel parameters: " << std::endl;

babakpst (commented):
Some cosmetic changes. Thanks.

Tensile/Source/lib/source/ContractionSolution.cpp
```diff
@@ -1770,4 +1753,47 @@ namespace Tensile
                       << " shiftPtrElemB=" << st.shiftPtrElemB << " depthUorMT0=" << st.depthUorMT0
                       << " depthUorMT1=" << st.depthUorMT1;
     }
+
+    std::ostream& operator<<(std::ostream& stream, const SizeMapping& sizeMapping)
+    {
+        return stream << std::right << std::setw(30)
+                      << "workGroupSize: " << sizeMapping.workGroupSize << std::endl
+                      << std::right << std::setw(30) << "threadTile: " << sizeMapping.threadTile
```
nakajee, line 1761: Could you put std::right << std::setw(30) only once here?
It looks like all lines here are in the same single sentence.
We need to specify setw only once per sentence.

Tensile/Source/lib/source/ContractionSolution.cpp
```diff
@@ -1151,55 +1144,44 @@ namespace Tensile
...
-        std::string regexp_string(KERNEL_NAME_MIINST_REGEX);
+
+        std::string regexp_string("_MI(\\d+)x(\\d+)x(\\d+)x(\\d+)_");
         std::regex  miRegex(regexp_string);
         std::smatch matches;
 
         std::string kName = this->KernelName();
 
-        if(std::regex_search(kName, matches, miRegex))
+        bool matched = std::regex_search(kName, matches, miRegex) && (matches.size() == 5);
```
nakajee, line 1154: Could you keep the comment you put for define here so that why we use 5 here?

Tensile/Source/lib/source/ContractionSolution.cpp
```diff
@@ -1151,55 +1144,44 @@ namespace Tensile
...
     {
-        std::string regexp_string(KERNEL_NAME_GSUA_REGEX);
+        std::string regexp_string("_GSUA([S|M])B_");
         std::regex  miRegex(regexp_string);
         std::smatch matches;
 
         std::string kName = this->KernelName();
 
-        if(std::regex_search(kName, matches, miRegex))
+        bool matched = std::regex_search(kName, matches, miRegex) && (matches.size() == 2);
```
nakajee, line 1176: same here

Tensile/Source/lib/source/ContractionSolution.cpp
smalekta: This was intentional, I added these for better readability of the print (currently the output looks aligned on `:` and its more readable). Please let me know if you still want me to remove it. Thanks

nakajee: I would like to have comment from other reviewers.
Personally, I would like to make output for each parameter as a single line (like line 1766).
@babakpst, do you think the current one is more readable?

nakajee: After discussing with @babakpst, it looks like this is what we can do.
  setw : we cannot remove them. Need to specify one by one
  right: we need only once. We can remove remaining. Probably better to reset right setting after printing all parameters

babakpst (approved):
Thanks, Siavash.

nakajee (approved):
Looks good.

PR #1809, opened 2023-10-31
Predicate for APU libs

AlexBrownAMD:
Adds new predicate to identify APU libraries. Specialized libraries that set a value (0 or 1) have higher sort priority over general libraries that only specify architecture. Added some basic unit tests, and validated correct function using a sample library on hardware.

### DIFF
--- a/Tensile/Hardware.py
+++ b/Tensile/Hardware.py
@@ -1,6 +1,6 @@
 ################################################################################
 #
-# Copyright (C) 2019-2022 Advanced Micro Devices, Inc. All rights reserved.
+# Copyright (C) 2019-2023 Advanced Micro Devices, Inc. All rights reserved.
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
@@ -33,20 +33,20 @@ class HardwarePredicate(Properties.Predicate):
         return cls("AMDGPU", value=cls("Processor", value=gfxArch))
 
     @classmethod
-    def FromHardware(cls, isa, cuCount=None, dmmAccessFromHost=None):
+    def FromHardware(cls, isa, cuCount=None, isAPU=None):
         gfxArch = Common.gfxName(isa)
-        if cuCount == None and dmmAccessFromHost == None:
+        if cuCount == None and isAPU == None:
             return cls("AMDGPU", value=cls("Processor", value=gfxArch))
         elif cuCount == None:
             return cls("AMDGPU", value=cls.And([cls("Processor", value=gfxArch),
-                                                cls("DMMAccessFromHost", value=dmmAccessFromHost)]))
-        elif dmmAccessFromHost == None:
+                                                cls("IsAPU", value=isAPU)]))
+        elif isAPU == None:
             return cls("AMDGPU", value=cls.And([cls("Processor", value=gfxArch),
                                                 cls("CUCount", value=cuCount)]))
         else:
             return cls("AMDGPU", value=cls.And([cls("Processor", value=gfxArch),
                                                 cls("CUCount", value=cuCount),
-                                                cls("DMMAccessFromHost", value=dmmAccessFromHost)]))
+                                                cls("IsAPU", value=isAPU)]))
 
     def __lt__(self, other):
         # Use superclass logic for TruePreds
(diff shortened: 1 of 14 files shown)


Tensile/Tests/unit/test_HardwarePredicates.py
```diff
@@ -82,8 +108,13 @@ def test_hardware_library_merge_order(libraries):
         lib.merge(lib2)
 
     assert lib.rows[-1]['predicate'] == HardwarePredicate('TruePred')
-    assert lib.rows[0]['predicate'] == HardwarePredicate.FromHardware((9,0,8), 64)
-    assert lib.rows[1]['predicate'] == HardwarePredicate.FromHardware((9,0,8), 60)
+    # assert lib.rows[0]['predicate'] == HardwarePredicate.FromHardware((9,0,8), 64)
```
nakajee: Can we remove these commented out lines?

Tensile/Tests/unit/test_HardwarePredicates.py
```diff
@@ -82,8 +108,13 @@ def test_hardware_library_merge_order(libraries):
...
+    # assert lib.rows[1]['predicate'] == HardwarePredicate.FromHardware((9,0,8), 60)
+    assert lib.rows[0]['predicate'] == HardwarePredicate.FromHardware((9,4,2), isAPU=1)
```
nakajee, line 130: Could you explain a bit about why the order is in this way?

Tensile/Hardware.py
```diff
@@ -53,30 +60,38 @@ def __lt__(self, other):
             myProcPred = next(iter(x for x in myAndPred.value if x.tag == "Processor"), None)
             myCUPred = next(iter(x for x in myAndPred.value if x.tag == "CUCount"), None)
             myCUCount = myCUPred.value if myCUPred != None else 0
+            myIsAPUPred = next(iter(x for x in myAndPred.value if x.tag == "IsAPU"), None)
```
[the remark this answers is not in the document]
AlexBrownAMD: Fixed

Tensile/Hardware.py
```diff
@@ -53,30 +60,38 @@ def __lt__(self, other):
...
+                myProcPredCopy = copy.deepcopy(myProcPred)
+                otherProcPredCopy = copy.deepcopy(otherProcPred)
+                myProcPredCopy.__class__ = otherProcPredCopy.__class__ = Properties.Predicate
+                return myProcPredCopy < otherProcPredCopy
 
-        # Higher priority given to higher CU count
-        return myCUCount > otherCUCount
+            # Higher priority given to higher CU count
+            return myCUCount > otherCUCount
+        return myIsAPU > otherIsAPU
```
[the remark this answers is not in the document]
AlexBrownAMD: Yes, if APU is set then that sorts first since it is a specialized tuning. Sort order is APU, XPU, generic. Updated the comments in this section with more detail.

Tensile/Tests/unit/test_HardwarePredicates.py
AlexBrownAMD: Fixed

Tensile/Hardware.py
```diff
@@ -33,13 +33,20 @@ def FromISA(cls, isa):
         return cls("AMDGPU", value=cls("Processor", value=gfxArch))
 
     @classmethod
-    def FromHardware(cls, isa, cuCount=None):
+    def FromHardware(cls, isa, cuCount=None, isAPU=None):
         gfxArch = Common.gfxName(isa)
-        if cuCount == None:
+        if cuCount == None and isAPU == None:
```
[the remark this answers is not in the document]
AlexBrownAMD: If neither CUCount or IsAPU is set, the the library generating code expects the architecture to just be a "Processor" predicate rather than a dictionary. Could be cleaned up a bit, but it'd require changes across a few other files as well. This might be a good candidate to clean up in the future, but I'd rather leave as is for this PR to unblock tuning work. I'll file a ticket for this one.

Tensile/Tests/unit/test_HardwarePredicates.py
AlexBrownAMD: The library sort order prioritizes specialized tunings over generic ones so that those kernels get picked before the fallback ones.
If an APU mode is set then those libs are sorted first. (IsAPU=1 is APU, IsAPU=0 is XPU, IsAPU=None or -1 is generic).
If a CU count is set then those libs are sorted next, with higher CU counts taking priority (in practice we don't have specialized libs with different CU counts so this is used as "set or not set").
Then everything else is sorted by architecture.
The code technically supports both APU mode and CU count being set, but there are no plans to use that option so ordering there doesn't matter much. But if we did use it, current priority goes to APU mode then CU count, but we could revise if a use-case came up.

nakajee (approved):
Looks good

PR #1812, opened 2023-11-06
Replace std::regex with fnmatch()/PathMatchSpec as a workaround to std::regex stack overflow known bug

rkamd:
Resolves : [SWDEV-427140](https://ontrack-internal.amd.com/browse/SWDEV-427140),

An underlying bug in std::regex ( libstdc++) is causing stack overflow and crashes the application. 

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=86164#c8 

This PR replaces the use of std::regex_search() with fnmatch().

### DIFF
--- a/Tensile/Source/lib/include/Tensile/Serialization/PlaceholderLibrary.hpp
+++ b/Tensile/Source/lib/include/Tensile/Serialization/PlaceholderLibrary.hpp
@@ -31,7 +31,12 @@
 
 #include <Tensile/MasterSolutionLibrary.hpp>
 #include <Tensile/PlaceholderLibrary.hpp>
+#ifdef WIN32
+#include "shlwapi.h"
+#include "windows.h"
+#else
 #include <fnmatch.h>
+#endif
 
 namespace Tensile
 {
@@ -68,7 +73,11 @@ namespace Tensile
                     for(auto condition : ctx->preloaded)
                     {
                         std::string pattern = RegexPattern(condition);
+#ifdef WIN32
+                        if(PathMatchSpec(lib.filePrefix.c_str(), pattern.c_str()))
+#else
                         if(fnmatch(pattern.c_str(), lib.filePrefix.c_str(), 0) == 0)
+#endif
                         {
                             lib.loadPlaceholderLibrary();
                             break;
--- a/Tensile/Source/lib/include/Tensile/Serialization/PlaceholderLibrary.hpp
+++ b/Tensile/Source/lib/include/Tensile/Serialization/PlaceholderLibrary.hpp
@@ -31,7 +31,7 @@
 
 #include <Tensile/MasterSolutionLibrary.hpp>
 #include <Tensile/PlaceholderLibrary.hpp>
-#include <regex>
+#include <fnmatch.h>
 
 namespace Tensile
 {
@@ -68,7 +68,7 @@ namespace Tensile
                     for(auto condition : ctx->preloaded)
                     {
                         std::string pattern = RegexPattern(condition);
-                        if(std::regex_search(lib.filePrefix, std::regex(pattern)))
+                        if(fnmatch(pattern.c_str(), lib.filePrefix.c_str(), 0) == 0)
                         {
                             lib.loadPlaceholderLibrary();
                             break;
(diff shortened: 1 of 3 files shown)


Tensile/Source/lib/include/Tensile/Serialization/PlaceholderLibrary.hpp
```diff
@@ -31,7 +31,7 @@
 
 #include <Tensile/MasterSolutionLibrary.hpp>
 #include <Tensile/PlaceholderLibrary.hpp>
-#include <regex>
+#include <fnmatch.h>
```
daineAMD, line 34: Might need windows-specific code?

rkamd: good catch, I will add support for windows

TorreZuk: Yes for windows sake as well did you consider what I suggested in the meeting?   Can the requirement be represented in simpler terms.  If you can put it in words first, it may be just two calls: strstr and maybe a strncmp suffice for what you need.

Tensile/Source/lib/include/Tensile/Serialization/PlaceholderLibrary.hpp
```diff
@@ -68,7 +74,11 @@ namespace Tensile
                     for(auto condition : ctx->preloaded)
                     {
                         std::string pattern = RegexPattern(condition);
-                        if(std::regex_search(lib.filePrefix, std::regex(pattern)))
+#ifdef WIN32
+                        if(PathMatchSpec(lib.filePrefix.c_str(), pattern.c_str()))
+#else
```
TorreZuk, line 79: Double check if tensile is using specific char set form to avoid runtime errors if build type changes, i.e. PathMatchSpecA may be preferred.   Also appears shlwapi usage varies a lot across rocm, delay load dll to static and other forms of use may cause deployment runtime errors.  See if you can find out from other libraries before adding this new dependency.

Tensile/Source/lib/include/Tensile/Serialization/PlaceholderLibrary.hpp
rkamd: Addressed windows version for pattern matching. @TorreZuk , Sorry I missed this comment. For the current use case, I think it could be done with just strstr. 
- strstr(lib_filename , ArchName) for arch specific 
- strstr(lib_filename, "") for loading all dat files.

My thought process to replace it with an equivalent pattern matching is, if lazy loading is expanded to load datatype specific or something like that, it would be easier to do it with pattern matching. 

I could use strstr if that is preferred.

TorreZuk: if !strlen(pattern) you load them all so that is simple.  Yes if you expect to need a full glob later then strstr may get too tricky.  The way you link did it bring in a .dll or a static library for sh?   We then static link this into rocblas.dll so confirm rocblas dependencies on windows and if it passes tests that is good enough for me.  Just saying it appears there is variation in consuming this library on windows so something to check out.

Tensile/Source/lib/include/Tensile/Serialization/PlaceholderLibrary.hpp
rkamd: shlwapi is used similar to how it is linked in this PR across other libraries. onnxrt uses delayload, but I do not think it might be needed in this case. I have compiled rocblas on windows with this change and it compiles fine and the pattern matching works as expected.

Tensile/Source/lib/include/Tensile/Serialization/PlaceholderLibrary.hpp
```diff
@@ -31,7 +31,13 @@
 
 #include <Tensile/MasterSolutionLibrary.hpp>
 #include <Tensile/PlaceholderLibrary.hpp>
-#include <regex>
+//Replace std::regex, as it crashes when matching long lines(GCC Bug #86164).
+#ifdef WIN32
+#include "shlwapi.h"
+#include "windows.h"
```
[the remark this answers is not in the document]
rkamd: @AlexBrownAMD ,Removed windows.h , as it was not required

AlexBrownAMD (approved):

TorreZuk (approved):
If the windows rocblas build can load tensile kernels then LGTM

PR #1814, opened 2023-11-08
Hotfix: Replace std::regex with fnmatch()/PathMatchSpec as a workaround to std::regex crash

rkamd:
Resolves : [SWDEV-427140](https://ontrack-internal.amd.com/browse/SWDEV-427140),

An underlying bug in std::regex ( libstdc++) is causing stack overflow and crashes the application.

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=86164#c8

This PR replaces the use of std::regex_search() with fnmatch().


TorreZuk (approved):
Looks the same as went to develop

AlexBrownAMD (approved):

rkamd:
Approved for hotfix.

PR #1817, opened 2023-11-14
CI: limit compile to 16 threads

eidenyoshida:
Limit to 16 threads or the number of threads available to the executor, whichever is smaller


babakpst (approved):

babakpst:
@eidenyoshida  Is this required for FC 6.1? I only have 50 min before I start. 
Please let us know if the failures could possibly related to this change. Thanks.

eidenyoshida:
Hi Babak, failures should not be impacted by this change. This only affects how many CPU threads are used in the 'make' step

AlexBrownAMD (approved):

PR #1824, opened 2023-11-23
tensile build with 16 threads

TorreZuk:
* testing compile limiter


TorreZuk:
So this seems to avoid the pre_checkin build failure if you want to approve and merge.  Your extended failures seem to be back to the old buffer bug.  Did you ever raise a compiler ticket for that?

AlexBrownAMD (approved):

PR #1826, opened 2023-11-23
Pre-apply offsets for strided batch kernels

AlexBrownAMD:
Reduce kernel arguments by applying offsets to pointers on host side.

Can't be done for general batch case where pointers are in a buffer, offset code is still present to support general batch kernels used by rocBLAS.

Updated DGEMM NN custom kernel to work with new arguments.

Cleaned up some offset code that is no longer being used.


nakajee:
It looks like hostLibraryTest is failing.
Would you please take a look?

AlexBrownAMD:
> It looks like hostLibraryTest is failing. Would you please take a look?

Fixed failing HostLibraryTests.

nakajee:
I grepped Tensile code with keyword "offsetA" (,B,C,D).
I see some more files using them.
  TensileCreateLibrary.py
  ContractionProblem.hpp
  ContractionProblem.cpp

Don't we need to change these files?

AlexBrownAMD:
> Don't we need to change these files?

The ones in ContractionProblem.hpp/cpp need to stay because we still support offsets, they're just pre-applied when possible (non-batch/strided-batch).

The ones in TensileCreateLibrary can be removed, but that is in a section of unused code, so that's part of a much larger cleanup that needs to happen to remove remnants of the old original client. I'll make a ticket for that.

TorreZuk:
We pre-apply these offsets in rocBLAS already for the two forms non-batched and strided, so is this something we should remove?  Or doesn't really matter as no impact I assume

AlexBrownAMD:
> We pre-apply these offsets in rocBLAS already for the two forms non-batched and strided, so is this something we should remove? Or doesn't really matter as no impact I assume

There's no impact so it can go either way. This Tensile change just allows us to reduce kernel arguments in case a user does pass us an offset value. At a quick scan through rocBLAS code it looked like this would still happen for other L3 functions that use GEMM to solve (like symm or trsm). But if you have the time we could have a call to make sure I didn't miss anything, and to discuss general batch case.

nakajee (approved):
Looks good if all tests pass

AlexBrownAMD:
Fixed a bug affecting rocblas fallback kernels. Ran rocblas gemm tests and all are now passing locally on gfx90a. Waiting for CI to re-run before merging

PR #1828, opened 2023-11-24
Clean up old unused code, mostly related to old client

AlexBrownAMD:


nakajee:
Please fix static analysis fails

PR #1835, opened 2023-11-30
Mode to dynamically adjust number of CUs used

AlexBrownAMD:
This change adds 3 new environment variables to control the way stream-k kernels are run on the GPU.
TENSILE_STREAMK_DYNAMIC_GRID enables dynamic grid mode, which automatically limits the number of CUs used for small problems to a subset based on the number of output tiles. (Default mode is off)
TENSILE_STREAMK_MAX_CUS allows the user to manually set maximum number of CUs used, which could free up some CUs for other operations to run in parallel with gemm. (Default mode is use all CUs)
TENSILE_STREAMK_GRID_MULTIPLIER lets you set how many workgroups are created per CU being used. (Default is 1)

This change also includes an improvement for the two-tile algorithm that allows it to run in full data-parallel mode if the number of output tiles is evenly divided by the number of CUs.

Also includes a bit of code cleanup from previous changeset.

### DIFF
--- a/Tensile/Source/lib/include/Tensile/AMDGPU.hpp
+++ b/Tensile/Source/lib/include/Tensile/AMDGPU.hpp
@@ -27,6 +27,7 @@
 #pragma once
 
 #include <Tensile/Tensile.hpp>
+#include <cstdlib>
 
 namespace Tensile
 {
@@ -203,6 +204,9 @@ namespace Tensile
         int         simdPerCu        = 4;
         int         computeUnitCount = 0;
         int         isAPU            = 0;
+        int         skDynamicGrid    = 0;
+        int         skMaxCUs         = 0;
+        int         skGridMultiplier = 1;
         std::string deviceName;
 
         virtual bool   runsKernelTargeting(Processor p) const;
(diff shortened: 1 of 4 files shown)


Tensile/Source/lib/include/Tensile/AMDGPU.hpp
```diff
@@ -218,6 +222,27 @@ namespace Tensile
 
         virtual std::string description() const;
 
+        const int getSKDynamicGrid() const
+        {
+            static const char* envStr = std::getenv("TENSILE_STREAMK_DYNAMIC_GRID");
```
nakajee, line 227: Where will you document these 3 env parameters?
Wiki? Common.py? Both?

nakajee: If we get better performance with these env variable and generate Tensile kernels based on these specific env values, users always need to specify these env values to get expected performance with StreamK kernels, right?
Can we really expect uses do that?

AlexBrownAMD: I'm not sure of the best place to document this. I think I'll add some notes about them to Common.py for now.
These are added as env values because they are not properties of the stream-k kernel - any stream-k kernel can run with more or fewer workgroups, and there may be benefits or different use-cases for different modes. I don't know if there is an optimal way to use this yet, the team is still researching the effect of this feature. If there is a better default, we can also update the defaults later.
I think there will mainly be 2 modes for most users: the current default is to use 100% of the GPU for each kernel, which is the same as our other data-parallel kernels. The other mode would be dynamic, or to just limit the CU count so other operations can run in parallel. I think this mode would generally be set by frameworks that know they get faster results for larger workflows this way.
We still need to do a bunch of research and data collection to find the best way to use this. The env variables will let us do that testing, and it might show there's a better set of defaults, or that we need a different way of configuring these settings later.

nakajee: I think adding some comment in Common.py is OK for now.
I am OK to use env variables for experimental or debug purpose.
I do not recommend to use env variables for actual tuning because the perf number might not be reproducible without setting correct env variables.
Would you please add some comment to clarify this is only for experimental purpose and not for actual tuning?

nakajee (approved):
I am OK with this change.
We need to make sure we do not use these env variables for rocblas logic file tuning.
(we can adjust them for runtime depending on what we run together)

PR #1840, opened 2023-12-06
Fix LLVM crash issue

AlexBrownAMD:
LLVM crash was caused by make trying to run the build step on embed files before they were finished generating. Embed files can be very large CPP files and may take more than a couple seconds to write to disk. The make target has a dependency on the cpp file, but it parallel builds it seems to activate anytime it notices the file is present, not necessarily when python finishes writing it and closes the handle.

To work around the issue, the embed file is now written to a .temp file which is renamed to .cpp when it is finished and closed. This prevents make from starting the build step early.

There should be a way to accomplish this with a custom cmake target+dependency instead, but my experiments with setting that up caused other build issues like missing build files.

This PR also reverts the recent changes to reduce build threads and increase test timeout to see if this change is enough to fix the build issues for CI.

### DIFF
--- a/.jenkins/common.groovy
+++ b/.jenkins/common.groovy
@@ -35,7 +35,6 @@ def runCompileCommand(platform, project, jobName, boolean debug=false)
     // avoid bug causing long build times of certain files.
     String buildType = 'Release' // debug ? 'Debug' : 'RelWithDebInfo'
     String parallelJobs = "export HIPCC_COMPILE_FLAGS_APPEND='-O3 -Wno-format-nonliteral -parallel-jobs=4'"
-    String buildThreads = '16' // if hipcc is used may be multiplied by parallel-jobs
 
     // comment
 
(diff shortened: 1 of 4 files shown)


.jenkins/common.groovy
```diff
@@ -63,7 +62,7 @@ def runCompileCommand(platform, project, jobName, boolean debug=false)
             pushd build
 
             export PATH=/opt/rocm/bin:\$PATH
-            cmake -DCMAKE_BUILD_TYPE=${buildType} -DCMAKE_CXX_COMPILER=${compiler} -DTensile_CPU_THREADS=${buildThreads} -DTensile_ROOT=\$(pwd)/../Tensile ../HostLibraryTests
```
TorreZuk, line 66: This may have been of use for unrelated OOM kills

Tensile/cmake/TensileConfig.cmake
```diff
@@ -295,7 +295,7 @@ function(TensileCreateLibraryFiles
       # command to be invoked at build time, not cmake time.
       TensileCreateCopyTarget(
       "${Tensile_VAR_PREFIX}_LIBRARY_TARGET"
-      "${Tensile_MANIFEST_CONTENTS}"
+      "${Tensile_EMBED_LIBRARY_SOURCE};${Tensile_MANIFEST_CONTENTS}"
```
TorreZuk, line 298: So this is the real fix to depend on the source, right?  The earlier temp file name isn't any locking system but just adds clarity

nakajee (approved):
All test passed!

.jenkins/common.groovy
AlexBrownAMD: Yea, looks like this was probably needed for gfx11 on CI. Posted PR 1849 to add this back in, but limited to gfx11 builds.

PR #1843, opened 2023-12-07
Kernarg preloading

sdquiring:
- Kernarg preloading is on by default if it's supported by the GPU/remaining kernel parameters.
 - The kernel parameter `DelayRemainingArguments` specifies whether we should initiate the load of the remaining (non-preloaded) arguments at the beginning of the kernel or after we've initiated the first loads of A and B.
 - This feature only supports a small set of kernel parameters. If it proves useful it might be possible to expand
     - for example, only preload the A related arguments if we need some of the sizes.

## Other things that were changed:

 - The ASM cap cache has been moved to its own file (`AsmCaps.py`).  If the `PrintLevel` is >= 1 and the cache is out of date, the output will include what that file should now contain.
 - The AsmRegisterPool now keeps track of which registers are arguments and which are preloaded
    - This is used to generate the metadata in the signature.
 - We should not be allocating registers for unused arguments and then immediately freeing them. This was always a potential bug but these changes caused this to create an actual problem
     - The solution is to conditionally send these arguments in/allocate them.  This is now the case and it should be maintained going forward.


nakajee:
Why precheckin and extended test failed?
Is it related to your change?

Tensile/SolutionStructs.py
```diff
@@ -2954,6 +2954,47 @@ def assignDerivedParameters(state):
...
+      if not globalParameters["AsmCaps"][isa]["KernargPreloading"]:
+        printWarning(f"{isa} doesn't support preloading.")
+        return False
+      
+      if not state["ProblemType"]["StridedBatched"]:
+        printWarning("Preloading only supported in StridedBatched mode.")
+        return False
+
+      dt = state["ProblemType"]["DataType"]
+      if not (dt.isSingle() or dt.isHalf() or dt.isBFloat16):
```
nakajee, line 2967: Don't we need () for isBFloat16?

Tensile/SolutionStructs.py
```diff
@@ -2954,6 +2954,47 @@ def assignDerivedParameters(state):
...
+
+      if not state["ProblemType"]["UseBeta"]:
+        printWarning("Preloading only supported with UseBeta.")
+        return False
+
+      if state["ProblemType"]["UseInitialStridesAB"]:
+        printWarning("Preloading not supported with initial strides AB.")
+        return False
+
+      if state["WorkGroupMapping"] != 0:
```
nakajee, line 2979: Can we changed this to abs(state["WorkGroupMapping"])>1 to allow preloading for WGM=1?

nakajee:
Can we have some some test cases under precheckin or extended to verify this feature?

sdquiring:
> Can we have some some test cases under precheckin or extended to verify this feature?

This is on by default when supported.  But I think it's a good idea to add an explicit test.

nakajee:
Still some fails in CI tests.
Looks like different from known issues.
Please check the fails.

Tensile/KernelWriterAssembly.py
```diff
@@ -1835,49 +1865,47 @@ def initKernel(self, kernel, tPA, tPB ):
...
-      SgprSlot.append(tempSgpr)
-    self.defineSgpr("Tensor2dSizeB", 2, 2)
+    # SgprSlot = []
+    # currentSize = self.sgprPool.size()
+    # while (1):
+    #   tempSgpr = self.sgprPool.checkOut(1,"fill empty slot temporarily",preventOverflow=0)
+    #   if tempSgpr >= currentSize:
+    #     self.sgprPool.checkIn(tempSgpr)
+    #     break
+    #   SgprSlot.append(tempSgpr)
```
sdquiring, line 1878: Delete commented code.

Tensile/Common.py
```diff
@@ -63,7 +64,7 @@
 globalParameters["CodeFromFiles"] = True          # if False byte arrays will be generated during Benchmarking phase as before
 globalParameters["SortProblems"] = False          # sort problems by size; else use order in YAML file
 globalParameters["PinClocks"] = False             # T=pin gpu clocks and fan, F=don't
-globalParameters["HardwareMonitor"] = True        # False: disable benchmarking client monitoring clocks using rocm-smi.
+globalParameters["HardwareMonitor"] = False        # False: disable benchmarking client monitoring clocks using rocm-smi.
```
nakajee, line 67: Is this change intended?
If so, please provide some explanation why you need to change this.

Tensile/Common.py
```diff
@@ -1373,6 +1354,16 @@ def getArchitectureName(gfxName):
     # 0=none, 1=add setprio, 2=add setprio and modify LDS to allow only 2 waves/simd
     "AggressivePerfMode":       [0,1,2],
 
+    # Use the feature whereby 56 bytes of kernel arguments can be preloaded in SGPRs
+    # before the kernel begins executing.  This is currently only supported in cases
+    # where that is enough to initiate the first load of the A and B tensors before
+    # the remaining arguments have been loaded.
+    "PreloadKernelArguments": [False, True],
```
nakajee, line 1361: I thought you would define the default values for these 2 new parameters.
Any benefit of not defining default valu?

Tensile/Source/lib/source/hip/HipSolutionAdapter.cpp
```diff
@@ -35,12 +35,6 @@
 #include <Tensile/hip/HipSolutionAdapter.hpp>
 #include <Tensile/hip/HipUtils.hpp>
 
-//@TODO add alternative for windows
```
nakajee, line 38: Why is this part of code removed?

Tensile/SolutionStructs.py
```diff
@@ -2948,6 +2948,55 @@ def assignDerivedParameters(state):
...
+      if state["ProblemType"]["UseInitialStridesAB"]:
+        return False, "Preloading not supported with initial strides AB."
+
+      if abs(state["WorkGroupMapping"]) > 1:
+        return False, "Preloading not supported with WorkGroupMapping."
+
+      if not state["PrefetchGlobalRead"]:
+        return False, "Preloading must use PrefetchGlobalRead."
+
+      if state["EdgeType"] == "ShiftPtr":
```
nakajee, line 2979: Why can't we use preload arg for ShiftPtr?
Is it difficult to support?
We usually use EdgeType=ShiftPtr.

Tensile/Tests/pre_checkin/preload_kernel_arguments_always_half.yaml
```diff
@@ -0,0 +1,56 @@
+TestParameters:
+  marks: [xfail-gfx900, xfail-gfx906, xfail-gfx908,
```
nakajee, line 2: Will this work on the following arch in the future?
I would like to make sure why you use "xfail" here instead of "skip".

Tensile/Common.py
sdquiring: No, this is unintentional.

Tensile/Common.py
sdquiring: Yes, this way it will default to on if supported, off if not.  We discussed changing the default to -1 or similar if it was causing problems, but it was not the cause of the problems.

I can still make it that way if needed.

Tensile/Source/lib/source/hip/HipSolutionAdapter.cpp
sdquiring: These includes are never used in this file, so this is just unnecessary.

Tensile/SolutionStructs.py
sdquiring: ShiftPtr will use some of the remaining tensor sizes in the calculations to shift the pointers before the first load of A and B.  There's not enough space for those sizes in the preloaded args.

It's possible we could defer shifting the pointers until after the first load but that would require some effort.

Tensile/Tests/pre_checkin/preload_kernel_arguments_always_half.yaml
sdquiring: No they won't.  I suppose we could just use skip here.

nakajee:
I am not sure why .args: section is same between PreloadKernelArguments=0 and 1.
Don't we need to change it for PreloadKernelArguments = 1?

nakajee:
Further optimization ideas:
- Move local read address code after prefetch global read
- Undef AddressC and D after they are used (not sure if this actually makes difference or not...)

Maybe we can implement them after merging this version first.
(I can take over optimization part)

Tensile/KernelWriterAssembly.py
```diff
@@ -2745,7 +2749,9 @@ def functionSignature(self, kernel ):
     # Emit declarations for all sgprs allocated with defineSgpr
     # in the order they were declared
     for skey in self.sgprs:
-      kStr += self.macroRegister("sgpr"+skey, self.sgprs[skey])
+      regNum = self.sgprs[skey]
+      count = self.sgprPool.checkOutSize[regNum] if regNum in self.sgprPool.checkOutSize else "unk"
```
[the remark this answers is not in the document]
sdquiring: The comment is made during `functionSignature` near the end of kernel generation.  When we free the registers, we also get rid of the `checkOutSize` for that register.  In the case of `sgprAddressB` it must have been checked out again, but as a single register.

Tensile/Common.py
```diff
@@ -1373,6 +1354,16 @@ def getArchitectureName(gfxName):
...
+
+    # If PreloadKernelArguments is True, specifies whether we delay initiating the load
+    # of the remaining arguments until after the load of A and B has been initiated.
+    "DelayRemainingArguments": [False, True],
+
```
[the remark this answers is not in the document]
sdquiring: I am not sure, this is something I think we will need to experiment with.

sdquiring:
> I am not sure why .args: section is same between PreloadKernelArguments=0 and 1. Don't we need to change it for PreloadKernelArguments = 1?

I don't think it affects anything but I fixed it.

Tensile/KernelWriterAssembly.py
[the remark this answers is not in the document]
nakajee: > The comment is made during `functionSignature` near the end of kernel generation. When we free the registers, we also get rid of the `checkOutSize` for that register. In the case of `sgprAddressB` it must have been checked out again, but as a single register.

sgprAddressA, B are undefined and reused for other registers.
  .set sgprAddressB, 8 // (1)
  ...
  .set sgprGlobalReadIncsA, 8 // (1)
  .set sgprGlobalReadIncsB, 9 // (1)

number of registers of sgprAddressB should be 2.
Any way to make it accurate?

Tensile/Common.py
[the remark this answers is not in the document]
nakajee: OK. I will try some benchmark with this option once this change is merged.

Tensile/SolutionStructs.py
nakajee: OK. 
I will try benchmark with branch.
If the performance is not as good as ShiftPtr, probably, it is better to move global read prefetch only for A side and enable it for more generic cases.

Tensile/KernelWriterAssembly.py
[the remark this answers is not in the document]
sdquiring: I added a separate data structure to keep these counts.

```
.set sgprKernArgAddress, 0 // (2)
.set sgprWorkGroup0, 16 // (1)
.set sgprWorkGroup1, 17 // (1)
.set sgprWorkGroup2, 18 // (1)
.set sgprLoopCounterL, 19 // (1)
.set sgprOrigLoopCounter, 20 // (1)
.set sgprSrdA, 24 // (4)
.set sgprSrdB, 28 // (4)
.set sgprSrdD, 32 // (4)
.set sgprSrdC, 36 // (4)
.set sgprTensor2dSizeA, 2 // (2)
.set sgprTensor2dSizeB, 4 // (2)
.set sgprAddressD, 40 // (2)
.set sgprAddressC, 42 // (2)
.set sgprAddressA, 6 // (2)
.set sgprAddressB, 8 // (2)
.set sgprAlpha, 10 // (1)
.set sgprBeta, 44 // (1)
.set sgprStridesD, 45 // (2)
.set sgprStridesC, 47 // (2)
.set sgprStridesA, 11 // (2)
.set sgprStridesB, 13 // (2)
.set sgprSizesFree, 49 // (3)
.set sgprSizesSum, 15 // (1)
.set sgprNumWorkGroups0, 52 // (1)
.set sgprNumWorkGroups1, 53 // (1)
.set sgprShadowLimitA, 0 // (2)
.set sgprShadowLimitB, 6 // (2)
.set sgprGlobalReadIncsA, 8 // (1)
.set sgprGlobalReadIncsB, 9 // (1)
.set sgprScalarGlobalReadOffsetA, 54 // (3)
.set sgprScalarGlobalReadOffsetB, 57 // (3)
```

nakajee:
Still some fails in 90a extended.
Do we need to modify custom kernel?
ck tests are failing.

nakajee (approved):
Current CI test result looks OK.

PR #1844, opened 2023-12-08
Remove WGM related kern args if they are not needed

AlexBrownAMD:
WGM mapping arguments are not used if WGM is -1, 0, 1. This change updates the signature so that these arguments don't need to be passed when not used.

Built rocblas using this change and all gemm-related tests passed locally on gfx90a.

### DIFF
--- a/Tensile/Components/Signature.py
+++ b/Tensile/Components/Signature.py
@@ -293,9 +293,10 @@ class SignatureDefault(Signature):
                 kStr += self.addArgument("skExtraIters",        '4', offset,"by_value", "u32"); offset += 4
                 # kStr += self.addArgument("dpTilesPerWG",        '4', offset,"by_value", "u32"); offset += 4
 
-        kStr += self.addArgument(                   "NumFullBlocks",     '4', offset,      "by_value",        "u32"); offset += 4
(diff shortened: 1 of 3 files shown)


nakajee:
The code change looks good.
gfx11 precheckin test failed due to build errors.
Would you check if it is related to your change?

nakajee:
gfx90a precheckin test failed.
The failed case is mOclSolutionAdapterTest.TimingMulti.
Would you please check if this is caused by your change or not?

AlexBrownAMD:
CI errors seem to be unrelated to this change - couldn't recreate the error locally, and rerunning CI has that test passing (now showing errors on gfx11 seen on other runs).

nakajee:
> CI errors seem to be unrelated to this change - couldn't recreate the error locally, and rerunning CI has that test passing (now showing errors on gfx11 seen on other runs).

gfx11 fail looks like the previous issue fixed by your llvm crash fix.
Not sure if this is related to your change.
Do you have any idea why this is happening again with llvm crash fix?

AlexBrownAMD:
Checked with CI team, and suspect gfx11 error is due to OOM on the test node. Going to post a separate PR to test thread reduction just on gfx11 tests.

nakajee (approved):
CI fail is not related to this change.

PR #1845, opened 2023-12-11
More cleanup on unused old client code

AlexBrownAMD:


nakajee:
Please fix static analysis fail.

I am not sure what SolutionWriter was for.
I am OK with this change as long as all test pass.

AlexBrownAMD:
Fixed static analysis. Also built rocblas and ran gemm-related tests against this change.

SolutionWriter was for old client - custom client code was written to launch each solution in old client which is why it required rebuild for each test. New client is more modular by loading kernels from codeobject and logic files.

nakajee (approved):
Thanks for your explanation.

PR #1848, opened 2023-12-13
Adding option of rotating buffers for timing with cache eviction

mahmoodw:
This PR adds the 2 arguments discussed between rocBLAS and Tensile group, which determine the number of buffers that will be cycled through during the hot timing loop. See argument description for usage.

The overall strategy is to create enough buffers and cycle through them on each iteration of the hot loop so that when the same buffer is repeated, the cache no longer contains those values.


Tensile/Source/client/main.cpp
```diff
@@ -523,19 +580,29 @@ int main(int argc, const char* argv[])
...
+        dataInit.push_back(DataInitialization::Get(args, problemFactory, maxWorkspaceSize));
+        listeners.addListener(dataInit[i]);
+    }
+
     listeners.addListener(solutionIterator);
     listeners.addListener(std::make_shared<ProgressListener>(args));
     if(runKernels)
     {
-        listeners.addListener(std::make_shared<ReferenceValidator>(args, dataInit));
+        listeners.addListener(std::make_shared<ReferenceValidator>(args, dataInit[0]));
```
mahmoodw, line 605: Adding only single listener to accommodate validation of the first warmup as previously done

Tensile/Source/client/main.cpp
```diff
@@ -523,19 +580,29 @@ int main(int argc, const char* argv[])
...
 
-    listeners.addListener(dataInit);
+    flush_count = calculate_flush_count(flush_count, 
+                                            flush_memory_size,
+                                            problemFactory);
+
+    for(size_t i = 0; i<flush_count; i++)
+    {
+        dataInit.push_back(DataInitialization::Get(args, problemFactory, maxWorkspaceSize));
+        listeners.addListener(dataInit[i]);
```
mahmoodw, line 598: AFAIK this listener has no effect

Tensile/Source/client/main.cpp
```diff
@@ -640,16 +711,17 @@ int main(int argc, const char* argv[])
...
+                                            kernels[flush_index], stream, startEvents[j], stopEvents[j]));
                                     else
                                         HIP_CHECK_EXC(adapter.launchKernels(
-                                            kernels, stream, nullptr, nullptr));
+                                            kernels[flush_index], stream, nullptr, nullptr));
                                 }
 
                                 listeners.postEnqueues(startEvents, stopEvents);
-                                listeners.validateEnqueues(inputs, startEvents, stopEvents);
+                                listeners.validateEnqueues(inputs[0], startEvents, stopEvents);
```
mahmoodw, line 724: AFAIK this listener has no effect

Tensile/Source/client/main.cpp
```diff
@@ -470,6 +492,41 @@ namespace Tensile
...
+    else if(arg_flush_memory_size != default_arg_flush_memory_size)
+    {
+        size_t cached_size = 0;
+
+        for(auto const& problem : problemFactory.problems())
+            cached_size = std::max(cached_size, problem.a().sizes()[0]*problem.a().sizes()[1]*problem.a().elementBytes() +
+                                                problem.b().sizes()[0]*problem.b().sizes()[1]*problem.b().elementBytes() +
+                                                problem.c().sizes()[0]*problem.c().sizes()[1]*problem.c().elementBytes());
+
+        flush_count = 1 + (arg_flush_memory_size - 1) / cached_size;
```
mahmoodw, line 525: rocBLAS adds some prints here to describe flush count used. Do we want to add that here too?

https://github.com/ROCm/rocBLAS-internal/blob/51a1056edd56a6f0a3f1e40f2f2eb0b2e4138f4f/clients/common/utility.cpp#L469

nakajee: I think it is better to print it in case we specify either flush_count or flush_memory_size.

Tensile/Common.py
```diff
@@ -69,6 +69,29 @@
 globalParameters["SyncsPerBenchmark"] = 1         # how iterations of the stream synchronization for-loop to do per benchmark data point
 globalParameters["EnqueuesPerSync"] = 1           # how many solution enqueues to perform per synchronization
 globalParameters["SleepPercent"] = 300            # how long to sleep after every data point: 25 means 25% of solution time. Sleeping lets gpu cool down more.
+globalParameters["FlushCount"] = 1                # Number of copies of arrays to allocate for cache flushing in timing code.
+                                                  # Functions are called iters times in a timing loop.
+                                                  # If the problem memory footprint is small enough, then arrays will be cached.
+                                                  # flush_count can be used to prevent caching.
+                                                  # For example, for sgemm with transA=transB=N:
+                                                  # problem_memory_footprint = (m*k + k*n + m*n) * sizeof(float).
```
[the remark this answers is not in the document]
nakajee: Maybe using rotate buffer for D (write only) is not necessary?
Is this same as rocblas?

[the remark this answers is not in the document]
mahmoodw: Yes we considered this and decided since it is write only it may not be necessary. We are not 100% sure of any speed up for write only when using MALL. In any case this is the more conservative method to ensure that we are creating more buffers rather than less. We can revisit at a later point if there is more clarity, same as rocBLAS for now

Tensile/Source/client/main.cpp
```diff
@@ -640,16 +711,17 @@ int main(int argc, const char* argv[])
 
                                 for(int j = 0; j < enq; j++)
                                 {
+                                    int flush_index = (j + i + 1) % flush_count;
```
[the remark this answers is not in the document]
mahmoodw: Good point, correcting in the next commit.

Tensile/Source/client/main.cpp
```diff
@@ -470,6 +492,67 @@ namespace Tensile
     } // namespace Client
 } // namespace Tensile
 
+inline void print_memory_size(size_t memory_size)
```
mahmoodw, line 495: Is there a better spot for these helper functions?

nakajee: Not sure actually...
Maybe @AlexBrownAMD can help?

sdquiring:
This feature would be far cleaner to implement within the DataInitialization class.

Tensile/Source/client/main.cpp
```diff
@@ -640,16 +737,17 @@ int main(int argc, const char* argv[])
 
                                 for(int j = 0; j < enq; j++)
                                 {
+                                    int flush_index = (j + i*enq + 1) % flush_count;
```
nakajee, line 740: Is there any reason we use + 1 here?
If I understand the code correctly, I am feeling we do validation for [0].
But, if we start from 1, we might not store the result to [0] if the number of iteration is less than flush buffer num.
Don't we need to worry about that?

nakajee:
Why many CI tests are failing?

By the way, can we have a test case in precheckin or extended to verify this feature?

Tensile/Source/client/main.cpp
mahmoodw: validation is done for [0] but only duiring the warm up loop. We are starting hot loop from 1 to not reuse cache values from warm up.

mahmoodw:
> Why many CI tests are failing?
> 
> By the way, can we have a test case in precheckin or extended to verify this feature?

I'll explore the CI failures. I was hoping I could get some suggestions for a subset of tests to replicate for this feature.

Tensile/Source/client/main.cpp
nakajee: Can NumWarmUp be 0?

nakajee:
> > Why many CI tests are failing?
> > By the way, can we have a test case in precheckin or extended to verify this feature?
> 
> I'll explore the CI failures. I was hoping I could get some suggestions for a subset of tests to replicate for this feature.

This is some quick tips.

(1) static analysis
   just follow the suggestions in the log
(2) precheckin
   it failed at the first yaml.
  The command line is normally in the log.
     python3 ./Tensile/bin/Tensile Tensile/Configs/build_client.yaml /tmp/.tensile-tox/py3/client pid=342

Tensile/Source/client/main.cpp
```diff
@@ -599,31 +692,35 @@ int main(int argc, const char* argv[])
...
                         while(listeners.needMoreRunsInSolution())
                         {
-                            auto inputs = dataInit->prepareGPUInputs(problem);
-
-                            auto kernels = solution->solve(problem, *inputs, *hardware);
+                            std::vector<std::shared_ptr<ContractionInputs>> inputs;
+                            std::vector<std::vector<KernelInvocation>> kernels;
+                            for(size_t i = 0; i<flush_count; i++)
+                            {
+                                inputs.push_back(dataInit[i]->prepareGPUInputs(problem));
```
nakajee, line 699: I am trying this but initialization is very slow.
Can we call initialization only once and copy initialized data to the remaining buffers?

nakajee: Even if I disabled new parameters, Tensile got stuck at "Generating kernels" with this code change...
It must be something wrong. 
Without the change, it is done in less than a minute.

mahmoodw: Adding the parameters to validParameters in Common.py as per Majed's suggestion caused the hang. 

Keep in mind these are global parameters, after the fix and moving the flag from your config with the rest of the global parameters with the following syntax seemed to work fine.

FlushMemorySize: 536870912

nakajee:
Please fix static-analysis fail.
Other fails seem to be known issues.
You can disable them by merging the latest change in develop.

mahmoodw:
> This feature would be far cleaner to implement within the DataInitialization class, turning `m_gpuInputs` into a vector.

After exploring this more it seems simple enough to only add that to DataInitialization, but looks quite a lot more complicated to actually simplify the main loop code by modifying ContractionInputs and then KernelInvocation to hide the extra logic. Please share some insight if you see something that I am missing.

nakajee:
CI test still fails.
Not sure if it is related to your change.
It might be better to re-run CI test.

Tensile/Source/client/main.cpp
AlexBrownAMD: Yea, Utils.cpp would probably be a better spot for this

Tensile/Source/client/main.cpp
AlexBrownAMD: Is the latest version of this changeset still causing slow init and hang problems?

Tensile/Source/client/main.cpp
```diff
@@ -599,31 +691,36 @@ int main(int argc, const char* argv[])
...
-                                    HIP_CHECK_EXC(adapter.launchKernels(kernels,
+                                    HIP_CHECK_EXC(adapter.launchKernels(kernels[0],
                                                                         stream,
                                                                         warmupStartEvents[i],
                                                                         warmupStopEvents[i]));
                                 else
-                                    HIP_CHECK_EXC(
-                                        adapter.launchKernels(kernels, stream, nullptr, nullptr));
+                                    HIP_CHECK_EXC(adapter.launchKernels(
+                                        kernels[0], stream, nullptr, nullptr));
```
AlexBrownAMD, line 718: Warmup is only done with the first buffer.
1) Is this how it is done in rocblas-bench?
2) Does this affect warmup or timed runs in any way?
3) Will the first timed run use the same buffer as the warmup runs?

Tensile/Source/client/main.cpp
nakajee: I have not heard from you about this.
Does this work with NumWarmUp=0?

Tensile/Source/client/main.cpp
mahmoodw: 1. Yes rocblas also only uses first buffer for warmups
2/3. The first timed run will not use the same buffer as warmup, which is the intended behaviour to avoid using cached memory.

AlexBrownAMD (approved):

PR #1849, opened 2023-12-15
Test limiting CI threads for only gfx11

AlexBrownAMD:


TorreZuk (approved):
LGTM, but it really is for any very high core machines not just gfx11 pipe, e.g. gfx942 or if we upgrade more build nodes

AlexBrownAMD:
> LGTM, but it really is for any very high core machines not just gfx11 pipe, e.g. gfx942 or if we upgrade more build nodes

Makes sense, what would you suggest as a good default? I'd like to prevent build errors, but also not slow down other build nodes unnecessarily. I think at 16 threads some nodes were hitting timeout. Did you test if 16 is the max that can be run on the gfx11 node?

TorreZuk:
Well groovy must support some math, you could limit by max(NPROC, 32 or 24) I can't recall what all failed but nodes change.
We do a heuristic in rmake.py for rocBLAS based on RAM, nproc and parallel env, if you have 128 cores you can't give them all big jobs.

AlexBrownAMD:
Good idea. Going to merge this change for now allow builds to continue, and will file a ticket to improve the script, may look for some help from CI team on this one.

TorreZuk:
Yes no rush if works.  min I should have said not max.  Here is some python.
def jobs_heuristic():
    # auto jobs heuristics
    jobs = min(OS_info["NUM_PROC"], 128) # disk limiter
    ram = OS_info["RAM_GB"]
    if (ram >= 16): # don't apply if below minimum RAM
        jobs = min(round(ram/2), jobs) # RAM limiter
        
But on windows have another limited, this was for a windows python bug.
   if os.name == "nt":
        jobs = min(61, jobs) # multiprocessing limit (used by tensile)
       
Not sure when you'll have windows CI running....

PR #1851, opened 2023-12-18
Fix HostLibraryTests on gfx942

AlexBrownAMD:
Test errors were caused by missing sample libraries. gfx942 cannot load libs marked gfx940. This change adds sample libs for gfx942.


nakajee:
Thanks for your quick update!!!
Can we have gfx941 as well?

AlexBrownAMD:
Added sample lib for 941 tests as well. Tested locally on 941 and 942 and all tests passed.

nakajee (approved):
Thanks for your update!

PR #1853, opened 2023-12-20
Limit build threads based on CPUs/RAM available on system

AlexBrownAMD:


.jenkins/common.groovy
```diff
@@ -35,11 +35,11 @@ def runCompileCommand(platform, project, jobName, boolean debug=false)
...
     String parallelJobs = "export HIPCC_COMPILE_FLAGS_APPEND='-O3 -Wno-format-nonliteral -parallel-jobs=4'"
-    String buildThreads = '-1' // if hipcc is used may be multiplied by parallel-jobs
-    if (platform.jenkinsLabel.contains('gfx11'))
-        buildThreads = '16'
-
-    // comment
+    
+    def systemCPUs = sh(script: 'nproc', returnStdout: true ).trim().toInteger()
+    def systemRAM = sh(script: 'free -g | grep -P "[[:digit:]]+" -m 1 -o | head -n 1', returnStdout: true ).trim().toInteger()
+    def maxThreads = min(min(systemCPUs, systemRAM / 10), 64)
```
TorreZuk, line 41: Maybe want a max(1, ...) and systemRAM / 10 , you don't need 10GB per thread do you?  / 2 ?  Otherwise fine builds on all pipes.

TorreZuk (approved):

.jenkins/common.groovy
AlexBrownAMD: Good idea for max(1, ...), I'll add that in the next update.
Not sure if 10GB is needed, but latest gfx11 build still crashed, so may need to reduce threading on that platform further. My code checks total available memory, but maybe I should update it to check available memory to account for other jobs running? But then threading selected may vary widely from run to run.
Also debugging some other unrelated errors that are appearing in CI builds, appear to be caused by CI's update to the latest docker.

AlexBrownAMD:
> Sorry last comment fell into a github commit so almost invisible: "Is this available on all OS and container controls? Why not get CI to tell you the limit, Eiden added ROCBLAS_CLIENT_RAM_GB_LIMIT but we could ask to change it to ROCM_CLIENT_RAM_GB_LIMIT."

That could be handy, since current method is version dependent and may be susceptible to future changes. Will check with Eiden.

PR #1854, opened 2023-12-20
script to summarize rocblas log

babakpst:
This script sorts and categorizes rocblas sizes used in an application.

### DIFF
--- a/tuning/automation/rocblas-parser.py
+++ b/tuning/automation/rocblas-parser.py
@@ -32,115 +32,33 @@ import argparse
 
 def gemmfinder(config):
     output ="unrecognized"
-    # BBS
-    if   (config.f == "gemm_ex" and config.transposeA == "N" and config.transposeB == "N" and config.a_type == "bf16_r" and config.b_type == "bf16_r"  and config.c_type == "bf16_r"  and config.d_type == "bf16_r" and config.compute_type == "f32_r"): 
-        output = "BBS_NN"
-    elif (config.f == "gemm_ex" and config.transposeA == "N" and config.transposeB == "T" and config.a_type == "bf16_r" and config.b_type == "bf16_r"  and config.c_type == "bf16_r"  and config.d_type == "bf16_r" and config.compute_type == "f32_r"): 
-        output = "BBS_NT"
-    elif (config.f == "gemm_ex" and config.transposeA == "T" and config.transposeB == "N" and config.a_type == "bf16_r" and config.b_type == "bf16_r"  and config.c_type == "bf16_r"  and config.d_type == "bf16_r" and config.compute_type == "f32_r"): 
-        output = "BBS_TN"
-    elif (config.f == "gemm_ex" and config.transposeA == "T" and config.transposeB == "T" and config.a_type == "bf16_r" and config.b_type == "bf16_r"  and config.c_type == "bf16_r"  and config.d_type == "bf16_r" and config.compute_type == "f32_r"): 
-        output = "BBS_TT"
-    # HHS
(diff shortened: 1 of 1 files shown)


tuning/automation/rocblas-parser.py
```diff
@@ -0,0 +1,240 @@
...
+# Usage
+# python3 rocblas-parser.py rocblas-configs.log sizes.log
+
+import os
+import argparse
+
+def gemmfinder(config):
+    output ="unrecognized"
+    # BBS
+    if   (config.f == "gemm_ex" and config.transposeA == "N" and config.transposeB == "N" and config.a_type == "bf16_r" and config.b_type == "bf16_r"  and config.c_type == "bf16_r"  and config.d_type == "bf16_r" and config.compute_type == "f32_r"):
```
nakajee, line 36: Can we simplify the code?
At least, we can add "_NN" or "_NT" or "_TN" or "_TT" separately and unify the code for orientation.

Also, we can add "_SB" separately and unify the code for "_SB"

Personally, I prefer to combine "BB" and "BS" like this.
if config.a_type == "bf16_r" and config.b_type == "bf16_r":
  if config.c_type == "bf16_r" and config.d_type == "bf16_r":
      output = "BB"
  elif config.c_type == "f32_r" and config.d_type == "f32_r":
      output = "BS"

tuning/automation/rocblas-parser.py
```diff
@@ -193,9 +116,13 @@ def readfile(filename):
         if not (gemmtype in matrices):
```
nakajee, line 116: I think it is better to skip the following code if gemmtype is not valid (including "unrecognized").
Maybe better to print some message to notify unsupported types?

tuning/automation/rocblas-parser.py
```diff
@@ -0,0 +1,367 @@
...
+    
+        elif ("BSS" in gemmtype):
+            problemDict["rocblas_function"] = "rocblas_gemm_strided_batched_ex" if "SB" in gemmtype else  "rocblas_gemm_ex"
+            problemDict["a_type"] = "bf16_r"
+            problemDict["b_type"] = "bf16_r"
+            problemDict["c_type"] = "f32_r"
+            problemDict["d_type"] = "f32_r"
+            problemDict["compute_type"] = "f32_r"
+
+        elif ("SGEMM" in gemmtype):
```
nakajee, line 269: Can we always use _ex version?
Rotating buffer support is only with _ex.

tuning/automation/rocblas-parser.py
```diff
@@ -0,0 +1,367 @@
...
+            problemDict["transA"] = "N"
+            problemDict["transB"] = "T"
+        elif ("TN" in gemmtype):
+            problemDict["transA"] = "T"
+            problemDict["transB"] = "N"
+        elif ("TT" in gemmtype):
+            problemDict["transA"] = "T"
+            problemDict["transB"] = "T"
+
+        if "SB" in gemmtype:
```
nakajee, line 289: Can we make default iteration smaller?
Plus, please make number of iteration for DB same as SB.

nakajee: I thought SB was SGEMM.
Looks like SB is stridedBatched....
Anyway, I would like to make iteration number smaller by default.

tuning/automation/rocblas-parser.py
babakpst: I'll add CGEMM and ZGEMM.

babakpst:
@nakajee all changes implemented.

nakajee (approved):
I am OK with the update as long as you confirmed it worked as expected.
( I did not have time to try it by myself).

PR #1861, opened 2024-01-10
Stream-K Batch

AlexBrownAMD:
Batch support for stream-k kernels.
Added some new test cases. 
Small fix for rocBLAS liblogic converter script to work with old liblogic files that don't have xf32 parameter listed.


nakajee:
Does this work with non strided batched cases?

nakajee:
Would you create a separate PR for the new tool?
Also, please put some brief comment about how to use it.

AlexBrownAMD:
> Does this work with non strided batched cases?

General batch is not supported, just strided batch or non-batch

nakajee:
> > Does this work with non strided batched cases?
> 
> General batch is not supported, just strided batch or non-batch

In that case, I think it is better to add a reject condition for general batch case.

AlexBrownAMD:
> Would you create a separate PR for the new tool? Also, please put some brief comment about how to use it.

tuning scripts moved to PR 1865

AlexBrownAMD:
> > > Does this work with non strided batched cases?
> > 
> > 
> > General batch is not supported, just strided batch or non-batch
> 
> In that case, I think it is better to add a reject condition for general batch case.

Added reject condition

nakajee:
The code change looks OK.
Do you know why extended CI is not triggered?
I would like to make sure the test cases you added passes.

AlexBrownAMD:
> The code change looks OK. Do you know why extended CI is not triggered? I would like to make sure the test cases you added passes.

Not sure why it didn't start auomtaically - just started a run manually.

nakajee (approved):
Looks good.

PR #1870, opened 2024-01-18
Can we structure the code in order i) to build a wheel ii) to use importlib for embedded non python file

bioinfornatics:
Dear Tensile team,

I would like to package ROCm sofware to industrialize our Rocm DeepLearning environment and Tensile is often a required dependencies. Thus could you move to poetry and use importlib ?
I started as an example what could be done,  it is not intended to be usable.

Thanks


babakpst (changes requested):
on hold until after FC.

bstefanuk:
Thank you for posting this PR, we are taking initiatives to improve the modularity of and importability of Tensile but don't have immediate plans to migrate to poetry. Since thins PR is an example and isn't intended as a complete solution, we are closing it. Another may be opened in the future to address this request.

PR #1877, opened 2024-02-01
Fix BufferLoad=False with stream-k

AlexBrownAMD:
Added new CI test cases for it as well

### DIFF
--- a/Tensile/KernelWriterAssembly.py
+++ b/Tensile/KernelWriterAssembly.py
@@ -5140,8 +5140,24 @@ class KernelWriterAssembly(KernelWriter):
       #kStr += self.bomb(0x13) # after addresses and SRD set
     else:
       tmp = self.vgprPool.checkOut(2, "tmp", self.preventVgprOverflowDuringNewTile)
-      kStr += inst("v_mov_b32", vgpr(tmp+0), sgpr("Address%s+0"%tP["tensorChar"]), "" )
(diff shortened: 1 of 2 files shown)


Tensile/KernelWriterAssembly.py
```diff
@@ -5140,8 +5140,24 @@ def graAddresses(self, kernel, tP, isPap=False):
       #kStr += self.bomb(0x13) # after addresses and SRD set
     else:
       tmp = self.vgprPool.checkOut(2, "tmp", self.preventVgprOverflowDuringNewTile)
-      kStr += inst("v_mov_b32", vgpr(tmp+0), sgpr("Address%s+0"%tP["tensorChar"]), "" )
-      kStr += inst("v_mov_b32", vgpr(tmp+1), sgpr("Address%s+1"%tP["tensorChar"]), "" )
+            
+      if kernel["StreamK"]:
+        # StreamK partial tile - offset to tile start index
+        tmpOffset = self.sgprPool.checkOut(2, "skStartOffset", preventOverflow=0)
```
nakajee, line 5146: Probably, we can simply remove tmpOffset and use tmp instead.
Then, we do not need v_mov_b32 to copy from tmpOffset to tmp.

nakajee: sorry. This is sgpr...

nakajee (approved):
Looks good if all tests pass

AlexBrownAMD:
One test failure on gfx90a, seems unrelated and unable to repro error locally. Going to rerun those tests.

bethune-bryant (approved):

PR #1880, opened 2024-02-06
Disable InitAccVgprOpt for Stream-K

AlexBrownAMD:
Stream-K can split partial tiles at any iteration, and cannot guarantee the minimum number of iterations needed for InitAccVgprOpt


nakajee (approved):
Looks good

PR #1882, opened 2024-02-09
Reduce extended test time

AlexBrownAMD:
This change refactors a number of extended tests, with the main purpose of reducing the overall run-time of the test suite. Some test files had the number of kernels or problem sizes reduced. The aim is to continue testing all the main features listed in the test description, but reduce repeat or similar cases.

Some tests were also refactored in a few ways: A few test files were not testing the particular cases described in the title or comments. Some test groups generated no kernel solutions at all, or none relevant to the test scenario. Some were performing extra steps that are not being validated in this process, like generating unused library logic files. These sorts of issues are also resolved where they were found in this investigation.

I also tried to reduce the total number of combinations being generated by test files. Combinations slow down the test suite in a few ways. First, combinations take time to permute. Some tests were spending significant time permuting solutions where the majority were rejected in SolutionStructs. Second, large number of solution kernels also takes significant time to compile (particularly source kernels). In this case I tried to reduce cases that were too similar, or spread out testing various features among test suites. Finally, some tests would generate kernels that would be skipped for given sizes due to incompatible asserts.

Based on local runs, this PR saves an estimated 2.25 hours of runtime on the extended test suite.


Tensile/Tests/extended/direct_to_lds/dtl_tsgr_f8.yaml
```diff
@@ -1,5 +1,5 @@
 TestParameters:
```
nakajee, line 1: This test is specific to gfx94x. If we skip this, this file is useless.

Tensile/Tests/extended/local_split_u/igemm_lsu_mfma.yaml
```diff
@@ -371,7 +370,7 @@ BenchmarkProblems:
         - SourceSwap: [0,1]
         #- TransposeLDS: [0,1]
         - GlobalReadVectorWidth: [4,8]
-        - VectorWidth: [2,4]
+        - VectorWidth: [4] # 2
```
nakajee, line 373: Could we keep 2 instead of 4?
If we use 4 only, most of MT sizes will not be used because MTWaveTile0 is not multiple of VectorWidth(=4).

Tensile/Tests/extended/direct_to_lds/dtl_tsgr_f8.yaml
AlexBrownAMD: Yea, that was intentional. Currently, new f8 work is happening in TensileLite repo, and there are still a few tests for f8 features in other files in extended tests. Maybe I should have moved the file to disabled rather than mark as skip, but goal was to reduce f8 tests being run. I can re-enable this one if you think these tests are important to keep.

Tensile/Tests/extended/local_split_u/igemm_lsu_mfma.yaml
AlexBrownAMD: Fixed

Tensile/Tests/extended/direct_to_lds/dtl_tsgr_f8.yaml
nakajee: Either way is OK.
We can skip this test for now.
This feature is not actually used in any logic files as long as I know.

nakajee (approved):
Looks good if all tests pass.

PR #1884, opened 2024-02-12
Temporarily disable stream-k tests on gfx94X

AlexBrownAMD:
Disable tests on gfx94X temporarily to allow CI to complete while scalar change is in the works.


nakajee (approved):
Looks good

PR #1885, opened 2024-02-13
Fix StreamK Partials Cache Behavior

bethune-bryant:
The partial store needs the sc1 flag enabled to bypass the cache and provide device wide access.


Tensile/KernelWriterAssembly.py
```diff
@@ -12905,9 +12905,21 @@ def partialsWriteProcedure(self, kernel, vectorWidths, elements, alpha, beta, ed
     kStr += inst("s_waitcnt", "vmcnt(0)", "wait for data store")
     kStr += inst("s_barrier", "store all data before setting flag")
     kStr += inst("s_lshl_b32", sgpr(tmpSgpr), sgpr("StreamKIdx"), log2(4), "flag offset based on CTA index")
-    kStr += inst("s_mov_b32", sgpr(tmpSgpr+2), 1, "flag data")
-    kStr += inst("s_store_dword", sgpr(tmpSgpr+2), sgpr("AddressFlags", 2), sgpr(tmpSgpr), "glc", "set flag")
-    kStr += inst("s_waitcnt", "lgkmcnt(0)", "wait for flag") # TODO just for testing
+    if self.version == (9,4,2): # TODO: Temporary workaround
+      flagReg = self.vgprPool.checkOut(1, "flagReg")
```
nakajee, line 12909: Is that safe to allocate vgpr here?
Can't we use tmpVgpr instead?

Tensile/KernelWriterAssembly.py
```diff
@@ -12905,9 +12905,21 @@ def partialsWriteProcedure(self, kernel, vectorWidths, elements, alpha, beta, ed
...
+      flagOffsetReg = self.vgprPool.checkOut(1, "flagOffsetReg")
+      kStr += inst("v_mov_b32 ", vgpr(flagReg), 1, "flag data")
+      kStr += inst("v_mov_b32 ", vgpr(flagOffsetReg), sgpr(tmpSgpr), "flag offset into vgpr")
+      kStr += inst("s_mov_b64  ", "exec", "0x00000001", "set exec mask for only the first thread")
+      kStr += inst("global_store_dword ", vgpr(flagOffsetReg), vgpr(flagReg), sgpr("AddressFlags", 2), "sc0", "set flag")
+      self.vgprPool.checkIn(flagReg)
+      self.vgprPool.checkIn(flagOffsetReg)
+      kStr += inst("s_mov_b64  ", "exec", "0xFFFFFFFF", "reset exec mask for all threads")
```
nakajee, line 12917: Shouldn't this be 0xFFFFFFFFFFFFFFFF for wave64?

Tensile/KernelWriterAssembly.py
AlexBrownAMD: I'd like to refactor more of our code to allocate temps closer to use at some point (and include names), cause passing tmps the way the code does runs a lot of risks that have caused errors in the past. But yea, this could use tmpVgpr instead, and it might reduce chances of running vgprs over the limit

nakajee: > I'd like to refactor more of our code to allocate temps closer to use at some point (and include names), cause passing tmps the way the code does runs a lot of risks that have caused errors in the past. But yea, this could use tmpVgpr instead, and it might reduce chances of running vgprs over the limit

From code design point of view, it is better to put allocation code closer to the point it is used.
However, if I remember correctly, the reason we have tmpVgpr as an argument is that allocating vgpr in this function (store code generation) might cause some conflict in vgpr allocation.
That is why I said it's safer to use tmpVgpr here.

bethune-bryant: I did have to switch to using the tmpVgpr here to get it to pass the tests. It caused a really weird error.

Tensile/KernelWriterAssembly.py
```diff
@@ -12905,9 +12905,19 @@ def partialsWriteProcedure(self, kernel, vectorWidths, elements, alpha, beta, ed
     kStr += inst("s_waitcnt", "vmcnt(0)", "wait for data store")
     kStr += inst("s_barrier", "store all data before setting flag")
     kStr += inst("s_lshl_b32", sgpr(tmpSgpr), sgpr("StreamKIdx"), log2(4), "flag offset based on CTA index")
-    kStr += inst("s_mov_b32", sgpr(tmpSgpr+2), 1, "flag data")
-    kStr += inst("s_store_dword", sgpr(tmpSgpr+2), sgpr("AddressFlags", 2), sgpr(tmpSgpr), "glc", "set flag")
-    kStr += inst("s_waitcnt", "lgkmcnt(0)", "wait for flag") # TODO just for testing
+    if self.version == (9,4,2): # TODO: Temporary workaround
```
nakajee, line 12908: Maybe 941 or 942 just in case we need 941?

nakajee:
I forgot about test cases.
Could we re-enable stream-k test cases in Tensile\Tests\extended\stream_k for gfx94x (remove skip for 940,941,942)?

nakajee:
The code change looks good.
It is better to make sure these tests pass with your code change on gfx942.

bethune-bryant:
This may not completely resolve the issue. I've run into some strange behavior with it, I'll continue investigating.

AlexBrownAMD (approved):

nakajee:
> I forgot about test cases. Could we re-enable stream-k test cases in Tensile\Tests\extended\stream_k for gfx94x (remove skip for 940,941,942)?

How about the stream-k test cases?

bethune-bryant:
> > I forgot about test cases. Could we re-enable stream-k test cases in Tensile\Tests\extended\stream_k for gfx94x (remove skip for 940,941,942)?
> 
> How about the stream-k test cases?

Sorry, I forgot to do that. They are no longer skipped now.

nakajee (approved):

bethune-bryant:
@nakajee @AlexBrownAMD 
It doesn't look like I have permission to merge this.
Could one of y'all squash&merge it once the tests are finished?

bethune-bryant:
@nakajee @AlexBrownAMD 
Should I restart the failed tests, or is it not unusual for those to fail?

nakajee:
You can ignore gfx1101 precheckin fail.
@AlexBrownAMD is trying to fix that.

PR #1888, opened 2024-02-16
Fix MasterSolutionLibrary indexing for multiple architecture build

yenong-amd:
Implements a fix for solution index collision in multi-architecture build by
- partitioning indices such that each architecture has a separate count by putting architecture type in high 16 bits and using low 16 bits for solution index.
- fallback uses 0000 in high 16 bits.
- optionally write out number of kernels per architecture post build in csv format.

### DIFF
--- a/Tensile/SolutionLibrary.py
+++ b/Tensile/SolutionLibrary.py
@@ -23,6 +23,7 @@
 ################################################################################
 
 import itertools
+import re
 
 from . import Properties
 from . import Hardware
@@ -31,12 +32,6 @@ from . import Contractions
 from .SolutionStructs import Solution as OriginalSolution
 from .Utils import state
 
-architectureList = ['fallback', 'gfx803', 'gfx900', 'gfx906', 'gfx908', 'gfx90a',
-                    'gfx940', 'gfx941', 'gfx942', 'gfx1010', 'gfx1011', 'gfx1012',
-                    'gfx1030', 'gfx1031', 'gfx1032', 'gfx1034', 'gfx1035', 'gfx1100', 
-                    'gfx1101', 'gfx1102']
-solutionIndexMap = {architectureName:int(offset*pow(2,16)) 
-                    for architectureName,offset in zip(architectureList,range(len(architectureList)))}
 
 class SingleSolutionLibrary:
     Tag = "Single"
@@ -254,6 +249,22 @@ class PredicateLibrary:
 
 class MasterSolutionLibrary:
     StateKeys = ["solutions", "library"]
+
+    @classmethod
+    def ArchitectureIndexMap(cls, architectureName):
+        # 'fallback', 'gfx803', 'gfx900', 'gfx906', 'gfx908', 'gfx90a',
+        # 'gfx940', 'gfx941', 'gfx942', 'gfx1010', 'gfx1011', 'gfx1012',
+        # 'gfx1030', 'gfx1031', 'gfx1032', 'gfx1034', 'gfx1035', 'gfx1100',
+        # 'gfx1101', 'gfx1102'
+        archval = -1
+        if architectureName == "fallback":
+            archval = 0
+        elif architectureName.startswith("gfx"):
+            archString = re.search('(?<=gfx)[0-9a-f]*', architectureName)
+            if archString is not None:
+                archLiteral = archString.group(0)
+                archval = (int(archLiteral, 16) << 18)
+        return archval
 
     @classmethod
     def FixSolutionIndices(cls, solutions):
--- a/Tensile/SolutionLibrary.py
+++ b/Tensile/SolutionLibrary.py
@@ -249,6 +249,7 @@ class PredicateLibrary:
 
 class MasterSolutionLibrary:
     StateKeys = ["solutions", "library"]
+    ArchitectureSet = set()
 
     @classmethod
     def ArchitectureIndexMap(cls, architectureName):
@@ -264,6 +265,11 @@ class MasterSolutionLibrary:
             if archString is not None:
                 archLiteral = archString.group(0)
                 archval = (int(archLiteral, 16) << 18)
+        # Check for duplicate architecture values
+        if archval >= 0 and not archval in cls.ArchitectureSet:
+            cls.ArchitectureSet.add(archval)
+        else:
+            raise RuntimeError("ERROR in architecture solution index mapping.")
         return archval
 
     @classmethod
--- a/Tensile/SolutionLibrary.py
+++ b/Tensile/SolutionLibrary.py
@@ -31,6 +31,12 @@ from . import Contractions
 from .SolutionStructs import Solution as OriginalSolution
 from .Utils import state
 
+architectureList = ['fallback', 'gfx803', 'gfx900', 'gfx906', 'gfx908', 'gfx90a',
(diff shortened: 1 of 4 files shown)


Tensile/SolutionLibrary.py
```diff
@@ -31,6 +31,12 @@
 from .SolutionStructs import Solution as OriginalSolution
 from .Utils import state
 
+architectureList = ['fallback', 'gfx803', 'gfx900', 'gfx906', 'gfx908', 'gfx90a',
+                    'gfx940', 'gfx941', 'gfx942', 'gfx1010', 'gfx1011', 'gfx1012',
+                    'gfx1030', 'gfx1031', 'gfx1032', 'gfx1034', 'gfx1035', 'gfx1100', 
+                    'gfx1101', 'gfx1102']
+solutionIndexMap = {architectureName:int(offset*pow(2,16)) 
+                    for architectureName,offset in zip(architectureList,range(len(architectureList)))}
```
TorreZuk, line 39: While you debug you might just want to use the gfx number as is hex for the high 16bits.   Solution indices aren't going to be preserved across release I thought so if there is ever trouble you can drop down to sequence int.

Tensile/SolutionLibrary.py
```diff
@@ -478,17 +484,44 @@ def applyNaming(self, naming=None):
...
+            lazyLibrary = {}
+            for name, lib in self.lazyLibraries.items():
+                reIndexMap = {}
+                newSolutions = {}
+
+                for k, s in lib.solutions.items():
+                    reIndexMap[s.index] = curIndex
+                    s.index = curIndex
+                    newSolutions[curIndex] = s
+                    curIndex += 1
```
TorreZuk, line 497: Should add guard for chosen bucket size overflow, can use constant or 65536

TorreZuk (commented):
Leave it to Tensile team to approve, but I would make as soon as you can add trivial test using your csv output to validate both the csv, and see that counts or something else match actual code object data. Could be post_build cmake step, or just a python test.

Tensile/Source/lib/source/UserDrivenTuningParser.cpp
```diff
@@ -122,6 +122,13 @@ namespace Tensile
             strideB = 1;
             strideC = 1;
 
+            if(b > 1)
```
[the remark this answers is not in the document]
yenong-amd: This assumes general batched and that strided batched will overwrite this computation in lines 144-146.

Tensile/SolutionLibrary.py
yenong-amd: Do you mean that you prefer the upper 16 bits to be, for example, int(0x90a, 16) right shifted 16 bits for gfx90a?

Tensile/SolutionLibrary.py
```diff
@@ -31,6 +31,12 @@
 from .SolutionStructs import Solution as OriginalSolution
 from .Utils import state
 
+architectureList = ['fallback', 'gfx803', 'gfx900', 'gfx906', 'gfx908', 'gfx90a',
+                    'gfx940', 'gfx941', 'gfx942', 'gfx1010', 'gfx1011', 'gfx1012',
+                    'gfx1030', 'gfx1031', 'gfx1032', 'gfx1034', 'gfx1035', 'gfx1100', 
+                    'gfx1101', 'gfx1102']
+solutionIndexMap = {architectureName:int(offset*pow(2,16))
```
nakajee, line 38: By the way, is 16bit large enough to cover each architecture?
What is the current largest number?

yenong-amd: The largest solution set is currently gfx90a with 43056.

Tensile/SolutionLibrary.py
TorreZuk: The idea was just while analyzing and debugging you don't have to look at that order list to figure out the sequence table to gfx number from high 4 bytes, you can just look at it in hex.

Tensile/SolutionLibrary.py
nakajee: Hmm.  It is almost reaching the max.
Maybe better to increase bits for index a bit.
  18 or 20 bit for index for each arch
  14 or 12 bit for arch
or something like this.

yenong-amd: We can do 14 bit arch + 18 bit solution indices.

nakajee:
I see gfx906 test failed with the following error message.

terminate called after throwing an instance of 'std::invalid_argument'
  what():  stoi

I am not sure if this is caused by your change or not.
I did not see gfx906 CI test for a long time (recently re-enabled?)
Would you please check the error?

Tensile/Source/lib/source/UserDrivenTuningParser.cpp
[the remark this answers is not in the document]
yenong-amd: For gemm_batched_ex, rocBLAS implicitly calculates these strides. The current code sets all strides to 1. Without these stride calculations for the override problem, the incoming gemm_batched_ex problem was not matching the override problem. For gemm_strided_batched_ex, the output csv contains specific strides such that if we used that to override best solution, the strides would overwrite these calculations with custom strides provided by the file.

nakajee:
Some merged code from develop is showing up in your change list.
Could you update your branch properly?

Tensile/Source/lib/source/UserDrivenTuningParser.cpp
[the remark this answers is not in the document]
nakajee: I am not sure actually. I am not familiar with rocblas side and not sure when we use 15 parameters and 18 parameters.
As long as you confirmed this worked with both gemm_batched_ex and gemm_strided_batched_ex, I am OK with that.

Tensile/SolutionLibrary.py
yenong-amd: Bucket sizes are now 262144. Instead, I added a check for architecture clobbering.

Tensile/Utilities/validate_library.py
```diff
@@ -0,0 +1,47 @@
+import argparse
```
nakajee, line 1: Would you put copyright sentence here?

nakajee:
I do not have any further comments.
I hope we have some more reviews from other Tensile members (especially from solution selection team).
I will wait for a while.

TorreZuk (approved):
Well as you may have failures in develop now will approve with the hope of pushing other reviewers.    The validation scan is manually run and passed I presume. 
Can it be run on the failing pipeline?

TorreZuk:
@bragadeesh or @AlexBrownAMD who is the scrum master for this sprint?   You should want to get reviews withing a day or so, or @yenong-amd you should assign the people who must review before merge to help push it along.

yenong-amd:
> @bragadeesh or @AlexBrownAMD who is the scrum master for this sprint? You should want to get reviews withing a day or so, or @yenong-amd you should assign the people who must review before merge to help push it along.

I think @lringham was going to test this on some of his tickets.

AlexBrownAMD (approved):

yenong-amd:
@nakajee Can you please help me merge? I don't have merge privilege. Thanks!

nakajee (approved):

nakajee:
> @nakajee Can you please help me merge? I don't have merge privilege. Thanks!

Done

PR #1899, opened 2024-03-06
Fix WorkspaceCheck implementation when used in rocBLAS

AlexBrownAMD:
Update WorkspaceCheck so value is checked correctly during rocBLAS selection. Also refactor predicate code to use canSolve rather than copied versions of the same code.


AlexBrownAMD:
Ran rocBLAS test case provided by Koji and observed corrected solution selection when using this change.

nakajee:
Which function call is to cover rocblas path?

AlexBrownAMD:
> Which function call is to cover rocblas path?

The predicate now does the calculation as needed rather than relying on the value being precomputed by the SolutionIterator. SolutionIterator is used by tuning client, but rocBLAS code path uses SolutionLibrary.

nakajee:
Thanks for your comment.
I am OK with the change once all tests (except for known issues) pass.
Looks like 908 integration test already failed.
Please check if it is caused by your change.

AlexBrownAMD:
> Looks like 908 integration test already failed. Please check if it is caused by your change.

Looks like it was caused by a network connection error (integration test downloads a rocBLAS package to verify TensileCreateLibrary works). Test is passing locally, just relaunched the CI run to see if it passes there as well.

nakajee (approved):
Looks good

AlexBrownAMD:
Thanks for the review. Also ran rocblas gemm tests locally to check it works.

PR #1900, opened 2024-03-09
New dynamic mode 

AlexBrownAMD:
New dynamic mode to run large problems at slightly reduced CU count if it improves work division and power. New mode can be enabled by setting environment variable TENSILE_STREAMK_DYNAMIC_GRID=2.
Option is still being benchmarked and evaluated for best use, but initial tests indicate this option should improve stream-k kernel performance on gfx942.

### DIFF
--- a/Tensile/Source/lib/source/ContractionSolution.cpp
+++ b/Tensile/Source/lib/source/ContractionSolution.cpp
@@ -1597,6 +1597,21 @@ namespace Tensile
         assert(pAMDGPU != nullptr && pAMDGPU->computeUnitCount != 0);
         size_t cuCount = pAMDGPU->computeUnitCount;
         size_t skGrid  = cuCount;
+        if(pAMDGPU->skDynamicGrid == 2 && tiles > skGrid)
+        {
+            for(size_t i = 1; i <= 32; i *= 2)
+            {
+                size_t tilesPerCU = CeilDivide(i * tiles, skGrid);
+                size_t reducedGrid = CeilDivide(i * tiles, tilesPerCU);
+                float utilization = ((float)reducedGrid) / ((float)skGrid);
+                if(utilization > 0.75f)
+                {
+                    if(utilization < 1.0f)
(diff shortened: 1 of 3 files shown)


Tensile/Source/lib/source/ContractionSolution.cpp
```diff
@@ -1597,6 +1597,21 @@ namespace Tensile
         assert(pAMDGPU != nullptr && pAMDGPU->computeUnitCount != 0);
         size_t cuCount = pAMDGPU->computeUnitCount;
         size_t skGrid  = cuCount;
+        if(pAMDGPU->skDynamicGrid == 2 && tiles > skGrid)
```
nakajee, line 1600: Any good location to put some explanation about how TENSILE_STREAMK_DYNAMIC_GRID works?

nakajee: Probably, better to put some brief explanation about TENSILE_STREAMK_DYNAMIC_GRID

nakajee: Does this new logic work with skGridMultiplier?

AlexBrownAMD: Posted an update to add a comment to Common.py alongside the documentation for the other STREAMK environment settings. The comment is brief for now since the exact behaviour and usage is likely to change a bit as it gets more test data collected. But now there's a high-level description and a note on how to use it.
The new logic can also work with the GridMultiplier. I'd guess that combing the two would not be the optimal way to run stream-k, but we need to run more tests to see if there is a use-case for it.

nakajee: Regarding GridMultiplier, what I am thinking is to apply this before calculating reduced grid.
Probably, we do not need to combine them.
You can think about that if needed.

nakajee (approved):

PR #1903, opened 2024-03-25
Env variable to test fixed grid size with SK kernels

AlexBrownAMD:
Small change to add another environment variable to make testing and benchmarking stream-k performance easier.

### DIFF
--- a/Tensile/Source/lib/source/ContractionSolution.cpp
+++ b/Tensile/Source/lib/source/ContractionSolution.cpp
@@ -1604,6 +1604,8 @@ namespace Tensile
     size_t ContractionSolution::getSKGrid(Hardware const& hardware, size_t tiles) const
     {
         AMDGPU const* pAMDGPU = dynamic_cast<AMDGPU const*>(&hardware);
+        if(pAMDGPU->skFixedGrid > 0)
+            return pAMDGPU->skFixedGrid;
         assert(pAMDGPU != nullptr && pAMDGPU->computeUnitCount != 0);
         size_t cuCount = pAMDGPU->computeUnitCount;
         size_t skGrid  = cuCount;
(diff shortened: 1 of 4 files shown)


Tensile/Source/lib/source/ContractionSolution.cpp
```diff
@@ -1604,6 +1604,8 @@ namespace Tensile
     size_t ContractionSolution::getSKGrid(Hardware const& hardware, size_t tiles) const
     {
         AMDGPU const* pAMDGPU = dynamic_cast<AMDGPU const*>(&hardware);
+        if(pAMDGPU->skFixedGrid > 0)
```
nakajee, line 1607: Don't we need max value check below (skMaxCUs or tiles)?
In other words, can we use any skFixedGrid value without limitation?

AlexBrownAMD: Yea this overrides all other options so we can test grids both larger and smaller than CU count. This option makes it easier to script benchmarking sweeps to analyze performance at different grid sizes.

nakajee: OK. Thanks for confirmation.

nakajee (approved):
I do not have further comments.
Not sure why extended test is not triggered, but most likely this change will not affect extended test result.

TorreZuk:
> I do not have further comments. Not sure why extended test is not triggered, but most likely this change will not affect extended test result.

I think everyone has to add ci:extended label now if you want extended pipelines to run

nakajee:
> > I do not have further comments. Not sure why extended test is not triggered, but most likely this change will not affect extended test result.
> 
> I think everyone has to add ci:extended label now if you want extended pipelines to run

OK. Thanks for letting me know. I was not aware of that. 
I will keep that in mind.
For this PR, I think we can skip extended test.

PR #1906, opened 2024-04-04
Atomic 2-tile strean-k and tuning parameter clean-up

AlexBrownAMD:
Add support for atomic version of 2-tile stream-k algorithm.
This change also cleans up the stream-k tuning parameters. Atomic mode is now a separate parameter.
StreamK=0 is off, regular DP kernel
StreamK=1 is basic stream-k
StreamK=2 is two-tile
StreamKAtomic toggles use of atomics for fixup step (and requirements for flags, workspace, and init kernel)
Atomic mode currently only works for SGEMM. Updated the test suite to use new parameters and include atomic 2-tile tests.

### DIFF
--- a/Tensile/Common.py
+++ b/Tensile/Common.py
@@ -1120,9 +1120,8 @@ validParameters = {
     # Total work units are calculated as (#MTs x #LoopIters) and divided among workgroups
     # In most cases each workgroup will calculate a partial tile that are accumulated in a fixup step in the same kernel
     # 0: Standard data-parallel kernel
-    # 1: Basic StreamK atomic (uses atomics to accumulate partial tiles)
-    # 2: Basic StreamK non-atomic (uses workspace to store partial tiles, accumulate in deterministic fix-up step)
-    # 3: Two-Tile StreamK (non-atomic, each WG completes an even number of sk iterations, followed by an even number of dp tiles)
+    # 1: Basic StreamK
+    # 2: Two-Tile StreamK (each WG completes an even number of sk iterations, followed by an even number of dp tiles)
     # StreamK kernels can adjust the number of CUs being used.
     # Using fewer sometimes increases overall throughput by allowing other kernels to run in parallel.
     # StreamK grid is controlled by setting these enviornment variables:
@@ -1138,7 +1137,11 @@ validParameters = {
     #   1 = 1 WG per CU (default)
     # TENSILE_STREAMK_FIXED_GRID lets you override the default grid size with a specific number
     #   0 = override disabled (default)
-    "StreamK": [0, 1, 2, 3],
+    "StreamK": [0, 1, 2],
+    # Determines if StreamK kernel uses atomics
+    # 0: uses workspace to store partial tiles, accumulate in deterministic fix-up step
+    # 1: uses atomics to accumulate partial tiles
+    "StreamKAtomic": [0, 1],
 
     # 0  : standard launch
     # N>0 : launch persistent kernel with N workgroups per compute unit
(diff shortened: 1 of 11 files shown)


Tensile/Common.py
```diff
@@ -1138,7 +1137,11 @@ def getArchitectureName(gfxName):
     #   1 = 1 WG per CU (default)
     # TENSILE_STREAMK_FIXED_GRID lets you override the default grid size with a specific number
     #   0 = override disabled (default)
-    "StreamK": [0, 1, 2, 3],
+    "StreamK": [0, 1, 2],
+    # Determines if StreamK kernel uses atomics
+    # 0: uses workspace to store partial tiles, accumulate in deterministic fix-up step
+    # 1: uses atomics to accumulate partial tiles
+    "StreamKAtomic": [0, 1],
```
nakajee, line 1144: Would you set StreamKAtomic to 0 if StreamK is 0 to avoid generating duplicated kernel for SK=0 + SKA=0 and SK=0 and SKA=0?

Tensile/Source/lib/source/ContractionSolution.cpp
```diff
@@ -419,7 +419,8 @@ namespace Tensile
             }
         }
 
-        if(sizeMapping.globalAccumulation && sizeMapping.streamK < 2)
+        if(sizeMapping.globalAccumulation
```
nakajee, line 422: This part looks inconsistent.
In SolutionStruct.py, _GlobalAccumulation is set only if streamKAtomic = 0.
However, the if condition here is (globalAccumulation and streamKAtomic).

I am thinking we just need to check globalAccumulation here.

Tensile/KernelWriterAssembly.py
```diff
@@ -13444,7 +13444,7 @@ def globalWriteElements(self, kernel, vectorWidths, elements,
       if not self.do["PostLoop"]: return ""
     kStr = ""
     atomic = (kernel["GlobalSplitU"] > 1) and (kernel["_GlobalAccumulation"] != 'MultipleBuffer')
-    atomic = atomic or kernel["StreamK"] == 1
+    atomic = atomic or (kernel["StreamK"] > 0 and kernel["StreamKAtomic"] == 1)
```
nakajee, line 13447: We can simplify this part. Just checking StreamKAtomic==1 is OK.

AlexBrownAMD: Fixed

Tensile/Source/lib/source/ContractionSolution.cpp
AlexBrownAMD: I reviewed the code, and I think it is right the way it is now. For non-atomic stream-k, we set GlobalAccumulation to PartialsBuffer, which is handled separated from the ws_d cause size isn't dependent on D dimensions. For atomic stream-k, GlobalAccumulation is not set because WS is not needed

nakajee: I am a little confused.
This condition is false if streamKAtomic=1 because globalAccumulation is not set, right?
Should this be like this?
  (globalAccumulation && streamK==0) || streamKAtomic

AlexBrownAMD: Oh, yes you're right, that makes the condition false. For stream-k atomic case, workspace pointers don't actually get set in the contraction inputs, and it should skip these arguments. I've updated it to just check
(globalAccumulation && streamK == 0)
Reran the stream-k atomic test locally to make sure it still passes.

nakajee (approved):
Looks good (if all tests pass)

PR #1905, opened 2024-04-02
Hotfix: Fix MasterSolutionLibrary indexing for multiple architecture build (#1888)

yenong-amd:
Fix MasterSolutionLibrary indexing to avoid index collision when building for multiple architectures which causes getAllSolutions gtests to fail.

PR #1897, opened 2024-03-05
Use fallback libraries for archs without optimized logic (v2)

GZGavinZhao:
Fixes #1757. Reintroducing #1862.

Enables architectures that don't have optimized logic files to also produce libraries when `--separate-architectures` or `--lazy-library-loading` is turned on. Previously, one must disable both of these two flags in order for rocBLAS to run on architectures like `gfx1010`.

Previously, there was a bug in Tensile solution indexing that caused #1862 to be reverted. Now, it seems like this issue has been fixed in #1888.

Test plan:
```
cmake -GNinja -B build -S . \
    -DCMAKE_C_COMPILER=hipcc \
    -DCMAKE_CXX_COMPILER=hipcc \
    -DBUILD_CLIENTS_TESTS=ON \
    -DBUILD_CLIENTS_BENCHMARKS=OFF \
    -DBUILD_CLIENTS_SAMPLES=OFF \
    -DBUILD_TESTING=ON \
    -DBUILD_WITH_TENSILE=ON \
    -DTensile_PRINT_DEBUG=ON \
    -DTensile_LIBRARY_FORMAT=msgpack \
    -DTensile_CPU_THREADS=14 \
    -DTensile_LAZY_LIBRARY_LOADING=ON \
    -DAMDGPU_TARGETS="..."
```
With `AMDGPU_TARGETS` being one of the following
- `AMDGPU_TARGETS=gfx1010`
- `AMDGPU_TARGETS=gfx1030;gfx1010`

In all cases,
`$ROCM_PATH/lib/rocblas/library/TensileLibrary_lazy_gfx1010.dat` is produced
and all other `*.dat` files remain unchanged.

In the second case, `./build/clients/staging/rocblas-test --gtest_filter='*gemm_ex_get_solutions*'` that previously failed now passes. I cannot run the full test suite due to limited memory on my GPU (I often get hipOutOfMemory when running stress tests). If this PR doesn't cause extra failures on AMD's CI or if someone can run the full test suite to ensure no additional failures are introduced, then I believe this PR should be good to go. Hopefully this PR can make it in before ROCm 6.1.

PR #1916, opened 2024-04-25
Cherry-pick RDNA1 fix into 6.1 release

GZGavinZhao:
This PR requests to cherry-pick #1897 into ROCm 6.1 release.

#1897 enables RDNA1 (gfx101\*) users to run rocBLAS with the rocblas package from AMD's official repository. Since rocBLAS is almost ubiquitous in ML workflows, this means that all RDNA1 users are essentially unable to run any programs related to ML right now (ROCm/ROCm#2527).

Admittedly, RDNA1 GPUs are [not officially supported](https://rocm.docs.amd.com/projects/install-on-linux/en/latest/reference/system-requirements.html#supported-gpus), but this is such a small change that would benefit _all_ users of RDNA1 GPUs, so I believe is very worth it.

This is a NFC for all architectures besides `gfx1010`.

#1897 requires #1888 to function properly, but it seems like #1888 is already cherry-picked into the `release/rocm-rel-6.1` branch in #1905, so cherry-picking #1897 would not break anything.

### DIFF
--- a/Tensile/TensileCreateLibrary.py
+++ b/Tensile/TensileCreateLibrary.py
@@ -943,11 +943,19 @@ def generateLogicDataAndSolutions(logicFiles, args):
     # logicData[problemType].append((scheduleName, deviceNames, \
     #     solutionsForSchedule, indexOrder, exactLogic, rangeLogic ))
 
+  (archs, _) = splitArchs()
   if globalParameters["SeparateArchitectures"] or globalParameters["LazyLibraryLoading"]:
     if "fallback" in masterLibraries.keys():
       for key, value in masterLibraries.items():
         if key != "fallback":
           value.insert(deepcopy(masterLibraries["fallback"]))
+      for archName in archs:
+        archName = archName.split('-', 1)[0]
+        if archName not in masterLibraries:
+          print1("Using fallback for arch: " + archName)
+          masterLibraries[archName] = deepcopy(masterLibraries["fallback"])
+          masterLibraries[archName].version = args.version
+
       masterLibraries.pop("fallback")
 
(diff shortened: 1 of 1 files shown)


nakajee:
Rocm6.1 release branch is already finalized and no further release plan for 6.1.
This will be included in rocm 6.2.

GZGavinZhao:
@nakajee Thank you for your quick response. Would you mind if I create a PR to `release-staging/rocm-rel-6.2`, or will someone from AMD cherry-pick this internally?

GZGavinZhao:
Nevermind, I see this is already included in `release-staging/rocm-rel-6.2`. Thank you!

cgmb:
> Rocm6.1 release branch is already finalized and no further release plan for 6.1.

To be clear, there is a ROCm 6.1.1 release planned. Though, perhaps there are no changes planned for Tensile specifically.

@GZGavinZhao, it is not a simple process to cherry-pick a change into a patch release. There is an internal change management process for cherry-picks. Unfortunately, it is mostly handled manually and it is very time-consuming. And even a small change in Tensile requires retesting a large suite of libraries and applications for performance and correctness.

This is a great PR and it highlights some of the difficulties with the existing cherry-pick process. I'm hopeful we can improve to the point that a PR like this could be accepted in the future.

GZGavinZhao:
@cgmb I see. I understand that the cherry-picking process may require extensive testing, hence this PR is more of a request to cherry-pick. I thought that maybe some Tensile changes are planned for the next patch release, so if this change can be considered together that'd be great, but releasing in ROCm 6.2 is totally fine as well. Anyway, Thank you so much for your explanation!

PR #1917, opened 2024-05-02
CMake cleanup to prevent redundant work in client builds

ellosel:
The CMake support in Tensile (namely TensileConfig.cmake) is doing unnecessary work. In turn, the configure/build times in downstream projects can take up to 20% longer than is necessary. The unnecessary work is due to:

1. Use of `execute_process` to invoke `TensileCreateLibrary  --generate-manifest-and-exit` which is eagerly run at configuration time regardless if manifest generation is required.
2. Using the contents of TensileManifest.txt as output for a subsequent `add_custom_command` call that is used to generate Tensile build artifacts.

In theory, it is reasonable to use the contents of the manifest as output for a subsequent `add_custom_command` call because this will ensure that the command is not considered satisfied unless all of the files in the manifest are generated. However, CMake needs to know what those files are at configuration time when it is configuring the custom command. Therefore, one must use `execute_process` to generate the manifest which has the consequence of redundantly calling to `TensileCreateLibrary --generate-manifest-and-exit` any time a client reconfigures whether or not it is necessary (and `TensileCreateLibrary --generate-manifest-and-exit` is expensive).

In practice, we would like to invoke `TensileCreateLibrary` once at build time prior to building dependent targets. By relaxing the output requirements of the custom command to just the manifest file. We can eliminate the `execute_process` call altogether. Further, we can provide an analog to checking that the custom command is satisfied (i.e. that it generated all of the expected files) by extending `TensileCreateLibrary` to enforce said check.

This PR includes the necessary changes to avoid the redundant work described above. Further, minor changes to verbosity related logic in `TensileCreateLibrary` were added to ensure users can turn off all output from Tensile.

The current default is set to turn off most of the Tensile output which gives the following output when configuring/building in the client (note the output at 0% which comes from the custom command in TensileConfig):

```
make -j32 install
[  0%] Generating libraries with TensileCreateLibrary
[  0%] Generating /mnt/host/projects/rocBLAS-internal/build/release/include/rocblas/internal/rocblas_device_malloc.hpp
[  1%] Generating prototypes from /mnt/host/projects/rocBLAS-internal/library/src.
[  1%] Built target rocblas_device_malloc
[  1%] Built target rocblas_proto_templates
Tensile::WARNING: Global parameter WriteMasterSolutionIndex = False unrecognized.
[  1%] Verifying files in /mnt/host/projects/rocBLAS-internal/build/release/Tensile/library/TensileManifest.txt were generated
Tensile::WARNING: Global parameter WriteMasterSolutionIndex = False unrecognized.
[  1%] Built target TENSILE_LIBRARY_TARGET
[  1%] Building CXX object Tensile/lib/CMakeFiles/TensileHost.dir/source/ArithmeticUnitTypes.cpp.o
[  1%] Building CXX object Tensile/lib/CMakeFiles/TensileHost.dir/source/AMDGPU.cpp.o
``` 

Users can reconfigure with `-DTensile_VERBOSE=1` or `-DTensile_VERBOSE=2` to get extra tensile output.

Initial testing includes:

- Running: `tox -e lint`
- Running: `tox -v -e py310 -- ./Tensile/Tests -m pre_checkin` 
- Running: `cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_COMPILER=hipcc -DTensile_CPU_THREADS=32 -DTensile_ROOT=$(pwd)/Tensile -S HostLibraryTests -B build; cmake --build build; ./build/TensileTests`

- Integration test with rocBLAS `./install.sh -c -t <path to tensile including these changes>` 
- Running: `./rocblas-test --gtest_filter=*quick*:*pre_checkin*-*known_bug*`

### DIFF
--- a/HostLibraryTests/CMakeLists.txt
+++ b/HostLibraryTests/CMakeLists.txt
@@ -192,10 +192,6 @@ endif()
 
 if(TENSILE_USE_HIP)
 
-    # Make sure the testing libraries get built & copied before the client is built
-    # to avoid segfaults on gtest integration.
-    add_dependencies(TensileTests ${HIP_TEST_LIBRARY_TARGET_DEPS})
-
     target_link_libraries(TensileTests PRIVATE "-Xlinker --whole-archive")
     target_link_libraries(TensileTests PUBLIC ${HIP_TEST_LIBRARIES})
     target_link_libraries(TensileTests PRIVATE "-Xlinker --no-whole-archive")
--- a/HostLibraryTests/hip/CMakeLists.txt
+++ b/HostLibraryTests/hip/CMakeLists.txt
@@ -21,13 +21,12 @@
 # SOFTWARE.
 #
 ################################################################################
-
 set(COMPILER "hipcc")
 set(CODE_OBJECT_VERSION "default")
 
 TensileCreateLibraryFiles(
     "${CMAKE_CURRENT_SOURCE_DIR}/../configs/lite_configs"
-    "${CMAKE_CURRENT_BINARY_DIR}/test_kernels_lite"
+    "${TEST_DATA_DIR}/test_kernels_lite"
     TENSILE_ROOT "${TENSILE_SCRIPT_ROOT}"
     EMBED_LIBRARY test_kernels_lite
     EMBED_KEY     kernels_lite
@@ -38,17 +37,9 @@ TensileCreateLibraryFiles(
     CODE_OBJECT_VERSION ${CODE_OBJECT_VERSION}
     )
 
-# Tests need library and kernels copied to data folder
-TensileCreateCopyTarget(
-    copy_kernels_lite
-    "${LITE_ALL_FILES}"
-    ${TEST_DATA_DIR}/kernels_lite
-    )
-add_dependencies(copy_kernels_lite LITE_LIBRARY_TARGET)
-
 TensileCreateLibraryFiles(
     "${CMAKE_CURRENT_SOURCE_DIR}/../configs/lite_configs_mixed"
-    "${CMAKE_CURRENT_BINARY_DIR}/test_kernels_lite_mixed"
+    "${TEST_DATA_DIR}/test_kernels_lite_mixed"
     TENSILE_ROOT "${TENSILE_SCRIPT_ROOT}"
     EMBED_LIBRARY test_kernels_lite_mixed
     EMBED_KEY     kernels_lite_mixed
@@ -59,17 +50,9 @@ TensileCreateLibraryFiles(
     CODE_OBJECT_VERSION ${CODE_OBJECT_VERSION}
     )
 
-# Tests need library and kernels copied to data folder
-TensileCreateCopyTarget(
-    copy_kernels_lite_mixed
-    "${LITE_MIXED_ALL_FILES}"
-    ${TEST_DATA_DIR}/kernels_lite_mixed
-    )
-add_dependencies(copy_kernels_lite_mixed LITE_MIXED_LIBRARY_TARGET)
-
 TensileCreateLibraryFiles(
     "${CMAKE_CURRENT_SOURCE_DIR}/../configs/lite_configs"
-    "${CMAKE_CURRENT_BINARY_DIR}/test_kernels_lite_2"
+    "${TEST_DATA_DIR}/test_kernels_lite_2"
     TENSILE_ROOT "${TENSILE_SCRIPT_ROOT}"
     EMBED_LIBRARY test_kernels_lite_2
     EMBED_KEY     kernels_lite_2
@@ -80,17 +63,9 @@ TensileCreateLibraryFiles(
     CODE_OBJECT_VERSION ${CODE_OBJECT_VERSION}
     )
 
-# Tests need library and kernels copied to data folder
-TensileCreateCopyTarget(
-    copy_kernels_lite_2
-    "${LITE_2_ALL_FILES}"
-    ${TEST_DATA_DIR}/kernels_lite_2
-    )
-add_dependencies(copy_kernels_lite_2 LITE_2_LIBRARY_TARGET)
-
 TensileCreateLibraryFiles(
     "${CMAKE_CURRENT_SOURCE_DIR}/../configs/tile_aware_selection"
-    "${CMAKE_CURRENT_BINARY_DIR}/test_tile_aware_selection"
+    "${TEST_DATA_DIR}/test_tile_aware_selection"
     TENSILE_ROOT "${TENSILE_SCRIPT_ROOT}"
     EMBED_LIBRARY test_tile_aware_selection
     EMBED_KEY     tile_aware_selection
@@ -100,14 +75,6 @@ TensileCreateLibraryFiles(
     COMPILER_PATH ${CMAKE_CXX_COMPILER}
     CODE_OBJECT_VERSION ${CODE_OBJECT_VERSION}
     )
-
-# Tests need library and kernels copied to data folder
-TensileCreateCopyTarget(
-    copy_kernels_tile_aware_selection
-    "${TILE_AWARE_ALL_FILES}"
-    ${TEST_DATA_DIR}/tile_aware_selection/library
-    )
-add_dependencies(copy_kernels_tile_aware_selection TILE_AWARE_LIBRARY_TARGET)
 
 set(test_yaml rocblas_sgemm_asm_single_kernel.yaml)
 file(COPY ${test_yaml} DESTINATION .)
@@ -127,14 +94,3 @@ set(HIP_TEST_LIBRARIES
     test_kernels_lite_mixed
     test_tile_aware_selection
     TensileClient PARENT_SCOPE)
-
-# Make sure that the TensileTests depends on targets below.
-# Otherwise the TensileTests executable will segfault out.
-set(HIP_TEST_LIBRARY_TARGET_DEPS
-    copy_kernels_lite
-    copy_kernels_lite_mixed
-    copy_kernels_lite_2
-    copy_kernels_tile_aware_selection
-    PARENT_SCOPE
-)
-
--- a/Tensile/Parallel.py
+++ b/Tensile/Parallel.py
@@ -1,6 +1,6 @@
 ################################################################################
 #
-# Copyright (C) 2016-2022 Advanced Micro Devices, Inc. All rights reserved.
+# Copyright (C) 2016-2024 Advanced Micro Devices, Inc. All rights reserved.
 #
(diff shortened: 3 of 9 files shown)


Tensile/cmake/TensileConfig.cmake
```diff
@@ -196,6 +173,11 @@ function(TensileCreateLibraryFiles
     set(Options ${Options} "--no-library-print-debug")
   endif()
 
+  set(Options ${Options} "--verbose=0")
+  if(Tensile_VERBOSE)
+    set(Options ${Options} "--verbose=${Tensile_VERBOSE}")
+  endif()
```
ellosel, line 179:
```suggestion
  if(Tensile_VERBOSE)
    set(Options ${Options} "--verbose=${Tensile_VERBOSE}")
  else()
    set(Options ${Options} "--verbose=0")
  endif()  
```

bstefanuk (approved):
PR looks good, would recommend removing "Minor" from the title.

HostLibraryTests/CMakeLists.txt
```diff
@@ -142,30 +143,39 @@ set(test_sources ${test_sources}
 if(TENSILE_USE_LLVM)
     find_package(LLVM REQUIRED CONFIG)
 
-    set(test_sources ${test_sources}
-        ContractionLibraryLoading_test.cpp
```
cgmillette, line 146: Are these three tests no longer needed?

HostLibraryTests/hip/CMakeLists.txt
```diff
@@ -80,17 +63,9 @@ TensileCreateLibraryFiles(
     CODE_OBJECT_VERSION ${CODE_OBJECT_VERSION}
     )
 
-# Tests need library and kernels copied to data folder
-TensileCreateCopyTarget(
```
cgmillette, line 84: If none of the copy targets are needed, then should we nix the macro?

HostLibraryTests/CMakeLists.txt
```diff
@@ -142,30 +143,39 @@ set(test_sources ${test_sources}
...
-        llvm/LibraryPerformance_test.cpp
-        )
+        llvm/LibraryPerformance_test.cpp)
 endif()
 
 if(Client IN_LIST TENSILE_COMPONENTS)
-    set(test_sources ${test_sources}
+    target_sources(TensileTests PRIVATE
         client/DataInitialization_test.cpp)
+    target_link_libraries(TensileTests PRIVATE Boost::filesystem)
```
cgmillette, line 158: Removal of ${Boost_LIBRARIES} may introduce a  weakness here. If someone were to add a Boost lib requirement in the future, you would have multiple places to update link dependencies between both TensileTests and TensileTestLib

Tensile/TensileCreateLibrary.py
```diff
@@ -141,7 +141,7 @@ def getAssemblyCodeObjectFiles(kernels, kernelWriterAssembly, outputPath):
         else:
           newCOFiles = [os.path.join(destDir, k + '_' + archName + '.co') for k in assemblyKernelNames]
 
-        for src, dst in Utils.tqdm(zip(origCOFiles, newCOFiles), "Copying code objects"):
+        for src, dst in zip(origCOFiles, newCOFiles) if globalParameters["PrintLevel"] == 0 else Utils.tqdm(zip(origCOFiles, newCOFiles), "Copying code objects"):
```
cgmillette, line 144: Is this the replacement for the copy targets?

Tensile/Tests/unit/test_TensileCreateLibrary.py
```diff
@@ -146,3 +147,32 @@ def test_CreateBenchmarkClientParametersForSizes(tmpdir):
...
+    with open("foo.asm", "x"):
+      assert TensileCreateLibrary.verifyManifest(manifestFile), "file in manifest should be on disk"
+  
+    manifest.write("bar.asm\n")
+    manifest.flush()
+    assert not TensileCreateLibrary.verifyManifest(manifestFile), "bar.asm in manifest should not be on disk"
+  
+    with open("bar.asm", "x"):
+      assert TensileCreateLibrary.verifyManifest(manifestFile), "files in manifest should be on disk"  
+
```
cgmillette, line 178: Please do address the 'no newline at EOF'

cgmillette (commented):
In general, I think the changes are good to help reduce the build time. Most of the code movement makes sense!

I would encourage checking a few more things while we are at it:
1) Are the rocBLAS builds compatible with the changes? I think the original work was to prevent rocBLAS from building the entire tensile library multiple times (during cmake stage then during the actual build).
2) If we were to change the contents of a YAML config file (say working on a library contents), would a tensile dirty build detect the changes and build accordingly with new manifest and libs copied to the correct place after? Or do you have to build the entire library from scratch again?
3) Some tests appear to have been removed. Good idea to verify they are not needed anymore.

HostLibraryTests/CMakeLists.txt
ellosel: They were duplicates.

Tensile/TensileCreateLibrary.py
ellosel: Utils.tqdm outputs a progress bar so we are asking if we want to silence all output and if so don't use the progress bar.

Tensile/Tests/unit/test_TensileCreateLibrary.py
ellosel: Good catch thanks

ellosel:
> In general, I think the changes are good to help reduce the build time. Most of the code movement makes sense!
> 
> I would encourage checking a few more things while we are at it:
> 
> 1. Are the rocBLAS builds compatible with the changes? I think the original work was to prevent rocBLAS from building the entire tensile library multiple times (during cmake stage then during the actual build).
> 2. If we were to change the contents of a YAML config file (say working on a library contents), would a tensile dirty build detect the changes and build accordingly with new manifest and libs copied to the correct place after? Or do you have to build the entire library from scratch again?
> 3. Some tests appear to have been removed. Good idea to verify they are not needed anymore.

1. Yes I tested these changes against the current version of develop to confirm we get the behavior we are after. rocBLAS only invokes TensileCreateLibrary during the build stage. There is one issue that remains in rocBLAS that I was going to fix once this is merged (i.e. the virtual env cmake scripts eagerly blow away the tensile install which has the effect of adding some extra config time).
2. Correct, in this implementation, the CMake support is setup to observe changes in the Logic dir and retrigger calls to TensileCreateLibrary as needed. If the logic files don't change then nothing happens. This is a nice feature when you want to do incremental development on rocBLAS without making changes that would impact Tensile artifacts. Currently, every time someone configures rocBLAS they have to eagerly run tensile.
3. Please let me know what tests and I will take another look!

HostLibraryTests/CMakeLists.txt
TorreZuk: @ellosel you can see in the blame I actually went the reverse to allow me to build and run a subset of tests without adding this dependency at all.  As Chris said the variable might allow less maintenance and there are some other boost components used.   As for my needs I am fine if you revert my changes on the link library name

ellosel: > @ellosel you can see in the blame I actually went the reverse to allow me to build and run a subset of tests without adding this dependency at all. As Chris said the variable might allow less maintenance and there are some other boost components used. As for my needs I am fine if you revert my changes on the link library name

Thanks for the input, I will take a look at the blame.

HostLibraryTests/CMakeLists.txt
ellosel: Reverting this change to keep PR more narrowly scoped.

HostLibraryTests/CMakeLists.txt
ellosel: Reverting this change to keep PR more narrowly scoped.

Tensile/Parallel.py
```diff
@@ -79,13 +78,16 @@ def ParallelMap(function, objects, message="", enable=True, multiArg=True):
...
   pcall = pcallWithGlobalParamsMultiArg if multiArg else pcallWithGlobalParamsSingleArg
   pargs = zip(objects, itertools.repeat(globalParameters))
   rv = Parallel(n_jobs=threadCount)(delayed(pcall)(function, a, params) for a, params in pargs)
+  if globalParameters["PrintLevel"] >= 1:
+    print("{0}Done.".format(message))    
+    sys.stdout.flush()  
   
-  print("{0}Done.".format(message))
-  sys.stdout.flush()
   return rv
```
cgmillette, line 93: newline @ EOF

cgmillette:
LGTM pending CI and newline EOF fix

cgmb (commented):
Is `DEPENDS_EXPLICIT_ONLY` applicable to the uses of `add_custom_command`? In theory, it would be possible to run the Tensile build in parallel with the build of the library when compiling with ninja. The linking is the only part that actually needs to be serialized, but `add_custom_command` will force the serialization of the build by default.

Whether any build system has enough memory to build Tensile in parallel with the library perhaps another question, but it may be a nice option to have available.

ellosel:
> Is `DEPENDS_EXPLICIT_ONLY` applicable to the uses of `add_custom_command`? In theory, it would be possible to run the Tensile build in parallel with the build of the library when compiling with ninja. The linking is the only part that actually needs to be serialized, but `add_custom_command` will force the serialization of the build by default.
> 
> Whether any build system has enough memory to build Tensile in parallel with the library perhaps another question, but it may be a nice option to have available.

Good find and I agree that would be nice. It looks like `DEPENDS_EXPLICIT_ONLY` was made available in cmake 3.27 which isn't available on all of the supported rocm platforms e.g. ubuntu 20 is 3.16. Currently, we have cmake_minimum_required set to 3.13.

cgmb:
> It looks like `DEPENDS_EXPLICIT_ONLY` was made available in cmake 3.27 which isn't available on all of the supported rocm platforms e.g. ubuntu 20 is 3.16. Currently, we have cmake_minimum_required set to 3.13.

One approach for those sorts of enhancements is to conditionally enable them. e.g.
```cmake
set(command_options "")
if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.27)
  set(command_options DEPENDS_EXPLICIT_ONLY)
endif()
add_custom_command(mycommand
  ${command_options}
  ...
)
```

TorreZuk:
Not sure what you are trying to build in parallel, but if using all the cores for each stage simpler to debug if steps are done in order.   I would do additions in a new PR if experimenting.

cgmillette (approved):
Looks good! thx

PR #1918, opened 2024-05-08
Two-tile algorithm with SK after DP

AlexBrownAMD:
Alternative implementation of the 2-tile algorithm that does DP tiles first and SK tiles after. This method should have a small boost in performance.

### DIFF
--- a/Tensile/Common.py
+++ b/Tensile/Common.py
@@ -1122,6 +1122,7 @@ validParameters = {
     # 0: Standard data-parallel kernel
     # 1: Basic StreamK
     # 2: Two-Tile StreamK (each WG completes an even number of sk iterations, followed by an even number of dp tiles)
+    # 3: Two-Tile StreamK with DP before SK tiles
     # StreamK kernels can adjust the number of CUs being used.
     # Using fewer sometimes increases overall throughput by allowing other kernels to run in parallel.
     # StreamK grid is controlled by setting these enviornment variables:
--- a/Tensile/Common.py
+++ b/Tensile/Common.py
@@ -1137,7 +1137,7 @@ validParameters = {
     #   1 = 1 WG per CU (default)
     # TENSILE_STREAMK_FIXED_GRID lets you override the default grid size with a specific number
     #   0 = override disabled (default)
-    "StreamK": [0, 1, 2],
+    "StreamK": [0, 1, 2, 3],
     # Determines if StreamK kernel uses atomics
     # 0: uses workspace to store partial tiles, accumulate in deterministic fix-up step
     # 1: uses atomics to accumulate partial tiles
--- a/Tensile/KernelWriterAssembly.py
+++ b/Tensile/KernelWriterAssembly.py
@@ -3851,7 +3851,7 @@ class KernelWriterAssembly(KernelWriter):
 
     if kernel["StreamK"]:
       # StreamK workgroup mapping
-      stmp = self.sgprPool.checkOut(4, "SKMappingTemp", preventOverflow=0)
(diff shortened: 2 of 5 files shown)


Tensile/Common.py
```diff
@@ -1137,7 +1137,7 @@ def getArchitectureName(gfxName):
     #   1 = 1 WG per CU (default)
     # TENSILE_STREAMK_FIXED_GRID lets you override the default grid size with a specific number
     #   0 = override disabled (default)
-    "StreamK": [0, 1, 2],
+    "StreamK": [0, 1, 2, 3],
```
nakajee, line 1140: Please add some comment for 3

Tensile/KernelWriterAssembly.py
```diff
@@ -3856,6 +3895,52 @@ def graWorkGroup(self, kernel, isPap):
...
+        self.sgprPool.checkIn(sIter)
+        kStr += inst("s_add_u32", sgpr(stmp+1), sgpr("StreamKIter"), sgpr(stmp+3), "Offset to start of SK section")
+        kStr += inst("s_add_u32", sgpr("StreamKIterEnd"), sgpr("StreamKIterEnd"), sgpr(stmp+3), "Offset to start of SK section")
+        # clamp to end of iterations
+        # TODO maybe remove clamp, since extra iters code should guarantee total iterations match
+        kStr += inst("s_min_u32", sgpr("StreamKIterEnd"), sgpr("StreamKIterEnd"), sgpr("TotalIters"), "Cap ending iter at total SK iters")
+        # check if this WG has no work to do
+        # TODO Shouldn't need this check!
+        kStr += inst("s_cmp_lt_u32", sgpr("StreamKIter"), sgpr("TotalIters"), "Make sure there's work to do")
+        kStr += self.longBranchScc0("label_%04u" % (self.getLabelNum("KernelEnd")), positiveOnly=True)
```
nakajee, line 3937: Long jump code uses 3 sgpr tmp and increases sgpr usage of this kernel.
Is that possible to release stmp before long jump generation and re-allocate same stmp after long jump code is generated?
This long jump goes to kernel end. No need to keep any values.

nakajee:
Is there any document which explains how this works?
It is difficult to understand the behavior from asm code.
Uploading to the corresponding ticket or any way is OK.

Tensile/Common.py
AlexBrownAMD: Fixed

Tensile/KernelWriterAssembly.py
AlexBrownAMD: Tried a couple ways to change this, but wasn't happy with either change, will send more details.

AlexBrownAMD:
> Is there any document which explains how this works? It is difficult to understand the behavior from asm code. Uploading to the corresponding ticket or any way is OK.

Longer description added to the ticket

Tensile/KernelWriterAssembly.py
nakajee: OK. In that case, we need to add an optional parameter in long jump function to optionally pass tmp sgpr registers to avoid allocating extra sgpr registers.
But, this is an optional.
I am OK to merge the current change for now.

AlexBrownAMD: Just thought the same thing :D going to post the change shortly

AlexBrownAMD: Posted the new change

nakajee (approved):
Thanks for your update.
Looks good.

PR #1919, opened 2024-05-16
Default value for long jump positive and negative

AlexBrownAMD:
I recently changed the long jump function to accept a parameter to optionally reuse temp registers (to reduce peak sgprs). The parameter defaults to None, and then allocates temp registers internally.
The previous change did not set a default "None" for the positive/negative versions, which can cause build errors for some kernel combos. This change sets a default to fix the build.


nakajee (approved):
The change looks good.
Let's make sure if all tests pass or not.

PR #1925, opened 2024-05-28
Update RTD configs

samjwu:
Python 3.10


ellosel:
@samjwu can you please provide some more context for these changes? Also, why is it that all of the tests are still running with docs-only?

samjwu:
> @samjwu can you please provide some more context for these changes? Also, why is it that all of the tests are still running with docs-only?

Sure. 

Currently the rocm-docs-core version here is <1, so this is to update it past version 1.0.0.
The update to the Read the Docs configuration is to use Python 3.10 instead of Python 3.8 since rocm-docs-core version 1.0.0 onwards has 3.10 as a requirement. 

The ci:docs-only label does not totally stop all CI. It just skips certain steps (eg: compile and tests source code), but these checks still run because sometimes libraries include static analysis or linting in other stages.

bstefanuk (approved):

bstefanuk:
Thanks for the additional clarification. Looks good to me, especially since these changes only affect the docs, which are still nascent and subject to change anyways.

ellosel:
> > @samjwu can you please provide some more context for these changes? Also, why is it that all of the tests are still running with docs-only?
> 
> Sure.
> 
> Currently the rocm-docs-core version here is <1, so this is to update it past version 1.0.0. The update to the Read the Docs configuration is to use Python 3.10 instead of Python 3.8 since rocm-docs-core version 1.0.0 onwards has 3.10 as a requirement.
> 
> The ci:docs-only label does not totally stop all CI. It just skips certain steps (eg: compile and tests source code), but these checks still run because sometimes libraries include static analysis or linting in other stages.

Can you confirm that the requirements.txt file is autogenerated? You just bumped the rocm-docs-core version and then generated the requirements.txt file correct?

Otherwise, I concur and marking approved.

PR #1927, opened 2024-05-31
adds tPrint and reconciles printing options

ellosel:
Based on feedback in a recent PR, it is of interest to reconsider how we print diagnostics in TensileCreateLibrary.  Currently, there are three print related options in Tensile:

1. globalParameters["LibraryPrintDebug"] solutions will print enqueue info when enqueueing a kernel
2. globalParameters["PrintLevel"] how much info to print in generator. 0=none, 1=standard, 2=verbose
3. globalParameters["PrintCodeCommands"] print the commands used to generate the code objects (asm,link,hip-clang, etc)

Until recently, `LibraryPrintDebug` was the only print control that had a command line option in TensileCreateLibrary and its' usage does not appear to do anything. That is, it is used to conditionally set the same print option on a generated call to TensileCreateLibrary in the clientWriter code. However, TensileCreateLibrary doesn't actually use the option in practice.
 
More recently the `--verbose` option was added to set PrintLevel which is used in several places to control printing. 
 
Lastly, there is `PrintCodeCommands` which set to false by default and doesn't have a command line option. So to use it in TensileCreateLibrary one would need to manually set it to True in Common.py. The way that it is currently used is if true we make calls to print(...) which doesn't flush to stdout.

This PR removes `LibraryPrintDebug` and `PrintCodeCommands`, and uses `PrintLevel` to control what is printed when executing Tensile. The new levels are:

0. no output
1. standard output
2. 1 + print code commands
3. 2 + verbose

In addition, a `tPrint` function was added that takes a verbosity level to avoid repeating code like print1, print2, and print3. Further, `tPrint` takes a variadic pack of arguments to align the syntax with python `print`.


ellosel:
Need to add VERBOSE option to TensileConfig.cmake

Tensile/Common.py
```diff
@@ -1915,15 +1914,11 @@ def getParamValues( name, structure ):
...
-    print(message)
-    sys.stdout.flush()
-def print2(message):
-  if globalParameters["PrintLevel"] >= 2:
-    print(message)
+def tPrint(verbosity: int, *args):
+  if globalParameters["PrintLevel"] >= verbosity:
+    print(args)
     sys.stdout.flush()
-
```
ellosel, line 1926: remove tab

Tensile/Configs/navi21/rocblas_hgemm_gb_nn_asm_full.yaml
```diff
@@ -1,7 +1,7 @@
 # headers
 GlobalParameters:
   MinimumRequiredVersion: 4.9.0
-  PrintLevel: 1
```
ellosel, line 4: This was change to 2 because it was assumed that the code commands were desired but that may not be the case.

bstefanuk (commented):
Looks good, straightforward change, simple syntax.

Have you run the program with the different verbosity levels to check that the output is what you expected?

babakpst:
Please squash and merge after fixing the conflict. Preferably, we'd like to avoid force-pushed.

PR #1928, opened 2024-05-31
XCC-based workgroup remapping for stream-k kernels

AlexBrownAMD:
By default, workgroups are scheduled in a round-robin fashion. This change adds a mode to reorganize workgroups based on which XCC it is running on for better locality of tiles and improved cache hit rate.


Tensile/KernelWriterAssembly.py
```diff
@@ -3763,6 +3763,31 @@ def openPersistentLoop(self, kernel):
     if kernel["PersistentKernel"] or kernel["StreamK"]:
       if kernel["StreamK"]:
         # Workload calculations
+        if kernel["StreamKXCCMapping"]:
+          sXCC = self.sgprPool.checkOut(1, "XCC", preventOverflow=0)
+          sGridC = self.sgprPool.checkOut(1, "sGridC", preventOverflow=0)
+          sGridF = self.sgprPool.checkOut(1, "sGridF", preventOverflow=0)
+          sGridM = self.sgprPool.checkOut(1, "sGridM", preventOverflow=0)
+          kStr += inst("s_add_u32", sgpr(sGridC), sgpr("skGrid"), 7, "ceil(grid/8)")
```
nakajee, line 3771: Does 8 mean number of XCD?

AlexBrownAMD: Yes, I'm looking into a way of polling a Hip API for the number of XCCs to make this code more generic. Planning to work on that next week and post it in a future PR.

nakajee: Essentially, this seems to be similar to WorkGroupMapping.
We should be able to generalize this logic by using StreamKXCCMapping=number of XCD.

nakajee: But, we need inverse value for non power of 2 cases....
(WGM case, we get it from kernel parameter...)

nakajee: Technically, we have a macro function to calculated generic division (this is used in tensilelite), but I have never tried that before.

nakajee: At least, I would suggest to have an ability to support different number of XCD as a tensile config parameter.
Ideally, it is better generalize the code and use same kernel for any XCD numbers, but I do not have a good idea so far.

Tensile/KernelWriterAssembly.py
```diff
@@ -3763,6 +3763,31 @@ def openPersistentLoop(self, kernel):
...
+          kStr += inst("s_lshr_b32", sgpr(sGridC), sgpr(sGridC), 3, "ceil(grid/8)")
+          kStr += inst("s_lshr_b32", sgpr(sGridF), sgpr("skGrid"), 3, "floor(grid/8)")
+          kStr += inst("s_lshl_b32", sgpr(sGridM), sgpr(sGridF), 3, "mod(grid,8)")
+          kStr += inst("s_sub_u32", sgpr(sGridM), sgpr("skGrid"), sgpr(sGridM), "mod(grid,8)")
```
nakajee, line 3775: We do not need 2 instructions to calculate mod8.
Simply
  sGridM = skGrid & 7

Tensile/KernelWriterAssembly.py
nakajee: I think what I said was not accurate.
If we know the divisor value when generating kernel, we can get magic divisor number and generate any division code.
Then, at least we can generate kernel code for any XCD numbers (we need to have different library for each XCD number)

For further generalization, we need to get number of XCD at runtime and pass the value through kernel parameter. Then, use generic division function to calculate this.
(we can refer division code from tensilelite WGM code)

AlexBrownAMD: Yea I was hoping to make it poll for XCD count, but then it would also become a run-time parameter that needs extra kernel arguments, and the only gain would be consolidating kernels across a couple variants of gfx942 (which should have small number of kernels needed for stream-k anyway).
You're right, this is probably better as a tuning parameter that we can set. Then it's statically generated, doesn't require extra arguments, and we can also use it for different partitioning modes as well (when running on fewer than all XCDs on the device).
I'll work on refactoring the division this week.

Tensile/KernelWriterAssembly.py
AlexBrownAMD: Updated version fixes this

Tensile/KernelWriterAssembly.py
```diff
@@ -3763,6 +3763,35 @@ def openPersistentLoop(self, kernel):
     if kernel["PersistentKernel"] or kernel["StreamK"]:
       if kernel["StreamK"]:
         # Workload calculations
+        if kernel["StreamKXCCMapping"]:
+          sXCC = self.sgprPool.checkOut(1, "XCC", preventOverflow=0)
+          sGridC = self.sgprPool.checkOut(1, "sGridC", preventOverflow=0)
+          sGridF = self.sgprPool.checkOut(1, "sGridF", preventOverflow=0)
+          sGridM = self.sgprPool.checkOut(1, "sGridM", preventOverflow=0)
+          stmp = self.sgprPool.checkOut(2, "stmp", preventOverflow=0)
```
nakajee, line 3771: It might be better to set stmp=None if StreamKXCCMapping is power of 2.

nakajee: We provably need to change the assert condition (no need assert for qReg == tmpSgpr == None case)

Tensile/KernelWriterAssembly.py
```diff
@@ -3763,6 +3763,35 @@ def openPersistentLoop(self, kernel):
...
+          # sGridC = ceil(grid / xccm)
+          kStr += inst("s_add_u32", sgpr(sXCC), sgpr("skGrid"), 7, "ceil(grid/xccm)")
+          kStr += scalarStaticDivideAndRemainder(sGridC, None, sGridC, kernel["StreamKXCCMapping"], stmp, doRemainder=0)
+          # sGridF = floor(grid / xccm)
+          # sGridM = grid % xccm
+          kStr += scalarStaticDivideAndRemainder(sGridF, sGridM, "skGrid", kernel["StreamKXCCMapping"], stmp)
+          # sXCC = wg0 % xccm
+          kStr += scalarStaticDivideAndRemainder(None, sXCC, "WorkGroup0", kernel["StreamKXCCMapping"], stmp, doRemainder=2)
```
nakajee, line 3779: Looks like some bugs in scalarStaticDivideAndRemainder.
- qReg==None seems not working if divisor is not power of 2
- not power of 2 cases, this actually uses 3 tmpSgpr (0,1,2) even though the comment says 2 SGPRs

nakajee: Please try XCC not power of 2 case

Tensile/KernelWriterAssembly.py
nakajee: Or, move assert in else part (not power of 2 case)

Tensile/KernelWriterAssembly.py
```diff
@@ -3768,25 +3768,29 @@ def openPersistentLoop(self, kernel):
...
-          kStr += inst("s_lshr_b32", sgpr(sGridC), sgpr(sGridC), 3, "ceil(grid/8)")
-          kStr += inst("s_lshr_b32", sgpr(sGridF), sgpr("skGrid"), 3, "floor(grid/8)")
-          kStr += inst("s_lshl_b32", sgpr(sGridM), sgpr(sGridF), 3, "mod(grid,8)")
-          kStr += inst("s_sub_u32", sgpr(sGridM), sgpr("skGrid"), sgpr(sGridM), "mod(grid,8)")
-          kStr += inst("s_lshr_b32", sgpr(sXCC), sgpr("WorkGroup0"), 3, "XCC group")
-          kStr += inst("s_lshl_b32", sgpr(sXCC), sgpr(sXCC), 3, "XCC group")
-          kStr += inst("s_sub_u32", sgpr(sXCC), sgpr("WorkGroup0"), sgpr(sXCC), "XCC group")
+          stmp = self.sgprPool.checkOut(2, "stmp", preventOverflow=0)
+          # sGridC = ceil(grid / xccm)
+          kStr += inst("s_add_u32", sgpr(sXCC), sgpr("skGrid"), 7, "ceil(grid/xccm)")
```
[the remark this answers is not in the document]
AlexBrownAMD: Thanks! Fixed

Tensile/KernelWriterAssembly.py
AlexBrownAMD: Updated to optionally pass temp quotient register for non-power of 2 case. Works for 8 XCC case. Trying to track down hardware to test other configurations. I'll make a ticket for testing non-power of 2 XCC hardware and if other updates are needed I'll post them in a separate PR.

Tensile/AsmUtils.py
```diff
@@ -181,19 +181,16 @@ def vectorStaticRemainder(rReg, dReg, divisor, tmpSgpr, comment=""):
...
     return kStr
 
-# only used for loop unroll and GlobalSplitU
+# only used for loop unroll and GlobalSplitU and XCC mapping
 # doRemainder==0 : compute quotient only
 # doRemainder==1 : compute quotient and remainder
 # doRemainder==2 : only compute remainder (not quotient unless required for remainder)
 # dreg == dividend
-# tmpSgpr must be 2 SPGRs
+# tmpSgpr must be 2 SPGRs (can be None if divisor is power of 2)
```
nakajee, line 189: Would you please change this comment (2 SPGRs -> 3)?
(SPGR -> SGPR....)

Tensile/KernelWriterAssembly.py
```diff
@@ -3763,6 +3763,42 @@ def openPersistentLoop(self, kernel):
...
+          sXCC = self.sgprPool.checkOut(1, "XCC", preventOverflow=0)
+          sGridC = self.sgprPool.checkOut(1, "sGridC", preventOverflow=0)
+          sGridF = self.sgprPool.checkOut(1, "sGridF", preventOverflow=0)
+          sGridM = self.sgprPool.checkOut(1, "sGridM", preventOverflow=0)
+          stmp = None
+          sqtmp = None
+          divisor = kernel["StreamKXCCMapping"]
+          if ((divisor & (divisor - 1)) != 0): # Need temp registers if not power of 2
+            stmp = self.sgprPool.checkOut(3, "stmp", preventOverflow=0)
+            sqtmp = stmp + 2
```
nakajee, line 3776: This is overwrapping with tmpSgpr+ 2 in scalarStaticDivideAndRemainder.
Looks like using same sgpr for qReg and tmpSgpr+2 does not actually cause an issue, but it might be better to put some comment so that using same reg for qReg and tmpSgpr+2 is intentional.

Tensile/AsmUtils.py
AlexBrownAMD: Updated

Tensile/KernelWriterAssembly.py
AlexBrownAMD: Updated

nakajee (approved):
Thanks for your update.
This change looks good.
Calling scalarStaticDivideAndRemainder seems problematic.
Many locations calling scalarStaticDivideAndRemainder do not actually allocate 3 tmpSgpr...
We will need another PR to fix the code using scalarStaticDivideAndRemainder.

AlexBrownAMD:
> We will need another PR to fix the code using scalarStaticDivideAndRemainder.

I will create a ticket to keep track of this

PR #1929, opened 2024-05-31
Analytical grid size prediction model for Stream-K

AlexBrownAMD:
This change integrates the analytical grid size prediction model developed by @neoblizz . The model is meant to predict optimal launch grid-size for a given problem when running stream-k based on the time it takes to complete various steps such as data parallel tiles, partial tiles, workspace writes, and fixup step coordination.

The new mode can be enabled by setting TENSILE_STREAMK_DYNAMIC_GRID=3 and is currently experimental. A future update may need to adjust the prediction weights. Once research is complete this mode will become the default option and the environment variable settings will be updated.


Tensile/Source/lib/source/ContractionSolution.cpp
```diff
@@ -40,6 +41,119 @@
...
+            int iters_per_cta = num_iters_per_cta(BLK_M, BLK_N, BLK_K, m, n, k, g);
+            int fixup_peers = num_fixup_peers(BLK_K, k, g, iters_per_cta);
+
+            return {a + (b * (fixup_peers > 1)) + (c * iters_per_cta) + (d * (fixup_peers - 1)), iters_per_cta, fixup_peers};
+        }
+
+        int best_predicted_grid_size(int BLK_M,
+                                     int BLK_N,
+                                     int BLK_K,
+                                     int m,
```
nakajee, line 99: Shouldn't this be size_t?
Can we use 32bit type?

Tensile/Source/lib/source/ContractionSolution.cpp
```diff
@@ -40,6 +41,119 @@
...
+            return m_tiles * n_tiles;
+        }
+
+        constexpr int num_fixup_peers(int BLK_K, int k, int g, int iters_per_cta)
+        {
+            return math::ceil_div(math::ceil_div(k, BLK_K), iters_per_cta);
+        }
+
+        std::tuple<double,int,int>
+        predicted_runtime(int BLK_M, int BLK_N, int BLK_K, int m, int n, int k, int g, int a, int b, int c, int d)
```
nakajee, line 88: Shouldn't a,b,c,d be double?

Tensile/Source/lib/source/ContractionSolution.cpp
```diff
@@ -40,6 +41,119 @@
...
+                                     int BLK_K,
+                                     int m,
+                                     int n,
+                                     int k,
+                                     int grid_start = 1,
+                                     int grid_end = 304)
+        {
+            static const bool debug = Debug::Instance().printStreamKGridInfo();
+            // Timing in us @ 1100/900 MHz
+            double a = 5.04 + 8.3; // 5544 + 9136;  // 5.04 us
```
nakajee, line 107: I do not understand what these values mean, but I am feeling this logic is too specialized to specific arch.

Tensile/Source/lib/source/ContractionSolution.cpp
```diff
@@ -40,6 +41,128 @@
...
 {
+    namespace streamk
+    {
+        namespace math
+        {
+            /**
+             * Performs `(n + d - 1) / d` (unsafe).
+             */
+            template <typename N, typename D>
+            __device__ __host__ inline constexpr N ceil_div(N n, D d)
```
nakajee, line 52: Where is this used?

Tensile/Source/lib/source/ContractionSolution.cpp
```diff
@@ -1601,36 +1724,92 @@ namespace Tensile
...
+        // Dynamically predict the best grid-size by weighing the cost of the fix-up
+        // step and the cost of processing MAC-loop instructions. When the cost of fix-up
+        // is the bottleneck, use smaller grid size.
+        // Architecture dependent.
+        else if(pAMDGPU->skDynamicGrid == 3)
+        {
+            dim3 dSize(1, 1, 1);
+            for(size_t i = 0; i < problem.freeIndicesA().size(); i++)
+            {
+                dSize.x *= problem.freeSizeA(i);
```
nakajee, line 1779: As I said in my previous comment, I thought this could be 64 bit value.
Are you sure we can use 32bit variable here?
I suggest to use size_t for x,y,z.

Tensile/Source/lib/source/ContractionSolution.cpp
```diff
@@ -1601,36 +1724,92 @@ namespace Tensile
...
+            {
+                dSize.z *= problem.boundSize(i);
+            }
+
+            return streamk::best_predicted_grid_size(sizeMapping.macroTile.x, sizeMapping.macroTile.y, sizeMapping.depthU, dSize.x, dSize.y, dSize.z, cuCount);
+        }
+
+        // Limit the CUs Stream-K is launched on either max or the specified,
+        // whichever is minimum.
+        else if(pAMDGPU->skMaxCUs > 0)
```
nakajee, line 1796: If we change this to else if, I suggest to update comments in common.py so that we know the priority of env parameters (TENSILE_STREAMK_DYNAMIC_GRID > TENSILE_STREAMK_MAX_CUS > TENSILE_STREAMK_GRID_MULTIPLIER)

Tensile/Source/lib/source/ContractionSolution.cpp
neoblizz: Do you want me to update that for all M, N, K, and BLK_M/N/K to size_t?

nakajee: BLK_M/N/K is MacroTile.x/y, depthU, right?
I think 32 bit is ok for them.
Once we multiply multiple sizes (like freeSizeA(0)*(1)), I think it should be 64bit value (size_t).

neoblizz: Got it, I will make those changes.

Tensile/Source/lib/source/ContractionSolution.cpp
```diff
@@ -1601,36 +1714,92 @@ namespace Tensile
...
+            return min(skGrid, tiles);
+        }
+
+        // Dynamically predict the best grid-size by weighing the cost of the fix-up
+        // step and the cost of processing MAC-loop instructions. When the cost of fix-up
+        // is the bottleneck, use smaller grid size.
+        // Architecture dependent.
+        else if(pAMDGPU->skDynamicGrid == 3)
+        {
+            dim3 dSize(1, 1, 1);
```
nakajee, line 1766: I did not clearly mention, but we should not use dim3 here.
The type of each member is uint32.
Please define size_t variables instead.

nakajee:
Please resolve the confict.

Tensile/Source/lib/source/ContractionSolution.cpp
```diff
@@ -1763,22 +1763,22 @@ namespace Tensile
         // Architecture dependent.
         else if(pAMDGPU->skDynamicGrid == 3)
         {
-            dim3 dSize(1, 1, 1);
+            size_t x, y, z = 1;
```
nakajee, line 1766: Shouldn't this be like this?
  x=1, y=1,z=1
Are x and y really initialized?

neoblizz: My bad, it should be.

nakajee:
The code change looks OK.
Please fix static-analysis fails.

nakajee:
Wait. Some fail in 90a precheckin... Not sure what is wrong...

nakajee:
Please check with @AlexBrownAMD to see if the fail is related to this change or not.

AlexBrownAMD:
> Please check with @AlexBrownAMD to see if the fail is related to this change or not.

90a failers do not appear to be related, but I just fixed clang format errors, so we can let the 90a tests runs again

nakajee:
90a test failed again.
I just added gfx942 label to make sure no fails on gfx942.

nakajee (approved):
gfx942 test passed.
Looks good.

neoblizz (approved):

PR #1932, opened 2024-06-05
Add Python unit test coverage report

bstefanuk:
**Objectives:**
- Add unit test coverage reporting for Tensile's Python applications.
- Refactor the location in CI where unit tests are being invoked.
- Clarify boundaries between tests of different types both in CI and in tox.ini
- Move towards standardized function definitions in groovy files

**Outcomes:**
- Unit tests are now run in the testing stage of the pipelines instead of the compile stage.
- Changes to testing environments available via `tox run -e <ENV>`
- Refactored dependencies and structure of tox.ini
- Removed unused lines of code in groovy bash commands.

**Notable changes:**
- JUnit reports are no longer created and saved in CI. This shadows other ROCm libraries which are removing JUnit because of it increases job times and provides redundant reports that are otherwise reported in the job logs.
- No more fallback testing environments; the user must be explicit about what they want to be run.
- Generic testing environment is renamed to `ci` instead of the previously used fallback commands.
- Changed `skip*Test` flag in groovy functions to `run*Test` and inverted conditionals.


.jenkins/common.groovy
```diff
@@ -122,71 +94,76 @@ def publishResults(project, boolean skipHostTest=false)
...
+            export HOME=/home/jenkins
+            export TENSILE_COMPILER=${compiler}
 
             gpuArch=`/opt/rocm/bin/rocm_agent_enumerator  | tail -n 1`
 
-            ${markSkipHostTest}pushd build
-            ${markSkipHostTest}./TensileTests ${markSkipExtendedTest} --gtest_output=xml:host_test_output.xml --gtest_color=yes
-            ${markSkipHostTest}HOST_ERR=\$?
-            ${markSkipHostTest}popd
+            if ! ${skipHostTest}; then
```
ellosel, line 128:
```suggestion
            if ${runHostTest}; then
```

samjwu (approved):

tox.ini
```diff
@@ -1,56 +1,48 @@
...
-# process to crash if a multiprocessing job has an exception. Fixed in 5.3.3.
+basepython = python3
 deps =
     -r{toxinidir}/requirements.txt
     joblib
-    pytest>=5.4.1
-    pytest-xdist>=1.32.0
     filelock
+    pytest>=7.2.0
+    pytest-xdist>=3.3.0
```
AlexBrownAMD, line 23: Looks like some requirements got updated versions, probably a good idea to update the dependancy list on the wiki as well

.jenkins/common.groovy
```diff
@@ -23,170 +23,102 @@
...
+            export PATH=/opt/rocm/bin:\${PATH}
             export TENSILE_COMPILER=${compiler}
-            tox -v --workdir /tmp/.tensile-tox -e ${pythonVersion} -- ${test_dir} -m "${test_marks}" --timing-file=\$(pwd)/timing-\$gpuArch.csv
-            PY_ERR=\$?
-            date
+            export GPU_ARCH=`/opt/rocm/bin/rocm_agent_enumerator  | tail -n 1`
+            export TIMING_FILE=`pwd`/timing-\${GPU_ARCH}.csv
+
+            tox --version
+            tox run -e ci -- -m ${testMark} --timing-file=\${TIMING_FILE} --numprocesses=4
```
AlexBrownAMD, line 101: Some of the process/thread code got re-arranged. Looks like it's all the same settings, but would be a good idea to double check the output on the CI run for this PR. If threading is too high, we get test failures due to the RAM on CI nodes being shared across multiple jobs. If the threading is too low, we sometimes hit the timeout limit (particularly for extended tests). Would be worth double checking that the test run-time is similar to previous CI jobs.

AlexBrownAMD (approved):

.jenkins/common.groovy
bstefanuk: I noticed this behaviour, originally, I set it to **auto** (which sets defaults to nproc) thinking that it would speed up our CI jobs but ran into the issues you're describing. I've updated the base numprocesses to 8 in *tox.ini*, so this will be used when running locally with `tox run -e ci ...`, but override that to 4 in CI, which it what it was set to before these changes---seems to be the sweet spot.

bstefanuk: I've noted some non-determinism with the setup I was using so I defaulted everything back to 4 processes (how it was before) and the tests now appear to be reliably passing.

PR #1937, opened 2024-06-14
Replace YAML operations with C libyaml backend

bstefanuk:
**Objectives**
- Reduce the runtime of Tensile applications and any downstream build chains by replacing the standard PyYAML Loader and Dumper with the C libyaml backend.

**Testing:**

*TensileCreateLibrary*

When timing the execution of *TensileCreateLibrary* on gfx90a architectures, the following differences were noted:
- Before (with SafeLoader): real   **6m26s**
- After (with CSafeLoader): real    **5m13s**
- Result: 19.0% runtime reduction

Command issued:
```bash
$ time Tensile/bin/TensileCreateLibrary ${rocblas_root}/library/src/blas3/Tensile/Logic/asm_full/aldebaran build_profiler HIP --merge-files --separate-architecture --lazy-library-loading --no-short-file-names --no-library-print-debug --code-object-version=default --cxx-compiler=hipcc --jobs=64 --library-format=msgpack --architecture=gfx90a
```

*rocBLAS*

When building *rocBLAS*, the following differences were noted:
- Before (with SafeLoader): real **53m46s**
- After (with CSafeLoader): real **51m57s**
- Result: 3.4% runtime reduction

Command issued:
```bash
$ ./install.sh
```

### DIFF
--- a/HostLibraryTests/configs/ConvertYAML.py
+++ b/HostLibraryTests/configs/ConvertYAML.py
@@ -55,7 +55,7 @@ def merge_libraries(args):
 
     with open(outFile, 'w') as outf:
         if True:
-            yaml.dump(outData, outf)
+            yaml.dump(outData, outf, yaml.CSafeDumper)
         else:
             import json
             json.dump(outData, outf, sort_keys=True, indent=2, separators=(",", ": "))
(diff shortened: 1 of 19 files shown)


AlexBrownAMD (approved):
Looks good. Just double check if this requires a new dependency that would need to get approved and documents.

PR #1938, opened 2024-06-17
Factors client config logic from TensileCreateLibrary main into createClientConfig

ellosel:
A portion of the TensileCreateLibrary related to creating a client config file main was factored into a function called `createClientConfig` to improve readability, testability, profilability and documentation. The function is now unit tested and documented. Further, when profiling TensileCreateLibrary, the time spent doing work in this function will be clearly associated with the function.

### DIFF
--- a/Tensile/TensileCreateLibrary.py
+++ b/Tensile/TensileCreateLibrary.py
@@ -1101,8 +1101,8 @@ def findLogicFiles(path: Path, logicArchs: Set[str], lazyLoading: bool, experime
     return list(str(l) for l in logicFiles)
 
 def createClientConfig(outputPath: Path, masterFile: Path, codeObjectFiles: List[str], configFile: str = "best-solution.ini") -> None:
-    """Outputs a client config file corresponding to a master library file and code-object parameters 
-       created by a TensileCreateLibrary invocation. Also sets best-solution-mode to True.
+    """Outputs a client config file corresponding to a master library file and code-object parameters
+           created by a TensileCreateLibrary invocation. Also sets best-solution-mode to True.
     
     Args:
         outputPath: The path to the tensile output directory where output files are written.
--- a/Tensile/TensileCreateLibrary.py
+++ b/Tensile/TensileCreateLibrary.py
@@ -1101,20 +1101,14 @@ def findLogicFiles(path: Path, logicArchs: Set[str], lazyLoading: bool, experime
     return list(str(l) for l in logicFiles)
 
 def createClientConfig(outputPath: Path, masterFile: Path, codeObjectFiles: List[str]) -> None:
-    """Outputs a client config file named best-solution.ini corresponding a master library 
+    """Outputs a client config file named best-solution.ini corresponding to a master library 
        file and code-object parameters created by a TensileCreateLibrary invocation. Also
        sets best-solution-mode to True.
     
     Args:
         outputPath: The path to the tensile output directory where output files are written.
-        masterFile: Path to the master library file (.dat or .yaml) e.g.:
-
-          "tensile-out/library/TensileLibrary_lazy_gfx908.dat"
-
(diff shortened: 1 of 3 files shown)


Tensile/TensileCreateLibrary.py
```diff
@@ -1099,6 +1100,41 @@ def findLogicFiles(path: Path, logicArchs: Set[str], lazyLoading: bool, experime
 
     return list(str(l) for l in logicFiles)
 
+def createClientConfig(outputPath: Path, masterFile: Path, codeObjectFiles: List[str]) -> None:
+    """Outputs a client config file named best-solution.ini corresponding a master library
```
bstefanuk, line 1104:
```suggestion
    """Outputs a client config file named best-solution.ini corresponding to a master library 
```

Tensile/TensileCreateLibrary.py
```diff
@@ -1099,6 +1100,35 @@ def findLogicFiles(path: Path, logicArchs: Set[str], lazyLoading: bool, experime
...
+    
+    Args:
+        outputPath: The path to the tensile output directory where output files are written.
+        masterFile: Path to the master library file (.dat or .yaml).
+        codeObjectFiles: List of code object files created by TensileCreateLibrary.
+    """
+    iniFile = outputPath / "best-solution.ini"
+    
+    def param(key, value):
+      f.write("{}={}\n".format(key, value))
```
bstefanuk, line 1116: Below you use `f"{...}"` f-string style formatting. Update here for consistency?

Tensile/TensileCreateLibrary.py
```diff
@@ -1099,6 +1100,35 @@ def findLogicFiles(path: Path, logicArchs: Set[str], lazyLoading: bool, experime
 
     return list(str(l) for l in logicFiles)
 
+def createClientConfig(outputPath: Path, masterFile: Path, codeObjectFiles: List[str]) -> None:
+    """Outputs a client config file named best-solution.ini corresponding to a master library 
+       file and code-object parameters created by a TensileCreateLibrary invocation. Also
+       sets best-solution-mode to True.
```
bstefanuk: I wonder if it would actually be better to refer to this file as a _manifest_ file, instead of a config file. Also, what is the "client" in this context, is it the library implementing Tensile?

Tensile/Tests/unit/test_TensileCreateLibrary.py
```diff
@@ -266,9 +266,56 @@ def verifyYamlAndYml():
...
+
+
+def cleanClientConfigTest(outputPath: Path, masterLibrary, codeObjectFiles, configFile):
+    with contextlib.suppress(FileNotFoundError):
+        os.remove(configFile)
+        os.remove(masterLibrary)        
+        for f in codeObjectFiles:
+            os.remove(outputPath / f)
+        next(outputPath.iterdir()).rmdir()
+        outputPath.rmdir()
```
bstefanuk, line 316: Why do have to get the next item in the iterator, remove it, and then remove the `outputDir`? Is L315 required?

Tensile/TensileCreateLibrary.py
ellosel: Config is correct here because it is used to configure the client app which is a stand-in for rocBLAS for testing purposes.

Tensile/Tests/unit/test_TensileCreateLibrary.py
ellosel: Removing the subdir from the parent dir before removing the parent dir. Path doesn't provide a nice way of doing this from what I could tell. I was trying to avoid pulling in another dependency to get this done cleaner.

bstefanuk: I see, this is for the **library** directory that you create, makes sense.

bstefanuk:
Looks good, approved.

bstefanuk (approved):

PR #1940, opened 2024-06-19
Removes deprecated package-library option

ellosel:
Based on the following comment from #1477:

> Some of this functionality is similar to the PackageLibrary option which I notice hasn't been updated in a while and is listed for possible deprecation. When using the PackageLibrary option, TensileCreateLibrary fails to build. I've put my changes alongside this option, although some of the functionality is the same.

and the fact that the `--package-library` option does not appear to work, this PR removes the option and all associated logic.


Tensile/TensileCreateLibrary.py
```diff
@@ -319,26 +315,11 @@ def buildSourceCodeObjectFile(CxxCompiler, outputPath, kernelFile):
...
-      destCosList += destCOs
-      for (src, dst) in zip(extractedCOs, destCOs):
-        shutil.copyfile(src, dst)
+    coFilenames = [name for name in coFilenames]
+    extractedCOs = [os.path.join(buildPath, name) for name in coFilenames]
+    destCosList = [os.path.join(destDir, name) for name in coFilenames]
+    for (src, dst) in zip(extractedCOs, destCosList):
+      shutil.copyfile(src, dst)
 
     return destCosList
```
bstefanuk, line 324: Inconsistent usage of code object acronym, `COs` vs `Cos`, not a huge deal, but I'm a fan of consistency.

bstefanuk (commented):
Straightforward PR, looks good. The only consideration I have is how existing YAML configs that contain "PackageLibrary" will respond to this change. My expectation is that it will be silent argument in these files. We may want to consider adding some sort of parameter deprecation warning in the future.

Tensile/TensileCreateLibrary.py
ellosel: This is showing up as new code but it isn't. I changed indentation. Still happy to update naming.

ellosel:
> Straightforward PR, looks good. The only consideration I have is how existing YAML configs that contain "PackageLibrary" will respond to this change. My expectation is that it will be a silent parameter in these files. We may want to consider adding some sort of parameter deprecation warning in the future.

Assuming that such a config file exists, that wouldn't be an unreasonable thing to do. I didn't find any config files in Tensile that use `PackageLibrary` as a key.

yoichiyoshida (approved):

bstefanuk (approved):

PR #1920, opened 2024-05-17
Move from hipcc to amdclang 

ellosel:
The goal of this change set is to move the default compiler from hipcc to amdclang++. The most significant difference between hipcc and amdclang are the flags used when invoking the compiler wrapped by hipcc. We attempt to delineate those differences below by noting if a flag is defaulted (d), set manually (m) or unset (blank):

 | Flag                                                        | hipcc    | amdclang | Notes |
| ------------------------------------------- | -------- |------------|-----------------------------------------|
| `-O3`                                                        | d | m | amdclang is `O2` by default                  |
| `-x hip`                                                     | d | m | Failure to compile if unset                     |
| `-D__HIP_HCC_COMPAT_MODE__=1`      | d | m | DecisionTree test segfaults when unset | 
| `-mrelax-all`                                             | d |      |    |
| `-mframe-pointer=all`                             | d |      |    |
| `-mframe-pointer=none`                        |    | d |    |
| `-Wno-format-nonliteral`                        | d |     |    | 
| `-fallow-half-arguments-and-returns`     | d |     |    |
| `-mllvm -amdgpu-early-inline-all=true`  | d |     |    |   
| `-mllvm -amdgpu-function-calls=false`   | d |     |    |
| `--genco` | m |     |    |
| `--cuda-device-only` | d | m |  This is set for hipcc when `genco` is passed  |

To reproduce one can do the following:

**Configure the HostTestLibrary directory**

```
# hipcc
cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_COMPILER=hipcc -DTensile_CPU_THREADS=32 -DTensile_ROOT=`pwd`/Tensile -S HostLibraryTests -B build

# amdclang++
cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_COMPILER=/opt/rocm/bin/amdclang++ -DTensile_CPU_THREADS=32 -DTensile_ROOT=`pwd`/Tensile -S HostLibraryTests -B build/host3 -DCMAKE_CXX_FLAGS="-D__HIP_HCC_COMPAT_MODE__=1"
```

**Run the build**

```
cmake --build build/host -j 32
```

**Remove DecisionTree test binaries, rebuild with verbosity and review the history for build command**

```
rm build/host/CMakeFiles/TensileTest.dir/DecisionTree_test.cpp.o
cmake --build build/host --verbose
```

**Run the build command from above with `-v`**

```
cd build/host
[...]

PR #1941, opened 2024-06-19
Update clang support for windows

ellosel:
This PR extends #1920 by using clang++ rather than hipcc on windows when amdclang++ is selected for the c++ compiler.

### DIFF
--- a/Tensile/ClientExecutable.py
+++ b/Tensile/ClientExecutable.py
@@ -27,7 +27,7 @@ import os
 import subprocess
 
 from . import Common
-from .Common import globalParameters, print2, supportedLinuxCompiler
+from .Common import globalParameters, print2, supportedCompiler
 from .Parallel import CPUThreadCount
 
 def cmake_path(os_path):
@@ -79,8 +79,8 @@ def clientExecutableEnvironment(builddir=None):
         builddir = os.path.join(globalParameters["OutputPath"], globalParameters["ClientBuildPath"])
     builddir = Common.ensurePath(builddir)
 
-    CxxCompiler = "clang++.exe" if ((os.name == "nt") and supportedLinuxCompiler(globalParameters['CxxCompiler'])) else globalParameters['CxxCompiler']
-    CCompiler   = "clang.exe"   if ((os.name == "nt") and supportedLinuxCompiler(globalParameters['CxxCompiler'])) else globalParameters['CCompiler']
+    CxxCompiler = "clang++.exe" if ((os.name == "nt") and supportedCompiler(globalParameters['CxxCompiler'])) else globalParameters['CxxCompiler']
+    CCompiler   = "clang.exe"   if ((os.name == "nt") and supportedCompiler(globalParameters['CxxCompiler'])) else globalParameters['CCompiler']
 
     options = {'CMAKE_BUILD_TYPE': globalParameters["CMakeBuildType"],
                'TENSILE_USE_MSGPACK': 'ON',
--- a/Tensile/ClientExecutable.py
+++ b/Tensile/ClientExecutable.py
@@ -80,7 +80,7 @@ def clientExecutableEnvironment(builddir=None):
     builddir = Common.ensurePath(builddir)
 
     CxxCompiler = "clang++.exe" if ((os.name == "nt") and supportedCompiler(globalParameters['CxxCompiler'])) else globalParameters['CxxCompiler']
-    CCompiler   = "clang.exe"   if ((os.name == "nt") and supportedCompiler(globalParameters['CxxCompiler'])) else globalParameters['Cxxompiler']
+    CCompiler   = "clang.exe"   if ((os.name == "nt") and supportedCompiler(globalParameters['CxxCompiler'])) else globalParameters['CxxCompiler']
 
     options = {'CMAKE_BUILD_TYPE': globalParameters["CMakeBuildType"],
                'TENSILE_USE_MSGPACK': 'ON',
--- a/Tensile/ClientExecutable.py
+++ b/Tensile/ClientExecutable.py
@@ -80,7 +80,7 @@ def clientExecutableEnvironment(builddir=None):
     builddir = Common.ensurePath(builddir)
 
     CxxCompiler = "clang++.exe" if ((os.name == "nt") and supportedCompiler(globalParameters['CxxCompiler'])) else globalParameters['CxxCompiler']
-    CCompiler   = "clang.exe"   if ((os.name == "nt") and supportedCompiler(globalParameters['CxxCompiler'])) else globalParameters['CCompiler']
+    CCompiler   = "clang.exe"   if ((os.name == "nt") and supportedCompiler(globalParameters['CxxCompiler'])) else globalParameters['Cxxompiler']
 
     options = {'CMAKE_BUILD_TYPE': globalParameters["CMakeBuildType"],
                'TENSILE_USE_MSGPACK': 'ON',
--- a/Tensile/ClientExecutable.py
+++ b/Tensile/ClientExecutable.py
@@ -80,7 +80,7 @@ def clientExecutableEnvironment(builddir=None):
     builddir = Common.ensurePath(builddir)
 
     CxxCompiler = "clang++.exe" if ((os.name == "nt") and supportedCompiler(globalParameters['CxxCompiler'])) else globalParameters['CxxCompiler']
-    CCompiler   = "clang.exe"   if ((os.name == "nt") and supportedCompiler(globalParameters['CxxCompiler'])) else globalParameters['CxxCompiler']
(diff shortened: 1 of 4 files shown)


Tensile/TensileCreateLibrary.py
```diff
@@ -1183,7 +1187,7 @@ def splitExtraParameters(par):
   arguments["Architecture"] = args.Architecture
   arguments["SeparateArchitectures"] = args.SeparateArchitectures
   arguments["LazyLibraryLoading"] = args.LazyLibraryLoading
-  arguments["CxxCompiler"] = args.CxxCompiler
+  arguments["CxxCompiler"] = CxxCompiler
```
TorreZuk, line 1190: You corrected args so not sure why you would not trust it here but use it later,  maintenance problem

Tensile/Common.py
```diff
@@ -2377,11 +2376,11 @@ def assignGlobalParameters( config ):
...
   # This may not be sufficient for amdclang
   try:
+    compiler = "hipcc"
     if os.name == "nt":
-      compileArgs = ['perl'] + [which('hipcc')] + ['--version']
+      compileArgs = ['perl'] + [which(compiler)] + ['--version']
       output = subprocess.run(compileArgs, check=True, stdout=subprocess.PIPE).stdout.decode()
     else:
-      compiler = "hipcc"
       output = subprocess.run([compiler, "--version"], check=True, stdout=subprocess.PIPE).stdout.decode()
```
TorreZuk, line 2384: All this should have been removed for linux hipcc -> amdclang and windows.  Ask hip team if hipconfig is stable. Use it for version instead of hipcc.  It appears to support --version on windows and linux.  You may need backward compatible fallbacks.

Tensile/Common.py
```diff
@@ -330,16 +330,15 @@ def getArchitectureName(gfxName: str) -> Optional[str]:
...
       Args:
           The name of a compiler to test for support.
       
       Return:
           If supported True; otherwise, False.
   """
-  if (compiler == "hipcc" or compiler == "amdclang++"): return True
-
-  return False
+  return (compiler == "hipcc" or compiler == "amdclang++" or compiler == "clang++")
```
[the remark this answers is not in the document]
ellosel: I agree and I am changing it now.

Tensile/TensileCreateLibrary.py
```diff
@@ -1173,6 +1173,10 @@ def splitExtraParameters(par):
   logicPath = args.LogicPath
   outputPath = args.OutputPath
   CxxCompiler = args.CxxCompiler
+  if os.name == "nt" and CxxCompiler == "amdclang++":
+    CxxCompiler = "clang++"
+    args.CxxCompiler = "clang++"
+    printWarning("On windows using clang++.")
```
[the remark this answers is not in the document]
ellosel: Glad you brought this up. I was considering adding clang++ as another cxx-compiler option but I wasn't sure how the rocBLAS build system would deal with it on windows. I personally would like to add clang++ as an option and drop this logic altogether.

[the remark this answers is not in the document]
TorreZuk: That is different issue allowing other compilers, a clang++ generic on linux we don't support as don't test it.  The amdclang is just to clarify it is not generic clang.    IMO it is better to just support what we test with.  Windows may package amdclang name soon so then both can be clear this is not a generic clang, but for now we use it.

Tensile/Common.py
ellosel: We have a ticket that will likely remove this altogether. We compute the version for use later when conditionally setting flags based on the compiler version. However, we no longer need to support the older version which uses different flags than the newer, supported compiler versions.

Either way I will look into using hipconfig for the time being.

Tensile/Common.py
```diff
@@ -329,17 +329,24 @@ def getArchitectureName(gfxName: str) -> Optional[str]:
...
-  if (compiler == "hipcc" or compiler == "amdclang++"): return True
-
-  return False
+  isSupported = (compiler == "hipcc")
+  if os.name == "nt": 
+    isSupported = (isSupported or compiler == "clang++")
+  else:
+    isSupported = (isSupported or compiler == "amdclang++")
+  
+  printWarning(f"{compiler} is unsupported for os {os.name}")
```
TorreZuk, line 347: This seems like it is printing false warning

Tensile/Common.py
```diff
@@ -2381,11 +2388,11 @@ def assignGlobalParameters( config ):
   
   # This may not be sufficient for amdclang
   try:
+    compiler = "hipcc"
```
TorreZuk, line 2391: You could comment that hipcc is used only for hip version information

Tensile/Common.py
```diff
@@ -2335,7 +2342,7 @@ def assignGlobalParameters( config ):
 
   if "TENSILE_ROCM_ASSEMBLER_PATH" in os.environ:
     globalParameters["AssemblerPath"] = os.environ.get("TENSILE_ROCM_ASSEMBLER_PATH")
-  elif globalParameters["AssemblerPath"] is None and supportedLinuxCompiler(globalParameters["CxxCompiler"]):
+  elif globalParameters["AssemblerPath"] is None and supportedCompiler(globalParameters["CxxCompiler"]):
     if os.name == "nt":
```
TorreZuk, line 2346: Not sure we even care if "and supportedCompiler" at this conditional, but fine

Tensile/Common.py
```diff
@@ -2268,7 +2275,7 @@ def capRow(caps, cap):
 def which(p):
     exes = [p+x for x in ['', '.exe', '.bat']]
     system_path = os.environ['PATH'].split(os.pathsep)
-    if supportedLinuxCompiler(p) and 'CMAKE_CXX_COMPILER' in os.environ and os.path.isfile(os.environ['CMAKE_CXX_COMPILER']):
+    if supportedCompiler(p) and 'CMAKE_CXX_COMPILER' in os.environ and os.path.isfile(os.environ['CMAKE_CXX_COMPILER']):
         return os.environ['CMAKE_CXX_COMPILER']
```
TorreZuk, line 2279: Could move this to first as quick return but fine

Tensile/TensileCreateLibrary.py
```diff
@@ -165,7 +165,7 @@ def which(p):
     else:
         exes = [p+x for x in ['', '.exe', '.bat']]
     system_path = os.environ['PATH'].split(os.pathsep)
-    if supportedLinuxCompiler(p) and 'CMAKE_CXX_COMPILER' in os.environ and os.path.isfile(os.environ['CMAKE_CXX_COMPILER']):
+    if supportedCompiler(p) and 'CMAKE_CXX_COMPILER' in os.environ and os.path.isfile(os.environ['CMAKE_CXX_COMPILER']):
         return os.environ['CMAKE_CXX_COMPILER']
```
TorreZuk, line 169: Same comment as in earlier duplicate code, but comparing the which( p ) I see this version has fix so on windows the .bat file is considered first (as simpler to use) but on the other version it considers versions without suffix.  Would think fixing one in Common.py and reusing by import if possible.  Or just copy the exes if from this one.    Strangely this is not which function if it allows a CXX compiler only env return which you are asking for which assembler.

Tensile/ClientExecutable.py
```diff
@@ -79,8 +79,8 @@ def clientExecutableEnvironment(builddir=None):
         builddir = os.path.join(globalParameters["OutputPath"], globalParameters["ClientBuildPath"])
     builddir = Common.ensurePath(builddir)
 
-    CxxCompiler = "clang++.exe" if ((os.name == "nt") and supportedLinuxCompiler(globalParameters['CxxCompiler'])) else globalParameters['CxxCompiler']
-    CCompiler   = "clang.exe"   if ((os.name == "nt") and supportedLinuxCompiler(globalParameters['CxxCompiler'])) else globalParameters['CCompiler']
+    CxxCompiler = "clang++.exe" if ((os.name == "nt") and supportedCompiler(globalParameters['CxxCompiler'])) else globalParameters['CxxCompiler']
+    CCompiler   = "clang.exe"   if ((os.name == "nt") and supportedCompiler(globalParameters['CxxCompiler'])) else globalParameters['CCompiler']
```
TorreZuk, line 83: Think the second "and supportedCompiler(globalParameters['CxxCompiler']) "should be C compiler 'CCompiler'.   Using the global var as a test for support but then ignoring it and using the hard coded one seems strange.

Tensile/Common.py
ellosel: I believe some of this logic i.e. and supportedCompiler is related to supporting opencl. We can create another ticket to investigate and remove.

yoichiyoshida (approved):

PR #1947, opened 2024-06-24
Add functions for writing master files 

ellosel:
This PR factors out logic from TensileCreateLibrary related to writing master files to disk into functions that are testable, profitable and documented.

### DIFF
--- a/Tensile/TensileCreateLibrary.py
+++ b/Tensile/TensileCreateLibrary.py
@@ -1139,7 +1139,7 @@ def generateMasterFileList(masterLibraries: dict, archs: List[str], lazy: bool) 
     result = [(baseName + arch, masterLibrary) for arch, masterLibrary in masterLibraries.items() if arch in archs]
     return generateLazyMasterFileList(result) if lazy else result
 
-def writeMasterFile(libraryPath: Path, name: str, lib: MasterSolutionLibrary, format: str, naming: dict):
+def writeMasterFile(libraryPath: Path, format: str, naming: dict, name: str, lib: MasterSolutionLibrary):
     """ Writes a master file to disk as a .yaml or .dat file.
 
     Args:
@@ -1367,10 +1367,12 @@ def TensileCreateLibrary():
              if globalParameters["AsmCaps"][arch]["SupportedISA"]]
 
   newLibraryDir = Path(outputPath) / "library"
+  newLibraryDir.mkdir(exist_ok=True)
 
   masterFileList = generateMasterFileList(masterLibraries, archs, args.LazyLibraryLoading) if args.SeparateArchitectures \
     else [("TensileLibrary", fullMasterLibrary)]
-  map(partial(writeMasterFile, outputPath=newLibraryDir, format=args.LibraryFormat, naming=kernelMinNaming), masterFileList)
+  for name, lib in masterFileList:
+    writeMasterFile(newLibraryDir, args.LibraryFormat, kernelMinNaming, name, lib)
   masterFile, fullMasterLibrary = masterFileList[0]
   
   ext = ".yaml" if globalParameters["LibraryFormat"] == "yaml" else ".dat"
--- a/Tensile/TensileCreateLibrary.py
+++ b/Tensile/TensileCreateLibrary.py
@@ -1110,11 +1110,11 @@ def generateClientConfig(outputPath: Path, masterFile: Path, codeObjectFiles: Li
       param("best-solution", True)
 
 def generateLazyMasterFileList(masterFileList: List[Tuple[str, MasterSolutionLibrary]]) -> List[Tuple[str, MasterSolutionLibrary]]:
-    """ Generates a list of tuples that represent the name and the state associated the lazy master libraries.
+    """ Generates a list of tuples that represent the name and the state associated with the lazy master libraries.
 
-    This function takes a list of MasterSolutionLibraries and traverses its lazy libraries.
+    This function takes a list of MasterSolutionLibraries and traverses each lazy libraries.
     It collects the items (i.e. the name and corresponding master file) and adds them to list
-    of master files to be written to disk.
+    of master files.
 
     Args:
         masterLibraries: A list of name / master solution library pairs.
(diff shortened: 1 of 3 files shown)


Tensile/TensileCreateLibrary.py
```diff
@@ -1109,6 +1109,47 @@ def param(key, value):
       
       param("best-solution", True)
 
+def generateLazyMasterFileList(masterFileList: List[Tuple[str, MasterSolutionLibrary]]) -> List[Tuple[str, MasterSolutionLibrary]]:
+    """ Generates a list of tuples that represent the name and the state associated the lazy master libraries.
```
ellosel, line 1113:
```suggestion
    """ Generates a list of tuples that represent the name and the state associated with  the lazy master libraries.
```

Tensile/TensileCreateLibrary.py
```diff
@@ -1109,6 +1109,47 @@ def param(key, value):
...
+
+    This function takes a list of MasterSolutionLibraries and traverses its lazy libraries.
+    It collects the items (i.e. the name and corresponding master file) and adds them to list
+    of master files to be written to disk.
```
ellosel, line 1117:
```suggestion
    of master files.
```

Tensile/TensileCreateLibrary.py
```diff
@@ -1109,6 +1109,47 @@ def param(key, value):
...
+        masterLibraries: A list of name / master solution library pairs.
+    """
+    return masterFileList + list(itertools.chain.from_iterable((t for t in lib.lazyLibraries.items()) for _, lib in masterFileList))
+
+def generateMasterFileList(masterLibraries: dict, archs: List[str], lazy: bool) -> List[Tuple[str, MasterSolutionLibrary]]:
+    """ Generates a list of tuples that represent the name and the state associated the architecture
+        specific master libraries.
+
+    This function takes a dictionary with keys corresponding to a target architecture and values 
+    corresponding to the master solution library for that architecture. The fucntion generates
```
ellosel, line 1129:
```suggestion
    corresponding to the master solution library for that architecture. The function generates a
```

Tensile/TensileCreateLibrary.py
```diff
@@ -1109,6 +1109,47 @@ def param(key, value):
...
+
+def writeMasterFile(libraryPath: Path, format: str, naming: dict, name: str, lib: MasterSolutionLibrary):
+    """ Writes a master file to disk as a .yaml or .dat file.
+
+    Args:
+        libraryPath: Path to library subdirectory located in the tensile output directory.
+        name: Name of the masterfile.
+        lib: Master solution library data.
+        format: Output format of written file (.dat or .yaml).
+        naming: Kernel minimum naming.
```
ellosel, line 1149:
```suggestion
        libraryPath: Path to library subdirectory located in the tensile output directory.
        format: Output format of written file (.dat or .yaml).
        naming: Kernel minimum naming.
        name: Name of the masterfile.
        lib: Master solution library data.
```

Tensile/TensileCreateLibrary.py
```diff
@@ -1109,6 +1109,47 @@ def param(key, value):
...
+    """ Generates a list of tuples that represent the name and the state associated with the lazy master libraries.
+
+    This function takes a list of MasterSolutionLibraries and traverses each lazy libraries.
+    It collects the items (i.e. the name and corresponding master file) and adds them to list
+    of master files.
+
+    Args:
+        masterLibraries: A list of name / master solution library pairs.
+    """
+    return masterFileList + list(itertools.chain.from_iterable((t for t in lib.lazyLibraries.items()) for _, lib in masterFileList))
```
bstefanuk, line 1122: Since you're immediately casting to a list, I don't see the value of using a generator here, wouldn't this be simpler without the call to `itertools.chain.from_iterable` and equally performant?

Tensile/TensileCreateLibrary.py
```diff
@@ -1109,6 +1109,47 @@ def param(key, value):
...
 
+def generateLazyMasterFileList(masterFileList: List[Tuple[str, MasterSolutionLibrary]]) -> List[Tuple[str, MasterSolutionLibrary]]:
+    """ Generates a list of tuples that represent the name and the state associated with the lazy master libraries.
+
+    This function takes a list of MasterSolutionLibraries and traverses each lazy libraries.
+    It collects the items (i.e. the name and corresponding master file) and adds them to list
+    of master files.
+
+    Args:
+        masterLibraries: A list of name / master solution library pairs.
```
bstefanuk, line 1120: Return type isn't specified.

Tensile/TensileCreateLibrary.py
```diff
@@ -1320,46 +1361,30 @@ def splitExtraParameters(par):
...
+  newLibraryDir.mkdir(exist_ok=True)
 
-  theMasterLibrary = fullMasterLibrary
-  if globalParameters["SeparateArchitectures"]:
-    theMasterLibrary = list(masterLibraries.values())[0]
+  masterFileList = generateMasterFileList(masterLibraries, archs, args.LazyLibraryLoading) if args.SeparateArchitectures \
+    else [("TensileLibrary", fullMasterLibrary)]
+  for name, lib in masterFileList:
+    writeMasterFile(newLibraryDir, args.LibraryFormat, kernelMinNaming, name, lib)
+  masterFile, fullMasterLibrary = masterFileList[0]
```
bstefanuk, line 1375: Has the consistency with previous implementation been tested for the various combinations of separate architectures, lazy loading, etc. to ensure the output is unchanged when using this later in embed library?

Tensile/TensileCreateLibrary.py
```diff
@@ -1109,6 +1109,47 @@ def param(key, value):
...
+    tuple consisting of a MasterSolutionLibrary and the associated name.
+
+    Args:
+        masterLibraries: A dictionary of architecture name / master solution library pairs.
+        archs: A list of supported architectures.
+        lazy: If True, add lazy library master files.
+    """
+    baseName = "TensileLibrary_lazy_" if lazy else "TensileLibrary_"
+    result = [(baseName + arch, masterLibrary) for arch, masterLibrary in masterLibraries.items() if arch in archs]
+    return generateLazyMasterFileList(result) if lazy else result
```
bstefanuk, line 1139: Not opposed to this, but the readability would be better IMO if it read like 
```python
    return result + generateLazyMasterFileList(result) if lazy else result
```
and the extension of the existing masterFileList isn't subsumed within the function call, thoughts?

Tensile/TensileCreateLibrary.py
ellosel: If we use list comprehension, we wind up with a list of lists and would still require a call to `itertools.chain.from_iterable`. I suspect the performance difference is negligible.

Tensile/TensileCreateLibrary.py
ellosel: I am good with that change

Tensile/TensileCreateLibrary.py
bstefanuk: Here's another option
```python
return [t for _, lib in masterFileList for t in lib.items()]
```

bstefanuk (approved):

PR #1952, opened 2024-06-26
Format TensileCreateLibrary

bstefanuk:
Formatting helps support smaller diffs in pull requests. This PR applies formatting for TensileCreateLibrary and will allow future changes to be made without having to format small sections.

Formatter used: [black](https://black.readthedocs.io/en/stable/index.html) with out-of-the-box configuration.

**Outcomes**
- Format job is added to tox and callable with `tox run -e format`
- Formatting is to be conducted *before pushing*. In CI, the formatter is called with the `--check` flag, which will cause a pipeline failure if incorrect formatting is encountered.
- So as to begin integrating the formatter without introducing changes to the whole code base, only _TensileCreateLibrary_ is formatted via tox. This will be updated in future iterations.

### DIFF
--- a/.jenkins/staticanalysis.groovy
+++ b/.jenkins/staticanalysis.groovy
@@ -47,23 +47,14 @@ def runCompileCommand(platform, project, jobName, boolean debug=false)
 
     def command = """#!/usr/bin/env bash
             set -ex
-
             hostname
-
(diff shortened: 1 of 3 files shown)


ellosel (approved):
Looks good!

PR #1953, opened 2024-06-26
Disable failing rhel9 tests

ellosel:
The rhel9 tests are failing on develop post-merge due to a link error when building the client via:

```
tox  run -e ci -- -m unit
```

and also when building the HostLibraryTests. Here we temporarily disable this testing for rhel9 until a permanent fix is identified.

### DIFF
--- a/.jenkins/common.groovy
+++ b/.jenkins/common.groovy
@@ -117,7 +117,8 @@ def runTestCommand(platform, project, jobName, testMark, boolean runHostTest=tru
         """
     platform.runCommand(this, command)
 
-    archiveArtifacts "${project.paths.project_build_prefix}/timing*.csv"
+    if (platform.os != "rhel9")        
+        archiveArtifacts "${project.paths.project_build_prefix}/timing*.csv"
     if (runUnitTest) {
         recordCoverage(tools: [[parser: 'COBERTURA', pattern: "${project.paths.project_build_prefix}/cobertura.xml"]])
     }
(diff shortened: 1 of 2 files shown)


.jenkins/common.groovy
```diff
@@ -97,21 +97,23 @@ def runTestCommand(platform, project, jobName, testMark, boolean runHostTest=tru
...
             if ${runUnitTest}; then 
               tox run -e unittest -- --cov-report=xml:cobertura.xml
               check_err
             fi
-
-            if ${runHostTest}; then
-              pushd build
-              ./TensileTests ${markSkipExtendedTest} --gtest_color=yes
+            echo The operating system is ${platform.os}
+            if [ ${platform.os} != "rhel9" ]
```
bstefanuk, line 105: Does this syntax work? When I write bash conditionals like this I typically use `if [...]; then`

TorreZuk:
Are the tests not applicable to rhel9 ?  Not clear why you are disabling them.....  if you want to ignore them temporarily for purposes of promotion you can just fast fwd develop to staging manually.

ellosel:
> Are the tests not applicable to rhel9 ? Not clear why you are disabling them..... if you want to ignore them temporarily for purposes of promotion you can just fast fwd develop to staging manually.

We are disabling everything except for the unit tests. Seems like it would be better to have some minimal testing in place.

yoichiyoshida (approved):

PR #1956, opened 2024-06-28
Fix comments on scalarStaticDivideAndRemainder

AlexBrownAMD:
Related to previous change in PR #1928 .
scalarStaticDivideAndRemainder requires 2 temp registers, previous change incorrectly updated doc comments. Revert comments and add assert to ensure temp register is passed when needed.


nakajee (approved):
Thanks for your update. Looks good.

PR #1957, opened 2024-06-28
Refactor generateLogicDataAndSolutions

ellosel:
The goal of this PR is to refactor the logic associated with the `generateLogicDataAndSolutions` into smaller testable functions with documentation. This will enable finer grained profile results that will help identify precisely where performance bottlenecks occur. Further, by testing the new functions, code coverage will improve and provide stability in the presence of future changes.

In addition to refactoring the function, we eliminated the need to carry multiple master solution libraries (i.e. `masterLibraries` and `fullMasterLibrary`) forward in `TensileCreateLibrary` which simplifies logic in downstream code.


Tensile/Tests/test_data/unit/solutions/hip_Cijk_Ailk_Bjlk_CB_GB.yaml
ellosel, line 1: Not sure if we need both files.

Tensile/Tests/unit/test_TensileCreateLibrary.py
```diff
@@ -328,6 +330,7 @@ def __init__(self, libraries, data):
         self.lazyLibraries = libraries
         self.data = data
 
+
```
ellosel, line 333: remove

Tensile/Tests/unit/test_TensileCreateLibrary.py
```diff
@@ -352,9 +355,52 @@ def test_generateMasterFileList():
...
 # ----------------
 # Helper functions
 # ----------------
+class Given:
+     def __init__(self, description: str):
+         self.description = description
+     def __enter__(self):
+         return self.desc
+     def __exit__(self, exc_type, exc_value, exc_tb):
+         print(exc_type, exc_value, exc_tb)
```
ellosel, line 396: remove

Tensile/Tests/conftest.py
```diff
@@ -157,3 +157,20 @@ def __exit__(self, exc_type, exc_value, traceback):
 
     return gpUpdater
 
+@pytest.fixture
+def initGlobalParametersForTCL():
+    from Tensile import Common
+    from Tensile.TensileCreateLib.ParseArguments import parseArguments
```
bstefanuk, line 163: Move these to top-level imports?

Tensile/Tests/unit/test_TensileCreateLibrary.py
```diff
@@ -352,9 +353,44 @@ def test_generateMasterFileList():
...
 
+
+def test_componentsOfLogicDataAndSolutionsConstruction(initGlobalParametersForTCL):
+    
+    requiredArgs = ["--jobs=2", "/unused/logic/path", "/unused/output/path", "HIP"]
+    
+    with initGlobalParametersForTCL(["--architecture=gfx900"]+requiredArgs):
+        for separateArch in [False, True]:
+            rootPath = Path(__file__).parent/"../test_data"/"unit"/"solutions"
+            yamlFiles = [rootPath/f for f in["vega10_Cijk_Ailk_Bjlk_CB_GB.yaml", "hip_Cijk_Ailk_Bjlk_CB_GB.yaml"]]
```
bstefanuk, line 364: I've implemented similar logic for reading a yaml file in another PR. I opted to use the pytest [request](https://docs.pytest.org/en/latest/reference/reference.html#request) fixture, provides the same end result. We should chat about how we want to maintain testing resource files such as these in a consistent manner.

Tensile/Tests/unit/test_TensileCreateLibrary.py
```diff
@@ -352,9 +353,44 @@ def test_generateMasterFileList():
...
     assert t[1].data == (idx+2), "Incorrect data."        
 
+
+def test_componentsOfLogicDataAndSolutionsConstruction(initGlobalParametersForTCL):
+    
+    requiredArgs = ["--jobs=2", "/unused/logic/path", "/unused/output/path", "HIP"]
+    
+    with initGlobalParametersForTCL(["--architecture=gfx900"]+requiredArgs):
+        for separateArch in [False, True]:
+            rootPath = Path(__file__).parent/"../test_data"/"unit"/"solutions"
```
bstefanuk, line 363:
```suggestion
            rootPath = Path(__file__).parent.parent/"test_data"/"unit"/"solutions"
```
Otherwise, when testing on Windows, it won't have the correct path separators. Alternatively, I believe `.parents[2]` will do the trick.

Tensile/TensileCreateLib/ParseArguments.py
```diff
@@ -46,8 +46,15 @@ def splitExtraParameters(par):
...
-    """Parse command line arguments for TensileCreateLibrary."""
+def parseArguments(input=None) -> Dict[str, Any]:
+    """Parse command line arguments for TensileCreateLibrary.
+
+    Args:
+        input: List of strings representing command line arguments.
+
+    Returns:
+        A dictionary containing the keys representing options and their values.
+    """
```
bstefanuk, line 57:
```suggestion
def parseArguments(input: List[str] | None=None) -> Dict[str, Any]:
    """Parse command line arguments for TensileCreateLibrary.

    Args:
        input: List of strings representing command line arguments. Only used during testing.

    Returns:
        A dictionary containing the keys representing options and their values.
    """
```

~~Since this isn't updated in TensileCreateLibrary main, what's the purpose of adding the `input` parameter?~~

I like this pattern after reading below and seeing it in use. However, it may be good to clarify in the doc-string that this parameter is only used when running tests.

Tensile/SolutionLibrary.py
```diff
@@ -500,33 +509,38 @@ def applyNaming(self, naming=None):
         for s in list(self.solutions.values()):
             s.name = OriginalSolution.getNameMin(s.originalSolution.getKernels()[0], naming)
 
-    def remapSolutionIndicesStartingFrom(self, curIndex):
+    def remapSolutionIndicesStartingFromImpl(self, library, solutions: dict, curIndex: int) -> tuple:
```
bstefanuk, line 512:
```suggestion
    def _remapSolutionIndicesStartingFromImpl(self, library, solutions: dict, curIndex: int) -> tuple:
```
Add underscore to identify it as a private method?

bstefanuk (commented):
Looking good, some minor comments and discussion points.

Tensile/Tests/unit/test_TensileCreateLibrary.py
ellosel: Reconcile in when we merge two different approaches?

Tensile/Tests/conftest.py
ellosel: I was keeping with convention for the file. Do you have a preference?

Tensile/Tests/unit/test_TensileCreateLibrary.py
ellosel: Good catch fixed.

Tensile/TensileCreateLib/ParseArguments.py
ellosel: Added.

bstefanuk (approved):
Looks good

Tensile/Tests/conftest.py
bstefanuk: Top-level as the preference makes sense to me, ensures import errors are thrown early, and makes it easy to see at a glance how complex the file is. However, as a new function, I don't see an issue keeping it there until it becomes more permanent fixture.

PR #1959, opened 2024-07-03
Restrict XCC mapping to gfx942

AlexBrownAMD:
The feature currently only applies to gfx942 and is untested on older architectures. Add reject case for other architectures for now until the feature is fully tested on those devices.

### DIFF
--- a/Tensile/SolutionStructs.py
+++ b/Tensile/SolutionStructs.py
@@ -2982,6 +2982,9 @@ class Solution(collections.abc.Mapping):
         reject(state, "Cannot enable both Stream-K and PersistentKernel")
       if not state["ProblemType"]["StridedBatched"]:
         reject(state, "General batch not supported with Stream-K")
+      if state["StreamKXCCMapping"] > 0:
+        if isa != [9,4,2]:
+          reject(state, "XCC mapping currently only on gfx942")
       if state["StreamKAtomic"] == 1:
         if not state["ProblemType"]["DataType"].isSingle():
           reject(state, "Atomic Stream-K currently only tested for SGEMM")
--- a/Tensile/SolutionStructs.py
+++ b/Tensile/SolutionStructs.py
@@ -2983,7 +2983,7 @@ class Solution(collections.abc.Mapping):
       if not state["ProblemType"]["StridedBatched"]:
         reject(state, "General batch not supported with Stream-K")
       if state["StreamKXCCMapping"] > 0:
-        if isa != [9,4,2]:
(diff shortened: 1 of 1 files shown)


nakajee:
Does sk test pass on 90a with this change?
I thought it would generate an error if no kernel was generated.

TorreZuk:
I think you were trying to fix extended, right?  So I added the label.

ellosel:
> Does sk test pass on 90a with this change? I thought it would generate an error if no kernel was generated.

90a tests are in progress.

nakajee:
I am thinking it might better to just skip the sk test cases for now.
We need to change the code again when we have new architectures.
We can put some comment in common.py to let users know this is verified on gfx942 only.

AlexBrownAMD:
> Does sk test pass on 90a with this change? I thought it would generate an error if no kernel was generated.

Yes it passes on 90a with this change. The test files you mentioned include both test cases for xccmapping=0 and 8. Tests with xccmapping enabled hang on gfx90a. Though it doesn't make sense to use the feature on 90a, it's possible the hang could also happen in other configurations as well so I've created a ticket to debug the issue. For now, the feature works in standard use-cases for 942 so this change restricts it to that until more testing is done.

ellosel (approved):
This looks good to me assuming tests pass.

nakajee (approved):
I am OK with this change for now (as a workaround).
It might be good to put some TODO comment (and restriction in Common.py).

bstefanuk:
Looks good, makes sense.

+1 Koji's comment, a config-level system to filter capabilities (e.g. Stream-K) for different architectures would be nice. It also speaks to our goal of extending support to a wider set of architectures.

bstefanuk (approved):

AlexBrownAMD:
> Looks good, makes sense.
> 
> +1 Koji's comment, a config-level system to filter capabilities (e.g. Stream-K) for different architectures would be nice. It also speaks to our goal of extending support to a wider set of architectures.

There is a system for this already. We can filter features based on the "AsmCaps" which poll for capabilities available on a given compiler+architecture. This system is preferred over checking for a specific architecture number, because sometimes features are not necessarily architecture specific but based on rocm versions. Also some architectures can have multiple variants with different capabilities.

For this change, the goal is to have all stream-k features working on all hardware, so this change is temporary (and hopefully to be improved soon). Also, I couldn't think of a specific asm cap that related to this feature in a way that made sense, so just wrote it to check the ISA for now.

PR #1962, opened 2024-07-08
Remove globals from prepAsm

bstefanuk:
Removes globals from `prepAsm` in *TensileCreateLibrary* and adds type hinting for `writeKernels`


Tensile/BenchmarkProblems.py
```diff
@@ -150,7 +150,7 @@ def writeBenchmarkFiles(stepBaseDir, solutions, problemSizes, \
     problemType = solutions[0]["ProblemType"]
     codeObjectFiles = writeKernels( \
             globalParameters["WorkingPath"], globalParameters["CxxCompiler"], \
-            [problemType], solutions, kernels, kernelHelperOjbs, \
+            globalParameters, solutions, kernels, kernelHelperOjbs, \
```
ellosel, line 153: Are we passing in the global params now?

Tensile/TensileCreateLibrary.py
```diff
@@ -396,69 +397,78 @@ def buildSourceCodeObjectFiles(CxxCompiler, kernelFiles, outputPath):
 
 
 ################################################################################
-def prepAsm(kernelWriterAssembly):
-    """
-    Create and prepare the assembly directory  - called ONCE per output dir:
+def prepAsm(kernelWriterAssembly: KernelWriterAssembly, buildPath: Path):
```
ellosel, line 400: Should we test this function as a part of this PR?

Tensile/TensileCreateLibrary.py
```diff
@@ -396,69 +397,78 @@ def buildSourceCodeObjectFiles(CxxCompiler, kernelFiles, outputPath):
...
+
+    Args:
+        kernelWriterAssembly: Assembly writer object.
+        buildPath: Path to directory where assembly files will be written.
     """
-    asmPath = ensurePath(os.path.join(globalParameters["WorkingPath"], "assembly"))
+    asmPath = buildPath / "assembly"
+    asmPath.mkdir(exist_ok=True)
+
     isa = globalParameters["CurrentISA"]
```
ellosel, line 414: What about this global?

Tensile/TensileCreateLibrary.py
bstefanuk: Yes, I will write some tests and add them.

Tensile/BenchmarkProblems.py
bstefanuk: The previous parameter `List[problemType]` was unused in the function. I replaced it with _params_, which, in TensileCreateLibrary main is just the _args_ dictionary. Elsewhere in the code, I replaced it with global parameters to appease the tests. This is primarily a stop-gap while we continue to migrate away from dependence on global parameters.

Tensile/TensileCreateLibrary.py
bstefanuk: Good catch, was focused on the "WorkingPath" parameter and had blinders on.

Tensile/BenchmarkProblems.py
bstefanuk: Alternatively, I remove this parameter altogether and we rely on the global parameters in **writeKernels**

ellosel (approved):

PR #1963, opened 2024-07-09
Refactor kernel duplicate matching

bstefanuk:
Simple refactor in *TensileCreateLibrary*, creating a function to enclose kernel matching logic.

**Testing**

Unit tests included

**Notable changes**

Raise a DeveloperWarning instead of normal printing when duplicates are encountered.


Tensile/TensileCreateLibrary.py
```diff
@@ -561,6 +561,42 @@ def buildKernelSourceAndHeaderFiles(results, outputPath, kernelsWithBuildErrs):
...
+    # Kernels may be intended for different .co files, but generate the same .o file
+    # Mark duplicate kernels to avoid race condition
+    # @TODO improve organization so this problem doesn't appear
+    visited = set()
+    count = 0
+    for kernel in kernels:
+        if kernel["KernelLanguage"] == "Assembly":
+            curr = kernelWriterAssembly.getKernelFileBase(kernel)
+            kernel.duplicate = curr in visited
+            count += curr in visited
```
ellosel, line 593:
```suggestion
            count += kernel.duplicate
```
is equivalent and doesn't require searching the hash table.

Tensile/Tests/unit/test_TensileCreateLibrary.py
```diff
@@ -427,9 +463,32 @@ def testWindows():
...
     testWindows()
 
+def test_markDuplicateKernels(setupSolutionsAndKernels):
+    _, kernels, kernelWriterAssembly, _ = setupSolutionsAndKernels
+
+    shortname_idx = 0
+    custom_idx1 = 1
+    custom_idx2 = 2
+
+    # Use deepcopy here, otherwise when the entry is updated laster, both entries will be
```
ellosel, line 483:
```suggestion
    # Use deepcopy here, otherwise when the entry is updated later, both entries will be
```

Tensile/Tests/unit/test_TensileCreateLibrary.py
```diff
@@ -427,9 +463,32 @@ def testWindows():
...
+    kernels_out = TensileCreateLibrary.markDuplicateKernels(kernels, kernelWriterAssembly)
+
+    for i, k in enumerate(kernels_out):
+        if i == custom_idx2:
+            assert k.duplicate == True, f"CustomKernelName: 'DUPLICATE' wasn't located, instead found {kernelWriterAssembly.getKernelFileBase(k)}"
+        elif i == len(kernels) - 1:
+            assert k.duplicate == True, f"Shortened name {kernelWriterAssembly.getKernelFileBase(k)} wasn't located"
+        elif k["KernelLanguage"] == "Assembly":
+            assert k.duplicate == False
+
```
ellosel: Can you add a check for the length of `kernels_out` because it is difficult to know how many loop iterations to expect and it also serves a high level check of the logic file in the event that it changes. 

Would both or one kernel wind up marked duplicate? I would expect that we only mark duplicates after we see it once otherwise we might remove all instances of the kernel.

The output on failure suggests that we didn't locate DUPLICATE but what if we did locate DUPLICATE and it wasn't marked appropriately. Isn't that what we are testing?

bstefanuk: Good points, I'll add a length check for sure.

> Would both or one kernel wind up marked duplicate? I would expect that we only mark duplicates after we see it once otherwise we might remove all instances of the kernel.

Only the second instance will be marked duplicate. I'll add more assert statements to ensure that the first instances are marked False.

> what if we did locate DUPLICATE and it wasn't marked appropriately. Isn't that what we are testing?

What we are testing is whether for two kernels, both with the same CustomKernelName, the second one is marked with the "duplicate" attribute. "DUPLICATE" is a property of the kernel, the statement is used to identify which kernel should have been marked as a duplicate.

I could change it to `f"Kernel with custom name {kernels[custom_idx2]["CustomKernelName"]} should be a duplicate, but isn't, found {kernelWriterAssembly.getKernelFileBase(k)} instead"`. Thoughts?

ellosel (approved):

PR #1970, opened 2024-07-15
Use conditional imports for new dependencies

bstefanuk:
This is a patch for resolving imports for *rich* and *CSafeLoader/CSafeDumper*---Python and OS-level dependencies, respectively.

Testing has demonstrated that, while Ubuntu 20.04 onwards comes equipped with appropriate libyaml development packages, RHEL and SLES don't necessarily. The changes in this PR add a *ConditionalImports.py* file so that that new dependencies can be added with stable fallbacks.

These changes are intended as a stopgap; a long-term solution may be to integrate the required dependencies into the CI staging images (in which case the changes in this PR may be reverted), if this is not plausible, then documenting steps for users to setup appropriate dependencies will be important.

### DIFF
--- a/HostLibraryTests/configs/ConvertYAML.py
+++ b/HostLibraryTests/configs/ConvertYAML.py
@@ -24,6 +24,7 @@
 
 from __future__ import print_function
 
+
 import itertools
 import sys
 import time
@@ -32,7 +33,7 @@ import yaml
 sys.path.append('../..')
 from Tensile.SolutionStructs import Solution
 from Tensile import Utils
-
+from Tensile.Utilities.ConditionalImports import yamlDumper
 
 def merge_libraries(args):
     inFiles = args[:-1]
@@ -55,7 +56,7 @@ def merge_libraries(args):
 
     with open(outFile, 'w') as outf:
         if True:
-            yaml.dump(outData, outf, yaml.CSafeDumper)
+            yaml.dump(outData, outf, yamlDumper)
         else:
             import json
             json.dump(outData, outf, sort_keys=True, indent=2, separators=(",", ": "))
@@ -82,7 +83,7 @@ def convert_one(args):
 
     with open(args[1], 'w') as outFile:
         if True:
-            yaml.dump(outData, outFile, yaml.CSafeDumper)
+            yaml.dump(outData, outFile, yamlDumper)
         else:
             import json
             json.dump(outData, outFile, sort_keys=True, indent=2, separators=(",", ": "))
--- a/HostLibraryTests/configs/ConvertYAML.py
+++ b/HostLibraryTests/configs/ConvertYAML.py
@@ -24,7 +24,6 @@
 
 from __future__ import print_function
 
-
 import itertools
 import sys
 import time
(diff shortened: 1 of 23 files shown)


HostLibraryTests/configs/ConvertYAML.py
```diff
@@ -24,6 +24,7 @@
 
 from __future__ import print_function
 
+
```
ellosel, line 27: whitespace

Tensile/TensileCreateLib/ParseArguments.py
```diff
@@ -46,7 +46,7 @@ def splitExtraParameters(par):
     return (key, value)
 
 
-def parseArguments(input: List[str] | None = None) -> Dict[str, Any]:
+def parseArguments(input: Optional[List[str]] = None) -> Dict[str, Any]:
```
ellosel, line 49: Is this preferred over |?

Tensile/Utilities/ConditionalImports.py
```diff
@@ -0,0 +1,24 @@
...
+    TENSILE_TERM_COLORS = True
+except ImportError:
+    _print = print
+
+
+def showwarning(message, category, filename, lineno, file=None, line=None):
+    msg = f"{category.__name__}: {message}"
+    if TENSILE_TERM_COLORS:
+        msg = f"[yellow]{msg}[/yellow]"
+    _print(msg)
```
ellosel, line 13: does this belong here?

bstefanuk: This is a tricky one. While it may belong somewhere more appropriate, I would prefer not to expose the conditional print to the whole program, however, it may be necessary.

bstefanuk: I found a (somewhat hacky) work around to importing `print` appropriately so the location of showwarning has been restored.

Tensile/TensileCreateLib/ParseArguments.py
bstefanuk: Some of the images I pulled for testing were still on Python3.6 which doesn't support the `|` operator for type-hints. I changed this to appease these older versions, however, there are still parts of the code that use `|` operators so it may have been a temporary environment issue. I will test the original again in the appropriate images and determine whether this change can be reverted.

Tensile/Tests/extended/convolution_config/YamlBuilder/YamlBuilder.py
```diff
@@ -27,6 +27,8 @@
 import copy, operator, pytest
 from functools import reduce
 from Tensile.SolutionStructs import ConvProblem
+from Tensile.Utilities.ConditionalImports import yamlDumper
+
 import yaml
```
smalekta, line 32: NBD but seems like after your change the yaml is not needed and can be removed from imports

ellosel: It is needed to qualify calls like `dump` and `load` no?

smalekta: yes, you are correct I missed that

ellosel (approved):
Looks good!

PR #1971, opened 2024-07-16
Generalize linux distro handling

trixirt:
Addresses this issue [ Generalize linux distro handling ](https://github.com/ROCm/Tensile/issues/1969)

On Fedora, /opt/rocm does not exist and the ROCm locations are determined by Fedora's packaging requirements.  So add some logic to find these tools if the exiting logic fails and add some tests to make sure nothing breaks for other folks.


Tensile/Common.py
```diff
@@ -2338,6 +2342,12 @@ def assignGlobalParameters( config ):
       tPrint(3, " %24s: %8s (unspecified)" % (key, defaultValue))
 
   globalParameters["ROCmPath"] = "/opt/rocm"
+  # /opt/rocm is not always the default location
+  if not os.path.isdir(globalParameters["ROCmPath"]):
+      if platform.system() == "Linux":
+          if distro.id() == "fedora":
+              # Fedora packages ROCm as part of its distro to system paths
+              globalParameters["ROCmPath"] = "/usr"
```
cgmb, line 2350: The comment is also true of Debian, Gentoo and Ubuntu. IMO, `/usr` is a standard path that should be considered on any Linux, as otherwise, it sends us down a rabbit hole of adding distro names one-by-one (e.g., 'linuxmint' is effectively just Ubuntu for this purpose).

trixirt: ok i will remove the distro check and go with '/usr' for everyone linux

Mystro256: you could also pass the prefix from cmake too (CMAKE_INSTALL_PREFIX)

Mystro256:
Honestly, a lot of the /opt/rocm paths could be replaced with @CMAKE_INSTALL_PREFIX@ with templating (see https://cmake.org/cmake/help/latest/command/configure_file.html)

Tensile/Common.py
cgmb: @Mystro256, it's fairly common to run Tensile from the source tree without building/installing it.

trixirt:
> Honestly, a lot of the /opt/rocm paths could be replaced with @CMAKE_INSTALL_PREFIX@ with templating (see https://cmake.org/cmake/help/latest/command/configure_file.html)

This sounds like a good idea for the cmake buildable rocm packages. I do not know if this applies to Tensile.

babakpst (changes requested):
on hold until after FC.

tcgu-amd:
Hi @trixirt, thank you for your effort trying to improve the out-of-box user experience of ROCm on Fedora/CentOS, we really do appreciate it! However, after some discussions, we felt that this change does not align with ROCm's current support policy, since it wouldn't really make sense to push out-of-box support for Tensile on Fedora in particular when other parts of ROCm are not being officially supported. This might prompt demands for support on other unsupported OS too which will be out of our current development scope. We had to draw the line somewhere, and we reasoned that if a user were to get a piece of software running on an unsupported OS, some level of tinkering should be expected; hence allowing path customization through the ROCM_PATH variable should be sufficient for now. 

At the time of writing, Fedora is doing independent distribution of ROCm that actually branches off from the official ROCm. I suggest trying to talk to the Fedora team for how to offer better out-of-box support for components such as Tensile on their OS for now. 

Thank you for your understanding!

PR #1973, opened 2024-07-16
Remove deepcopy in TensileCreateLibrary

ellosel:
The `deepcopy` function is used throughout Tensile and one of the primary bottlenecks. This PR removes the deep copies of Solutions - namely those that occur in `TensileCreateLibrary`.  This significantly improves the turnaround time of `TensileCreateLibrary` by up to 40% depending on the input to the main.

Ideally, we could make the Solution class immutable as a part of these changes (which we attempted) but the solution changes in ways during runtime that would require a major refactor. 

To ensure that the program still behaves correctly, we ran the following tests:

- Standard CI tests
- A/B comparison to devel of manifest generation using TensileCreateLibrary 
- A/B comparison to devel of manifest generation   using rocBLAS install.sh
- Ran rocBLAS-tests on the feature branch without failure


Tensile/SolutionStructs.py
```diff
@@ -1720,60 +1721,61 @@ class Solution(collections.abc.Mapping):
...
   ########################################
   # get a list of kernel parameters for this solution
   def getKernels(self):
-    kernel = deepcopy(self)
-    kernel._state.update({"Kernel": True})
+    kernel = self
+    kernel._mutable_state.update({"Kernel": True})
     kernels = []
     kernels.append(kernel)
     return kernels
```
bstefanuk, line 1781: This function will now return a reference to the object itself instead of a full copy, which could have side-effects if calling code requires the original Solution object unchanged. This isn't necessarily avoidable, but it would be great if we could consume the original object.

Tensile/SolutionStructs.py
```diff
@@ -4945,20 +4947,27 @@ def getParameterValueAbbreviation( key, value ):
...
   def __getitem__(self, key):
+    if key in self._mutable_state:
+      return self._mutable_state[key]
     return self._state[key]
 
   def __setitem__(self, key, value):
+    printWarning(f"Mutating solution: {self._name} {key}")
     self._name = None
-    self._state[key] = value
+    self._mutable_state[key] = value
```
bstefanuk, line 4971: What if a key that already exists on `_state` is added via __set_item__? Should we reject over-writing immutable state with mutable state?

Similarly, if a key exists in both `_state` and `_mutable_state`, __getitem__ will prefer the mutable option. Is this the behaviour we're after in all cases?

Tensile/TensileCreateLibrary.py
```diff
@@ -1373,36 +1372,6 @@ def writeBenchmarkClientFiles(libraryWorkingPath, tensileSourcePath, solutions,
...
-    metaData = {"ProblemType": problemType}
-    LibraryIO.writeYAML(mataDataFilePath, metaData)
-
-    codeObjectFiles, newLibrary = writeBenchmarkClientFiles(
-        libraryWorkingPath, tensileSourcePath, solutionList, cxxCompiler
-    )
-
-    return (codeObjectFiles, newLibrary)
-
-
```
bstefanuk, line 1405: Why is this deleted? No longer needed?

bstefanuk (commented):
Changes look promising, a few questions about default behaviour and potential side-effects.

Tensile/SolutionStructs.py
ellosel: The mutated state should be considered the "truth" at least that is my intention. Probably need to clean up a few of the functions to be sure that this is enforced.

Tensile/TensileCreateLibrary.py
ellosel: I was planning to remove the deepcopy but decided to look to see if this function was used anywhere and I cannot find evidence of it being used.

Tensile/SolutionStructs.py
ellosel: This impl was abandoned

babakpst (dismissed):
on hold until after FC.

nakajee:
How did you verify this works same as before?

ellosel:
> How did you verify this works same as before?

- Standard CI tests
- A/B comparison to devel of manifest generation using TensileCreateLibrary
- A/B comparison to devel of manifest generation using rocBLAS install.sh
- Ran rocBLAS-tests on the feature branch without failure
- Ran rocblas PTS tests

nakajee (approved):
Thanks for your comment.
I do not have further comment as long as you confirmed TensileCreateLibrary and rocblas build.

PR #1974, opened 2024-07-16
Improve rocBLAS build output

bstefanuk:
rocBLAS's build should be as clean as possible, with some build reporting details. This PR resolves a warning that was previously emitted in ContractionSolution.cpp and suppresses warnings that we are not in a position to resolve at the moment when verbosity is set to 0.

Collectively, these changes print the following output when building rocBLAS, as it relates to Tensile.

```
[  0%] Generating libraries with TensileCreateLibrary
[  0%] Generating /mnt/host/rocBLAS-internal/build/release/include/rocblas/internal/rocblas_device_malloc.hpp
[  0%] Built target rocblas_device_malloc
[  1%] Generating prototypes from /mnt/host/rocBLAS-internal/library/src.
[  1%] Built target rocblas_proto_templates
# Reading logic files: 32 threads, 154 tasks... Done in 3.1 secs
# Generating kernels: 32 threads... Done in 10.2 secs
# Compiling source kernels: 32 threads... Done in 1.2 secs
[  1%] Verifying files in /mnt/host/rocBLAS-internal/build/release/Tensile/library/TensileManifest.txt were generated
```

### DIFF
--- a/Tensile/Parallel.py
+++ b/Tensile/Parallel.py
@@ -80,18 +80,15 @@ def ParallelMap(function: Callable, objects: Any, message: str="", enable: bool=
     from . import Utils
     threadCount = CPUThreadCount(enable)
     
-    if threadCount <= 1 and globalParameters["ShowProgressBar"]:
-      # Provide a progress bar for single-threaded operation.
+    if threadCount <= 1:
       return list(map(lambda objs: function(*objs), Utils.tqdm(objects, desc=message)))
-    
-    message += f": {threadCount} threads"
-    try:
-      message += f", {len(objects)} tasks"
-    except TypeError: pass
-    
-    pcall = pcallWithGlobalParamsMultiArg if multiArg else pcallWithGlobalParamsSingleArg
+        
     inputs = list(zip(objects, itertools.repeat(globalParameters)))
+    message += f": {threadCount} threads, {len(inputs)} tasks"
+
     pargs = Utils.tqdm(inputs, desc=message)
+    pcall = pcallWithGlobalParamsMultiArg if multiArg else pcallWithGlobalParamsSingleArg
+
     rv = Parallel(n_jobs=threadCount)(delayed(pcall)(function, a, params) for a, params in pargs)
     
     return rv
--- a/Tensile/Parallel.py
+++ b/Tensile/Parallel.py
@@ -24,7 +24,6 @@
 
 import itertools
 import os
-import sys
 
 from joblib import Parallel, delayed
 
@@ -72,13 +71,14 @@ def ParallelMap(function, objects, message="", enable=True, multiArg=True, verbo
     # Provide a progress bar for single-threaded operation.
     return list(map(lambda objs: function(*objs), Utils.tqdm(objects, msg=message)))
   
-  message += f" with {threadCount} threads"
+  message += f": {threadCount} threads"
   try:
-    message += f" for {len(objects)} tasks"
+    message += f", {len(objects)} tasks"
   except TypeError: pass
   
   pcall = pcallWithGlobalParamsMultiArg if multiArg else pcallWithGlobalParamsSingleArg
-  pargs = Utils.tqdm(zip(objects, itertools.repeat(globalParameters)), msg=message)
+  inputs = list(zip(objects, itertools.repeat(globalParameters)))
+  pargs = Utils.tqdm(inputs, msg=message)
   rv = Parallel(n_jobs=threadCount, verbose=verbose)(delayed(pcall)(function, a, params) for a, params in pargs)
   
   return rv
--- a/Tensile/Parallel.py
+++ b/Tensile/Parallel.py
@@ -24,6 +24,7 @@
 
 import itertools
 import os
+from typing import Any, Callable
 
 from joblib import Parallel, delayed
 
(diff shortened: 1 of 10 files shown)


Tensile/TensileCreateLibrary.py
```diff
@@ -244,7 +244,7 @@ def buildSourceCodeObjectFile(CxxCompiler, outputPath, kernelFile):
         archFlags = ["--offload-arch=" + arch for arch in cmdlineArchs]
 
         # needs to be fixed when Maneesh's change is made available
-        hipFlags = ["-D__HIP_HCC_COMPAT_MODE__=1"]
+        hipFlags = ["-D__HIP_HCC_COMPAT_MODE__=1", "-v"]
```
ellosel, line 247: My expectation was that this would output:

gfx900
gfx90a
etc

as we begin to build for those targets. Is this the case?

bstefanuk: I don't see this output in the rocBLAS build output, no.

TorreZuk (approved):
That looks good as long as you aren't hiding warnings that indicate a bad build.  Maybe cmake can just echo the gfx list via status, don't need to see a bunch of compile details (maybe Cory needs more).   In rocblas we do a verbose compile of a single file for QC purposes so fine if you did that too.

AlexBrownAMD (approved):

Tensile/TensileCreateLib/ParseArguments.py
```diff
@@ -297,6 +297,8 @@ def parseArguments(input: List[str] | None = None) -> Dict[str, Any]:
     if args.GenerateSourcesAndExit:
         # Generated sources are preserved and go into output directory
         arguments["WorkingPath"] = arguments["OutputPath"]
+    if args.PrintLevel == 0:
```
nakajee, line 300: I do not think it is a good way to ignore warning at any print level.

nakajee: Any request to do so?

nakajee: I might misunderstand this.
Isn't this related to compiler warning?

nakajee: If this is not related to compiler warning, I am OK with this.

bstefanuk:
To address concerns about missing warnings about bad builds, I have updated the patch to only filter DeveloperWarnings, and changed the default warning to UserWarning. This way, any existing warnings will continue to be emitted when `--verbose=0`, unless they're explicitly set as a DeveloperWarning.

Tensile/TensileCreateLib/ParseArguments.py
bstefanuk: This is not directly related to compiler warnings, though some TensileCreateLibrary warnings related to compiler issues may be emitted. Please see my newest comment.

nakajee (approved):
Thanks for your update.
I do not have further comments.

Tensile/TensileCreateLibrary.py
```diff
@@ -700,7 +700,7 @@ def writeKernels(
         itertools.repeat(kernelWriterSource),
         itertools.repeat(kernelWriterAssembly),
     )
-    results = Common.ParallelMap(processKernelSource, kIter, "Generating kernels")
+    results = Common.ParallelMap(processKernelSource, kIter, "Generating kernels", verbose=0)
```
ellosel, line 703: what is verbose=0 doing here?

bstefanuk: Removed.

Tensile/Parallel.py
```diff
@@ -55,39 +55,43 @@ def pcallWithGlobalParamsSingleArg(f, arg, newGlobalParameters):
...
-  
-  return rv
+    Returns:
+        A list containing the results of applying **function** to each item in **objects**.
+    """
+    from .Common import globalParameters
+    from . import Utils
+    threadCount = CPUThreadCount(enable)
+    
+    if threadCount <= 1 and globalParameters["ShowProgressBar"]:
```
ellosel, line 83: This is interesting. I wonder if is truly necessary.

Tensile/Parallel.py
```diff
@@ -55,39 +55,43 @@ def pcallWithGlobalParamsSingleArg(f, arg, newGlobalParameters):
...
+      # Provide a progress bar for single-threaded operation.
+      return list(map(lambda objs: function(*objs), Utils.tqdm(objects, desc=message)))
+    
+    message += f": {threadCount} threads"
+    try:
+      message += f", {len(objects)} tasks"
+    except TypeError: pass
+    
+    pcall = pcallWithGlobalParamsMultiArg if multiArg else pcallWithGlobalParamsSingleArg
+    inputs = list(zip(objects, itertools.repeat(globalParameters)))
```
ellosel, line 93: Do we know how this impacts our memory footprint?

Tensile/Source/lib/source/ContractionSolution.cpp
```diff
@@ -1787,7 +1787,7 @@ namespace Tensile
             }
         }
 
-        return remainder >= get<1>(predictionCurve.back()) ? 0 : 1;
+        return remainder >= std::get<1>(predictionCurve.back()) ? 0 : 1;
```
ellosel, line 1790: Did we confirm that the compiler warning goes away in rocblas?

Tensile/TensileCreateLib/ParseArguments.py
```diff
@@ -297,6 +297,8 @@ def parseArguments(input: Optional[List[str]] = None) -> Dict[str, Any]:
...
+        warnings.filterwarnings("ignore", category=DeveloperWarning)
```
ellosel, line 301: This is nice.  We could probably replace tPrint with something like this.

Tensile/TensileCreateLibrary.py
ellosel, line 247: Are we removing this now?

Tensile/Parallel.py
bstefanuk: I have not profiled the memory consumption, only runtime stats.

Tensile/Utils.py
```diff
@@ -22,30 +22,137 @@
...
+        sys.stdout.write('\r' + self.msg + f'... Done in {elapsedTime:.1f} secs\n')
         sys.stdout.flush()
 
 def iterate_progress(obj, *args, **kwargs):
-    try:
-        progress = ProgressBar(len(obj))
-    except TypeError:
-        progress = SpinnyThing()
+    if 'desc' not in kwargs:
+        printWarning("No message provided for TQDM progress bar/spinner", DeveloperWarning)
```
ellosel, line 147: Should we use the name of the function via introspection to set the description if we don't have one?

Tensile/Source/lib/source/ContractionSolution.cpp
bstefanuk: Confirmed.

Tensile/Utils.py
```diff
@@ -22,30 +22,137 @@
...
+        kwargs['desc'] = 'Processing unknown function'
+    if 'total' in kwargs:
+        progress = ProgressBar(kwargs['total'], kwargs['desc'])
+    else:
+        try:
+            progress =  ProgressBar(len(obj), kwargs['desc'])
+        except TypeError:
+            progress = SpinnyThing(kwargs['desc'])
```
ellosel, line 155: It appears we now have the ability to consistently use the progress bar - am I correct?

Tensile/TensileCreateLibrary.py
bstefanuk: Good catch, will remove.

Tensile/Utils.py
bstefanuk: We do now have the ability to every call to `Utils.tqdm` to use a progress bar. Either the object has to have `__len__` defined on it or you can pass in the kwarg `total=N` where N is the known length. If neither of these are true, it will default to the Spinner.

requirements.txt
```diff
@@ -2,4 +2,4 @@ pyyaml
 msgpack
 joblib
 rich
-
+tqdm
```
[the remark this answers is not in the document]
ellosel: This won't cause breaks downstream because it is optional - correct?

Tensile/Utils.py
bstefanuk: We *could*, however, I would prefer the user provide a useful description, and this will encourage appropriate usage.

requirements.txt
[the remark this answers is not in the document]
bstefanuk: This was temporary and has been removed. In the future I would like to migrate away from the tqdm dependency since the custom progress bar and spinner provide good functionality without needing another library.

Tensile/Parallel.py
bstefanuk: This looks like a new addition from the diff, but was already there. Not sure it's adding much. In what case would we not want to see progress bars? Even so, reaching into this conditional is an edge case and it's not fully operational via the `enable` flag.

I would be happy to remove global parameters from this functional altogether.

PR #1976, opened 2024-07-18
Reorder extensions for windows in which function

ellosel:
Resolving windows failure:

[11/677] Verifying files in C:/github/rocBLAS/build/release/Tensile/library/TensileManifest.txt were generated
DeveloperWarning: HardwareMonitor currently disabled for gfx941, gfx942,
gfx1100, gfx1101, gfx1102, gfx1200, gfx1201
Bareword found where operator expected at C:\hipCI\bin\hipcc.bat line 1, near "@set HIPCC"
        (Missing operator before HIPCC?)
Array found where operator expected at C:\hipCI\bin\hipcc.bat line 2, near "@perl"
        (Missing semicolon on previous line?)
Operator or semicolon missing before %HIPCC at C:\hipCI\bin\hipcc.bat line 2.
Ambiguous use of % resolved as operator % at C:\hipCI\bin\hipcc.bat line 2.
syntax error at C:\hipCI\bin\hipcc.bat line 1, near "@set HIPCC"
Execution of C:\hipCI\bin\hipcc.bat aborted due to compilation errors.
DeveloperWarning: Error: hipcc running --version Command '['perl',
'C:\\hipCI\\bin\\hipcc.bat', '--version']' returned non-zero exit status 255.

---

**Resolution**

On Windows we do something like 
```python
cmd = "perl " + which("hipcc") + " --version"
```
But `perl hipcc.bat --version` doesn't work (though the file exists), however, `perl hipcc --version` does work, and hipcc.exe isn't a thing so we skip it when calling `which(hipcc)`. Additionally, we want `which(clang++)` to return clang++ or clang++.exe when on Windows.

### DIFF
--- a/Tensile/Common.py
+++ b/Tensile/Common.py
@@ -2299,7 +2299,7 @@ def which(p):
     if supportedCompiler(p) and 'CMAKE_CXX_COMPILER' in os.environ and os.path.isfile(os.environ['CMAKE_CXX_COMPILER']):
         return os.environ['CMAKE_CXX_COMPILER']
     if os.name == "nt":
-        exes = [p+x for x in ['', '.bat']]  # bat may be front end for file with no extension
+        exes = [p+x for x in ['.exe', '', '.bat']]  # bat may be front end for file with no extension
     else:
         exes = [p+x for x in ['', '.exe', '.bat']]
     system_path = os.environ['PATH'].split(os.pathsep)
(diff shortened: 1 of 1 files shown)


ellosel:
@TorreZuk my only concern with these changes is the interaction with launcher here:

https://github.com/ROCm/Tensile/blob/develop/Tensile/TensileCreateLibrary.py#L263

it isn't clear to me how launcher is used in practice.

TorreZuk:
> @TorreZuk my only concern with these changes is the interaction with launcher here:
> 
> https://github.com/ROCm/Tensile/blob/develop/Tensile/TensileCreateLibrary.py#L263
> 
> it isn't clear to me how launcher is used in practice.

I added commentary in the internal ticket.   Launcher could be something like ccache but as hipcc is now deprecated I would think it would be applied on clang++.exe which would be fine.   hipcc is only for version detection, I suggested alternative but @yoichiyoshida mentioned some other constraints so ask him if you want to revisit that.   perl invocation should only be done on perl files, just like python file.exe we don't want.

bstefanuk (approved):
I've supplemented the description with details about why this solution works. Code change looks good.

PR #1980, opened 2024-07-19
Fix printing ASM capabilities for ROCm < 6.3

bstefanuk:
Fixes a defect whereby ROCm compilers under 6.3 print ~800 lines to stdout because of an assembler capability mismatch between those computed at runtime, and those cached in *AsmCaps.py*.

### DIFF
--- a/Tensile/Common.py
+++ b/Tensile/Common.py
@@ -26,9 +26,10 @@ from . import __version__
 from . import Parallel
 from .Utilities.ConditionalImports import print, TENSILE_TERM_COLORS
 from collections import OrderedDict
+
 from copy import deepcopy
 from .AsmCaps import CACHED_ASM_CAPS
-from typing import Optional
+from typing import Any, NamedTuple, Optional, Tuple, Dict
 
 import math
 import os.path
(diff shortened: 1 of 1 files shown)


Tensile/Common.py
```diff
@@ -2310,6 +2318,45 @@ def which(p):
...
+        This function modifies `globalParameters` and `cachedAsmCaps` in place.
+    """
+    compilerVersion = CompilerVersion(
+        *[int(c) for c in globalParameters["HipClangVersion"].split(".")[:2]]
+    )
+    supportedISA = globalParameters["SupportedISA"]
+    to_remove = []
+
+    for v in supportedISA + [(0, 0, 0)]:
+        if v[0] == 12 and not (compilerVersion.major == 6 and compilerVersion.minor == 3):
```
nakajee, line 2348: Should this be like 
  not (major > 6 or (major ==6 and minor >=3))

nakajee:
Thanks for taking care of this issue.
I have been waiting for fix for this print.

ellosel (approved):

Tensile/Common.py
bstefanuk: I agree, I will update this to include all compiler versions 6.3 and up.

nakajee (approved):
Thanks for your update.
Looks good.

PR #1981, opened 2024-07-20
Don't redundantly remove duplicates

ellosel:
This PR removes a few pessimizations from Tensile. In particular, there is a call to fromkeys on a kernels list to deduplicate the list, but this work is already done on the input to the function. This operation is expensive because it requires hashing a solution which requires computing the solution name and computing solution names is a bottleneck. Further, a existence operation was cleaned up from:

```
key in list(mydict.keys())
```

to 

```
key in mydict
```

which has a small but measurable impact on runtime. Lastly, the `Solution.getKernels()` function return value was changed from returning a single element list (presumably to support operator `+` for list concat) to return the kernel in favor of `list.append()`.

### DIFF
--- a/Tensile/BenchmarkProblems.py
+++ b/Tensile/BenchmarkProblems.py
@@ -127,7 +127,7 @@ def writeBenchmarkFiles(stepBaseDir, solutions, problemSizes, \
 
     # get unique kernels and kernel helpers
     for solution in Utils.tqdm(solutions, desc="Finding unique solutions"):
-        solutionKernels = solution.getKernels()
+        solutionKernels = [solution.getKernels()]
         for kernel in solutionKernels:
             kName = Solution.getNameFull(kernel)
             if kName not in kernelNames:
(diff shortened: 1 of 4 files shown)


Tensile/BenchmarkProblems.py
```diff
@@ -127,7 +127,7 @@ def writeBenchmarkFiles(stepBaseDir, solutions, problemSizes, \
 
     # get unique kernels and kernel helpers
     for solution in Utils.tqdm(solutions, desc="Finding unique solutions"):
-        solutionKernels = solution.getKernels()
+        solutionKernels = [solution.getKernels()]
```
bstefanuk, line 130: Won't this necessarily be a single element list? In which case we can remove the for loop?

ellosel: Fixed

babakpst (approved):

bstefanuk (approved):
Looks good.

PR #1983, opened 2024-07-23
Add kernel helper write fxns

bstefanuk:
This is a part of a larger refactor of `writeKernels` in _TensileCreateLibrary_.

**Notable changes**

- The logic that encapsulated the writing of kernel helper objects has been reduced to 4 lines within `writeKernels`.
- A new KernelFileContextManager has been created to enforce proper file IO and thereby reduces the chance of bugs.

**Testing**

- All new functions have been tested with coverage >90%
- TensileManifest.txt, generated during rocBLAS installation, matches before (develop) and after these changes (unsorted match)


babakpst (dismissed):
on hold until after FC.

Tensile/TensileCreateLib/KernelFileContext.py
```diff
@@ -0,0 +1,156 @@
...
+    Returns:
+        A tuple containing two `TextIOWrapper` objects for the source and header files respectively.
+
+    Notes:
+        - The source file is named "Kernels.cpp" and the header file is named "Kernels.h".
+        - Both files are opened in append mode with UTF-8 encoding.
+    """
+    srcFilename = Path(outputPath).resolve() / "Kernels.cpp"
+    hdrFilename = Path(outputPath).resolve() / "Kernels.h"
+    return open(srcFilename, "a", encoding="utf-8"), open(hdrFilename, "a", encoding="utf-8")
```
ellosel, line 78: Minor style suggestion:
```suggestion
    return (open(srcFilename, "a", encoding="utf-8"),  
                open(hdrFilename, "a", encoding="utf-8"))
```

Tensile/TensileCreateLib/KernelFileContext.py
```diff
@@ -0,0 +1,156 @@
...
+
+    Returns:
+        A tuple containing two `TextIOWrapper` objects for appending to the source and header files respectively.
+
+    Notes:
+        - The base file name is derived by removing the ".cpp" extension from the first kernel file name.
+        - The source and header files are named using the base name with ".cpp" and ".h" extensions respectively.
+        - Files are opened in append mode with UTF-8 encoding.
+    """
+    filename = kernelFiles[0].replace(".cpp", "")
```
ellosel, line 95: Is this an append?

Tensile/TensileCreateLib/KernelFileContext.py
```diff
@@ -0,0 +1,156 @@
...
+    return open(filename + ".cpp", "a", encoding="utf-8"), open(
```
ellosel, line 96: The wrapping of open is a little awkward.

Tensile/TensileCreateLibrary.py
```diff
@@ -682,6 +684,64 @@ def noBuildError(kernel):
...
+
+    Arguments:
+        ko: Kernel object to extract content from.
+
+    Returns:
+        Tuple of data: (error code, source code, header code, kernel name)
+
+    Notes:
+        - This function is very similar to **processKernelSource**---the functionality
+        should be merged.
```
ellosel, line 698: Do we want this in the docs?

Tensile/TensileCreateLibrary.py
```diff
@@ -682,6 +684,64 @@ def noBuildError(kernel):
...
+    """
+    name = ko.getKernelName()
+    err, src = ko.getSourceFileString()
+    hdr = ko.getHeaderFileString()
+    return err, [CHeader, src], [CHeader, hdr], name
```
ellosel, line 703: https://pypi.org/project/result/

maybe in the future...

Tensile/Utilities/toFile.py
```diff
@@ -37,10 +40,25 @@ def toFile(outputFile: Path, contents: List[str], delimiter: str = "\n") -> None
...
+        if outputFile.closed:
+            raise ValueError(f"Provided file {outputFile.name} is already closed: outputFile must be an open file, or a Path.")
+
+    f = outputFile if isTextIO else open(outputFile, "w")
 
-    with open(outputFile, "w") as generatedFile:  
-      for filePath in contents:
-        generatedFile.write(f"{filePath}{delimiter}")
+    for chunk in contents:
+        f.write(f"{chunk}{delimiter}")
```
ellosel, line 59: If this fails and we aren't a TextIO then we leak a reseource.

Tensile/TensileCreateLib/KernelFileContext.py
```diff
@@ -0,0 +1,156 @@
...
+            outputPath (Path): The output path where kernel files are located or will be created.
+            kernelFiles (Optional[List[str]]): A list of kernel file names.
+        """
+        self.numMergedFiles = params["NumMergedFiles"]
+        self.mergeFiles = params["MergeFiles"]
+        self.lazyLoading = params["LazyLibraryLoading"]
+        self.outputPath = outputPath
+        self.kernelFiles = kernelFiles
+        self.kernelSourceFile = None
+        self.kernelHeaderFile = None
```
ellosel, line 129: Are there any set of parameter values that we know are violating assumptions about this context manager that we can test here?

For example are all of the following ok:

1. mergeFiles = True, lazyLoading = True
2. mergeFiles = True, lazyLoading = False
3. mergeFiles = False, lazyLoading = True
4. mergeFiles = False, lazyLoading = False
5. All of the above with different values for numMergedFiles?
6. What if outputPath is empty?
7. What if kernelFiles is empty?

Also, rather than taking a dict should we take three params explicitly and type them?

Tensile/TensileCreateLib/KernelFileContext.py
bstefanuk: This is black's default formatting. I don't love it either, but still support using automated formatting even if sometimes it feels a little awkward on edge cases.

Tensile/TensileCreateLibrary.py
bstefanuk: If we don't add it somewhere in the docs, I suspect it will never be resolved. Alternatively, we could track it with a ticket.

Tensile/TensileCreateLib/KernelFileContext.py
ellosel: I read this in reverse... this is filename without an extension.

Tensile/TensileCreateLibrary.py
bstefanuk: I would love to add then instead of using magic return codes. However, it adds another dependency which, as we've learned, is a heavier lift that expected.

Tensile/Utilities/toFile.py
bstefanuk: The re-usability of this function is quite nice IMO, so another option would be to use a try/except and handle any failures here gracefully.

Tensile/TensileCreateLib/KernelFileContext.py
bstefanuk: I like this. I can add in a method to verify pre-conditions and update the function to accept explicit arguments.

Tensile/TensileCreateLibrary.py
ellosel: I would prefer the latter

ellosel (approved):

PR #1984, opened 2024-07-23
Cache abbreviations used to compute kernel names

ellosel:
One of the most expensive operations in Tensile is the cost of calling `Solution.getNameFull`. In part, this function is expensive because we currently map each solution parameter key to an abbreviated name using only the capital letters of the parameter name. This is an O(N) operation and applied to every kernel parameter for every kernel. These abbreviations do not change during runtime, so we are redundantly computing the abbreviations. 

In this PR, we compute all possible parameter abbreviations in `Common` once and use them later. This has a measurable improvement on the turnaround time for TensileCreateLibrary.


babakpst (changes requested):
on hold until after FC.

Tensile/Common.py
```diff
@@ -1519,6 +1519,14 @@ def supportedCompiler(compiler: str) -> bool:
     "MinKForGSU":                   [16,32,64,128,256]
     }
 
+def getNameAbbreviation(name):
+  specialValues = {
+    'MACInstruction': '' # Conflicts with MatrixInstruction, but _MAD and _FMA should be enough differentiation for the kernel name.
```
bstefanuk, line 1524: We want to return an empty string for the "MACInstruction"?

Tensile/Common.py
```diff
@@ -1519,6 +1519,14 @@ def supportedCompiler(compiler: str) -> bool:
...
+  }
+  if name in specialValues: return specialValues[name]
+  return ''.join([c for c in name if not c.islower()])
```
bstefanuk, line 1527: Minor, but a generator expression here would eliminate the need to use an intermediary list
```python
return ''.join(c for c in name if not c.islower())
```

ellosel: I will update to develop and make this change

bstefanuk (approved):

PR #1988, opened 2024-07-25
Disable test that hangs

ellosel:
The f8gemm_lsu_mfma test hangs is potentially causing a hang in pipelines.


nakajee:
Thanks for your update.
Looks OK.
I added gfx942 flag to see if we can trigger gfx942 CI test.
(I did not see it before)

nakajee:
I also added ci:extended because the updated test case was in extended.

PR #1991, opened 2024-07-28
Restore original behavior

ellosel:


nakajee:
gfx942 CI test passed. That is good.
Not sure what is wrong with rhel9+gfx11.
(seems not related to this change)

babakpst (approved):

PR #1993, opened 2024-07-30
Do not use version specific requirements.txt

ellosel:
We have to support different versions of python across different operating systems. For example, sles has 3.6 whereas ubuntu has 3.10. As such the python dependencies are different so the requirements.txt file was failing sles builds.


nakajee:
How can we make sure this works?

bstefanuk:
> How can we make sure this works?

We re-run the CQE test commands in the provided SLES image.

bstefanuk (approved):

nakajee (approved):
OK. Looks good.

Issue #1989, opened 2024-07-26
[Feature]: Optional use of joblib


trixirt:
### Suggestion Description

For building Tensile as a RHEL package, the joblib package requires many packages that are not part of RHEL. 

Instead of requiring joblib, try to import joblib and when it fails to fall back to the pre joblib code or even a single threaded option.

### Operating System

RHEL

### GPU

ALL

### ROCm Component

ALL

cgmb:
Ideally, the parallelism would be managed through the build system. It is inefficient to have two resource pools (make/ninja and joblib) manage the same set of resources (CPU cores, memory).

For example, if you have a 16-core CPU, then you could give Tensile 16 threads and also give Make 16 threads. However, that means the two job queues may use 32 threads total and this could be inefficient (or use too much memory). However, if you only have 16 cores and you give 8 to Tensile and 8 to Make, then you won't get to use all your cores on the parts of the build after Tensile completes.

I expect that it's _much_ easier just to make joblib optional using a fallback like in #1970. That's would be a good start, but I hope we can eventually do better.

bstefanuk:
> Ideally, the parallelism would be managed through the build system. It is inefficient to have two resource pools (make/ninja and joblib) manage the same set of resources (CPU cores, memory).

I completely agree, however, I don't see this sort of change being implemented soon so it is may still be in our interest to support a single-threaded option. In fact, Tensile's main parallelization entry point---`ParallelMap` [here](https://github.com/ROCm/Tensile/blob/2f9ef2fbe7732dd9923f1c7a22a530ba2b651ea2/Tensile/Parallel.py#L58)---claims to support a single-threaded option via the **enable** parameter. Unfortunately, it's experimental and in need of testing.

Profilability, is another consideration. In the past I've faced issues with certain profilers that can't cope with the process churn that joblib induces. It also obscures internal function calls making it hard to profile. Having a single threaded option could help us highlight bottlenecks

Another longer-term solution that I've considered is to establish a single fan-out fan-in design with joblib so that we don't continually spawn and terminate processes. 

At the very least it's worth stabilizing the **enable** option and providing a fallback if joblib isn't installed.

ellosel:
We keep removing "optional" dependencies from the requirements files. This hasn't been a problem up to now but if requirements.txt does not contain joblib and we use the fallback then TensileCreateLIbrary will slow down considerably.

LunNova:
[...]

PR #2000, opened 2024-08-05
Single threaded

trixirt:
For this issue https://github.com/ROCm/Tensile/issues/1989
If importing joblib fails, handle like a single threaded case.
Tested with rocBLAS building.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,3 @@
 pyyaml
 msgpack
-joblib
 rich
--- a/Tensile/Parallel.py
+++ b/Tensile/Parallel.py
@@ -26,11 +26,16 @@ import itertools
 import os
 from typing import Any, Callable
 
-from joblib import Parallel, delayed
+try:
+  from joblib import Parallel, delayed
+except:
+  joblib = None
 
 def CPUThreadCount(enable=True):
   from .Common import globalParameters
   if not enable:
+    return 1
+  elif joblib is None:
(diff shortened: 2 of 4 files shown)


requirements.txt
```diff
@@ -1,4 +1,3 @@
 pyyaml
 msgpack
-joblib
```
bstefanuk, line 3: If we do want to support the optional use of joblib, I'm hesitant to use it as an opt-in dependency and would prefer it to be an opt-out dependency. This will reduce runtime significantly, particularly in scripted environments that only install the contents of **requirements.txt**.

bstefanuk:
@trixirt, thank you for your contribution. I notice that in Issue #1989 you mention
> For building Tensile as a RHEL package, the joblib package requires many packages that are not part of RHEL.
In order to best understand the context for this change could you provide some details regarding the packaging process you are using, and which transitive dependencies are being added by joblib that are not a part of RHEL 9?

bstefanuk:
We will integrate single-threaded support, along with removing the requirement for joblib, but this change has been posted in PR #2012, with an alternative implementation. Since the appropriate changes will be applied there, I am closing this ticket.

bstefanuk:
We will integrate single-threaded support, along with removing the requirement for joblib, but this change has been posted in PR #2012 with an alternative implementation. Since the appropriate changes will be applied there, I am closing this ticket.

PR #2002, opened 2024-08-07
Resolve failures on SLES

ellosel:
These changes solve test failures in SLES containers when running TensileTests. 

**Fixes**

1. The decision tree test fix is a bug (out of bounds error) in the test code that isn't caught on other platforms. 
2. The change in _Loading.cpp_ was necessary to avoid another segfault when running `getFile`. The other function achieves the same outcome but does not depend on boost optional which seems to cause a segfault when initializing in the context of this code.

### DIFF
--- a/HostLibraryTests/DecisionTree_test.cpp
+++ b/HostLibraryTests/DecisionTree_test.cpp
@@ -369,6 +369,8 @@ TEST(DecisionTree, DecisionTreeBatch)
     using SizeInRange = Predicates::Contraction::SizeInRange;
     using Range       = Predicates::Contraction::Range;
     using And         = Predicates::And<ContractionProblem>;
+    using Key         = std::array<float, 4>;
+    using DTree       = Tree<Key, std::shared_ptr<ContractionLibrary>, std::shared_ptr<ContractionSolution>>;    
     using BForest     = BasicForest<Key,
                                 ContractionProblem,
                                 std::shared_ptr<ContractionLibrary>,
--- a/HostLibraryTests/DecisionTree_test.cpp
+++ b/HostLibraryTests/DecisionTree_test.cpp
@@ -2,7 +2,7 @@
  *
  * MIT License
  *
- * Copyright (C) 2024 Advanced Micro Devices, Inc. All rights reserved.
+ * Copyright (C) 2022-2024 Advanced Micro Devices, Inc. All rights reserved.
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
--- a/HostLibraryTests/DecisionTree_test.cpp
+++ b/HostLibraryTests/DecisionTree_test.cpp
@@ -2,7 +2,7 @@
  *
  * MIT License
  *
- * Copyright (C) 2022 Advanced Micro Devices, Inc. All rights reserved.
+ * Copyright (C) 2024 Advanced Micro Devices, Inc. All rights reserved.
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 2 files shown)


nakajee:
Please always update year in Copyright if it is not 2024.

HostLibraryTests/DecisionTree_test.cpp
```diff
@@ -369,6 +369,8 @@ TEST(DecisionTree, DecisionTreeBatch)
     using SizeInRange = Predicates::Contraction::SizeInRange;
     using Range       = Predicates::Contraction::Range;
     using And         = Predicates::And<ContractionProblem>;
+    using Key         = std::array<float, 4>;
```
nakajee, line 372: I am not sure how Key works, but Key is used at line 374. Isn't it a problem?

HostLibraryTests/DecisionTree_test.cpp
```diff
@@ -369,6 +369,8 @@ TEST(DecisionTree, DecisionTreeBatch)
...
+    using DTree       = Tree<Key, std::shared_ptr<ContractionLibrary>, std::shared_ptr<ContractionSolution>>;
```
bstefanuk, line 373: I see that previously we were pushing an array of size 4 (L413) into a container that's defined at file-scope to have size 3 (L40). Odd that other platforms aren't catching this error. However, I also notice this behaviour at L294, L313, etc. Should lines these be updated as well? Why aren't they failing too? Which line did the original error come from?

HostLibraryTests/DecisionTree_test.cpp
ellosel: The dimension of key should be greater than or equal to the number of features (or `features.size()` (however, maybe someone more familiar with solution selection could confirm).

The last test was added along with the additional feature but the `Key` was reused from previous tests where the size was 3. This caused a segfault which is unsuprising because we are trying to access the fourth element of a size three array. What may be a surprise is that this is only an issue on SLES. However, this is UB and the compiler is free to do what it wants. It is possible that this code was optimized out on the other platforms.

As for line 342, we are shadowing the `Key` definition that is at the top of the file. Perhaps we should move these aliases into each of the test cases.

HostLibraryTests/DecisionTree_test.cpp
ellosel: The important distinction between this change and the others is that in the previous test the size of the features vector is three (note that we call `push_back` three times). In this test the size of features is four. That is why it was failing prior to this change. Why it wasn't failing on other platforms (shrug... undefined behavior).

nakajee: Thanks for your comment.
Maybe better to put some comment in source code.
Do we need to change other locations using Key?

ellosel: A comment in the test code would probably help a little. We do not need to change the other `Key`s but I think I will move the aliases into the tests where they are used to avoid confusion.

ellosel:
@nakajee @bstefanuk please resolve comments.

HostLibraryTests/DecisionTree_test.cpp
```diff
@@ -2,7 +2,7 @@
  *
  * MIT License
  *
- * Copyright (C) 2022 Advanced Micro Devices, Inc. All rights reserved.
```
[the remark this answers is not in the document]
ellosel: Good catch thanks

nakajee:
Thanks for your update.
I do not have further comments for code change.
I am not sure how to confirm if this change works or not.
You ran some local test and confirmed this change worked, right?

ellosel:
> Thanks for your update. I do not have further comments for code change. I am not sure how to confirm if this change works or not. You ran some local test and confirmed this change worked, right?

In addition to local testing I ran in a manual pipeline and everything looks good. Once this is merged I will enable sles testing.

nakajee (approved):
Thanks for your comment.
Looks good.

PR #2003, opened 2024-08-07
Call min with matching types

bstefanuk:
**Summary:**

A build failure has been created in Tensile while fixing a compiler bug related to the `min` function in the LLVM project. The issue arises from a call to `min(unsigned long, int)` in `ContractionSolution.cpp`. The compiler now only defines multiple functions that don't require an implicit case, but none match this specific signature.

**Outcomes:**

This change modifies the problematic line to use an explicit cast when invoking `min` such that `min(unsigned long, unsigned long)` is called.

**Testing and Environment:**
- ROCm 6.3, Ubuntu 22, Python 3.10.
- Tested against rocBLAS PTS (fill in results)

### DIFF
--- a/Tensile/Source/lib/source/ContractionSolution.cpp
+++ b/Tensile/Source/lib/source/ContractionSolution.cpp
@@ -1871,7 +1871,7 @@ namespace Tensile
         // whichever is minimum.
         else if(pAMDGPU->skMaxCUs > 0)
         {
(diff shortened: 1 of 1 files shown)


yoichiyoshida (approved):

yxsamliu (approved):

nakajee (approved):
Looks good

PR #2004, opened 2024-08-09
Don't add fallbacks if they weren't requested

ellosel:
**Summary:**

When running TensileCreateLibrary for separate archs where no logic files are present with Hip in the name, the program fails. We need to inspect the dictionary to see if fallback is a key prior to attempting to add fallback libraries.

**Outcomes:**

Fixes a regression that prevents TensileCreateLibrary from completing when fallback logic is not available.

**Testing and Environment:**

Ran pre_checkin, unit and static tests locally. Awaiting pipeline results.


nakajee (approved):
Looks good.
I confirmed rocblas build worked with -l option.

ellosel:
> Looks good. I confirmed rocblas build worked with -l option.

Many thanks!

PR #2005, opened 2024-08-09
Catch accvgpr overflow

bstefanuk:
**Summary:**

Currently, the assembly kernel writer does not catch accvgpr overflows, which is causing failures in QA testing for incorrect Tensile (YAML) configurations. While we can always request the testing configs to be changed, Tensile should still gracefully reject configurations with accvpgr >= 256.

**Outcomes:**

1. Add overflow handling logic if to reject kernels with accvgprs >= 256
2. Update print statement with rejection reason.

**Testing and Environment:**

- Device: gfx942
- OS: RHEL 9
- HIP version: 6.2.41133-dd7f95766
- AMD clang version 18.0.0

Command(s):
```bash
./Tensile/bin/Tensile --cxx-compiler=hipcc ../fp16.yaml logs-fp16 | tee output-fp16.txt
./Tensile/bin/Tensile --cxx-compiler=hipcc ../fp32.yaml logs-fp16 | tee output-fp32.txt
./Tensile/bin/Tensile --cxx-compiler=hipcc ../bf16.yaml logs-fp16 | tee output-bf16.txt
```

### DIFF
--- a/Tensile/KernelWriterAssembly.py
+++ b/Tensile/KernelWriterAssembly.py
@@ -3193,9 +3193,14 @@ class KernelWriterAssembly(KernelWriter):
         msg = "unknown"
 
       if globalParameters["PrintSolutionRejectionReason"]:
-        printWarning("%s overflowed resources.  errorCode=%d, msg=\"%s\", vgprs=%u, sgprs=%u, accvgprs=%u" \
-          % (self.kernelName, self.overflowedResources, msg, \
-          self.vgprPool.size(), self.sgprPool.size()), self.agprPool.size())
+        printWarning(
+            f"{self.kernelName} overflowed resources. errorCode={self.overflowedResources}" \
+             ", msg='{msg}', vgprs={self.vgprPool.size()}, sgprs={self.sgprPool.size()}" 
+            + f"accvgprs={self.agprPool.size()}" if kernel["EnableMatrixInstruction"] else ""
+        )
+        # printWarning("%s overflowed resources.  errorCode=%d, msg=\"%s\", vgprs=%u, sgprs=%u, accvgprs=%u" \
+        #   % (self.kernelName, self.overflowedResources, msg, \
+        #   self.vgprPool.size(), self.sgprPool.size(), self.agprPool.size()))
       kStr += "s_endpgm // overflowed resources\n"
       kStr += ".if 0\n"
 
@@ -15403,7 +15408,7 @@ class KernelWriterAssembly(KernelWriter):
       self.overflowedResources = 1
     elif self.sgprPool.size() > self.maxSgprs:
       self.overflowedResources = 2
-    elif self.agprPool.size() > self.maxAgprs:
+    elif kernel["EnableMatrixInstruction"] and self.agprPool.size() > self.maxAgprs:
       self.overflowedResources = 7
 
     if kernel["ScheduleIterAlg"] == 2 and \
--- a/Tensile/KernelWriterAssembly.py
+++ b/Tensile/KernelWriterAssembly.py
@@ -3196,7 +3196,7 @@ class KernelWriterAssembly(KernelWriter):
         printWarning(
             f"{self.kernelName} overflowed resources. errorCode={self.overflowedResources}" \
             f", msg='{msg}', vgprs={self.vgprPool.size()}, sgprs={self.sgprPool.size()}" 
-            + f"accvgprs={self.agprPool.size()}" if kernel["EnableMatrixInstruction"] else ""
+            + f", accvgprs={self.agprPool.size()}" if kernel["EnableMatrixInstruction"] else ""
         )
       kStr += "s_endpgm // overflowed resources\n"
       kStr += ".if 0\n"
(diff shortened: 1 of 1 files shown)


nakajee:
Precisely, accvgprs > 256 (=256 is OK)

Tensile/KernelWriterAssembly.py
```diff
@@ -3186,13 +3187,15 @@ def functionSignature(self, kernel ):
...
       elif self.overflowedResources == 6:
         msg = "SIA2 better with occupancy 2"
+      elif self.overflowedResources == 7:
+        msg = "too many accvgprs"
       else:
         msg = "unknown"
 
       if globalParameters["PrintSolutionRejectionReason"]:
-        printWarning("%s overflowed resources.  errorCode=%d, msg=\"%s\", vgprs=%u, sgprs=%u" \
+        printWarning("%s overflowed resources.  errorCode=%d, msg=\"%s\", vgprs=%u, sgprs=%u, accvgprs=%u" \
```
nakajee, line 3196: I am thinking it might be better to print accvgprs only when kernel["EnableMatrixInstruction"] is true.

nakajee: This will not cause an error even if kernel["EnableMatrixInstruction"] is false, but it is unnecessary information.

Tensile/KernelWriterAssembly.py
```diff
@@ -15400,6 +15403,8 @@ def functionSuffix(self, kernel):
       self.overflowedResources = 1
     elif self.sgprPool.size() > self.maxSgprs:
       self.overflowedResources = 2
+    elif self.agprPool.size() > self.maxAgprs:
```
nakajee, line 15406: same here. It might make more sense to check kernel["EnableMatrixInstruction"]
like
   elif kernel["EnableMatrixInstruction"]  and self.agprPool.size() > self.maxAgprs:

Tensile/KernelWriterAssembly.py
```diff
@@ -3186,13 +3187,15 @@ def functionSignature(self, kernel ):
...
+        msg = "too many accvgprs"
       else:
         msg = "unknown"
 
       if globalParameters["PrintSolutionRejectionReason"]:
-        printWarning("%s overflowed resources.  errorCode=%d, msg=\"%s\", vgprs=%u, sgprs=%u" \
+        printWarning("%s overflowed resources.  errorCode=%d, msg=\"%s\", vgprs=%u, sgprs=%u, accvgprs=%u" \
           % (self.kernelName, self.overflowedResources, msg, \
-          self.vgprPool.size(), self.sgprPool.size()))
+          self.vgprPool.size(), self.sgprPool.size()), self.agprPool.size())
```
nakajee, line 3198: Looks like self.agprPool.size() is put incorrectly. we need 2 )) at the right.

Tensile/KernelWriterAssembly.py
```diff
@@ -3186,13 +3187,17 @@ def functionSignature(self, kernel ):
...
       else:
         msg = "unknown"
 
       if globalParameters["PrintSolutionRejectionReason"]:
-        printWarning("%s overflowed resources.  errorCode=%d, msg=\"%s\", vgprs=%u, sgprs=%u" \
-          % (self.kernelName, self.overflowedResources, msg, \
-          self.vgprPool.size(), self.sgprPool.size()))
+        printWarning(
+            f"{self.kernelName} overflowed resources. errorCode={self.overflowedResources}" \
+            f", msg='{msg}', vgprs={self.vgprPool.size()}, sgprs={self.sgprPool.size()}"
```
nakajee, line 3198: Don't we need + here?
Could you double-check if the message format here is same as before change (except for acc part)?

bstefanuk: The plus allows a simple inline if/else to be used to control whether accvgprs=... is printed. I will verify for good measure.

bstefanuk: @nakajee confirmed the following output with the latest commit,
```bash
# after
UserWarning: Cijk_Alik_Bljk_SB_MT16x16x64_MI16x16x4x1_... overflowed resources. errorCode=0, msg='too many accvgprs', vgprs=256, sgprs=66, accvgprs=0
# before
UserWarning: Cijk_Alik_Bljk_SB_MT16x16x64_MI16x16x4x1_... overflowed resources.  errorCode=0, msg="too many accvgprs", vgprs=256, sgprs=66
```

nakajee: Thanks for confirmation.
I thought we needed "+" to concatenate 2 strings (I am not very good at Python though)...
I am OK as long as you confirmed it worked as expected.

nakajee:
I am not sure why many CI tests are failing....
I do not think this is related to the change in this PR....
  TypeError: parseLibraryLogicFile() takes 1 positional argument but 152 were given
Do you have any idea?

bstefanuk:
> TypeError: parseLibraryLogicFile() takes 1 positional argument but 152 were given
> Do you have any idea?

It appears as though CI is attempting to run in single-threaded mode, which is currently unsupported due to the observed bug.

ellosel:
> > TypeError: parseLibraryLogicFile() takes 1 positional argument but 152 were given
> > Do you have any idea?
> 
> It appears as though CI is attempting to run in single-threaded mode, which is currently unsupported due to the observed bug.

Could it be that joblib is unavailable and it is falling back on a broken feature? Its better to fail IMO because serial would be very slow.

nakajee (approved):
The change looks good.
However, I am not sure how to make CI test work.

PR #2007, opened 2024-08-13
Add logic to remove temporary build artifacts

ellosel:
**Summary:**

The Tensile disk usage can reach up to hundreds of GBs in some cases. This work adds new behavior that removes temporary build artifacts and generated code during the build process to keep the disk space high-water mark down. The old behavior can be enabled by using the new option `--keep-build-tmp`.
 
**Outcomes:**

Tensile hard disk usage is reduced by as much as 80%.

**Notable changes:**

`--keep-build-tmp` option was added.

**Testing and Environment:**

Routine CI pipelines and equivalent testing run locally

### DIFF
--- a/Tensile/TensileCreateLibrary.py
+++ b/Tensile/TensileCreateLibrary.py
@@ -96,7 +96,7 @@ def processKernelSource(kernel, kernelWriterSource, kernelWriterAssembly):
     return (err, src, header, kernelName, filename)
 
 
-def getAssemblyCodeObjectFiles(kernels, kernelWriterAssembly, outputPath):
+def getAssemblyCodeObjectFiles(kernels, kernelWriterAssembly, outputPath, removeTemporaries):
     destDir = ensurePath(os.path.join(outputPath, "library"))
     asmDir = kernelWriterAssembly.getAssemblyDirectory()
     assemblyKernels = list([k for k in kernels if k["KernelLanguage"] == "Assembly"])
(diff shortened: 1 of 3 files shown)


bstefanuk:
> Tensile hard disk usage is reduced by as much as 80%.

@ellosel how difficult would it be to quantify this for a specific use-case, say, gfx942?

ellosel:
> > Tensile hard disk usage is reduced by as much as 80%.
> 
> @ellosel how difficult would it be to quantify this in MB/GB for a specific architecture, say, gfx942?

Not too difficult - here are some disk size samples for gfx 900 before (looking at the size of the Tensile project directory including build_tmp and tensile output dir the initial size is 90 MB):

```
88M  .
92M  .
282M  .
475M  .
661M  .
665M  .
676M  .
688M  .
697M  .
703M  .
767M  .
770M  .
```

and after:

```
90M  .
93M  .
116M  .
140M  .
206M  .
208M  .
211M  .
216M  .
217M  .
220M  .
225M  .
```

Which is a 70% reduction if you include the initial 90M. It is closer to an 80% if you don't include the 90M.

babakpst (approved):

Tensile/TensileCreateLibrary.py
```diff
@@ -164,6 +166,10 @@ def getAssemblyCodeObjectFiles(kernels, kernelWriterAssembly, outputPath):
                     print(err.output)
                     raise
 
+                #if removeTemporaries:
```
ellosel, line 169: This doesn't have a significant impact on the high-water mark and is breaking when targeting gfx90a.

PR #2008, opened 2024-08-14
Remove OCL backend

pghysels:
resolves #840

**Summary:**

Remove the OCL backend, which no longer works, and can probably not be made to work anymore.


Tensile/Source/lib/CMakeLists.txt
```diff
@@ -73,14 +73,6 @@ if(TENSILE_USE_HIP)
         )
 endif()
 
-if(TENSILE_USE_OPENCL)
```
nakajee, line 76: I found one more "if(TENSILE_USE_OPENCL)" in this file.
Better to remove it as well.

Tensile/KernelWriterSource.py
```diff
@@ -111,22 +85,13 @@ def __init__( self, kernelMinNaming, kernelSerialNaming ):
   # Open String
   ##############################################################################
   def openString(self, kernel):
-    kStr = ""
-    if self.language == "OCL":
-      kernelName = self.getKernelName(kernel)
-      kStr += "\n"
-      kStr += "std::string %s_src_%u = \"" % (kernelName, self.stringIdx)
-    return kStr
+    return ""
```
nakajee, line 88: openString and closeString actually does nothing for both Asm and Source.
Maybe we can remove the function itself.

nakajee:
I found "Ocl" in .jenkins/common.groovy.
Probably better to update it as well.

nakajee:
maybe better to grep "OpenCL" as well as "ocl"
I found the following line in Signature.py.
`        kStr += "    .language:                   %s%s" % ("OpenCL C", writer.endLine)
`
I am not sure if we need to change this or not.

pghysels:
Sorry, I thought I did grep for both "ocl" and "opencl", case insensitive.

nakajee:
OK. Thanks for your update. I would like to wait for comments from @AlexBrownAMD

pghysels:
One of these tests is still pending, is there anything I need to do here? I can't tell from the testing details what is wrong.

AlexBrownAMD:
CI didn't have any log files so not sure if they were deleted or the tests didn't run. I just kicked off a rebuild so we can check back in a bit

AlexBrownAMD:
Looks like the only failing pipeline is for gfx1151, and the PR to enable 1151 support just got merged. So you could pull latest changes to your branch and rerun CI to make sure those tests pass as well, otherwise I think this change is safe to merge.

pghysels:
It says I'm not authorized to merge this PR.

PR #2014, opened 2024-08-23
Linearize asm register search

bstefanuk:
**Summary:**

Profiling revealed an $O(n^2)$ implementation that can be converted into a linear $O(n)$ implementation in `AsmRegisterpool::findFreeRange`, which currently consumes 17% of the runtime of TensileCreateLibrary (under the execution conditions below). This PR updates this function to use a two-pointer algorithm that linearizes the time complexity while also avoiding eager evaluation of register availability.

**Outcomes:**

Runtime reduction when benchmarked against the current implementation is:
```
Before change: 32 jobs, gfx90a:
    7744.10s user 641.74s system 1993% cpu 7:00.56 total <--- before
After change: 32 jobs, gfx90a:
    6546.97s user 648.71s system 1908% cpu 6:17.01 total <--- after
```
Which cuts 43.55 seconds (10.4%) off the wall time of the following invocation:
```
time Tensile/bin/TensileCreateLibrary \
              ../rocBLAS/library/src/blas3/Tensile/Logic/asm_full \
              _build_tcl \
              HIP \
              --merge-files \
              --separate-architecture \
              --lazy-library-loading \
              --no-short-file-names \
              --code-object-version=default \
              --cxx-compiler=amdclang++ \
              --jobs=32 \
              --library-format=msgpack \
              --architecture='gfx90a' \
              --verbose=1 | tee results-jobs-$(date -Iseconds).log
```

**Notable changes:**

The internal call to `isRangeAvailable` has been removed from `findFreeRange`, however it is still used elsewhere in the codebase.

**Testing and Environment:**

Standard Tensile testing; Ubuntu 22.01, ROCm 6.3

### DIFF
--- a/Tensile/AsmRegisterPool.py
+++ b/Tensile/AsmRegisterPool.py
@@ -27,6 +27,7 @@ from .Utils import roundUpToNearestMultiple
 from .AsmUtils import inst, vgpr, sgpr
 import traceback
 from enum import Enum
+from typing import Optional
 
 ################################################################################
 # RegisterPool
@@ -181,21 +182,49 @@ class RegisterPool:
 
     return True
 
-  def findFreeRange(self, size, alignment, preventOverflow=-1, wantedStatus=Status.Available):
+
+  def findFreeRange(
+    self, 
+    size: int, 
+    alignment: int, 
+    preventOverflow: int=-1, 
+    wantedStatus: Status=Status.Available
+  ) -> Optional[int]:
+    """Find a free range of registers of a given size, alignment, and status from the register pool.
+
+    Args:
+        size: The required size of the register block.
+        alignment: The alignment requirement for the block. Each block must start at an index
+            that is a multiple of this value.
+        preventOverflow: Flag to prevent overflow. If set to -1, it uses the default value.
+        wantedStatus: The status of the block to be found.
+
+    Returns:
+        1. The starting index of the block if found. 
+        2. None if preventOverflow is set and no block is found.
+        3. The start of the last available block rounded up to the alignment if preventOverflow is
+            not set and no block is found. 
+    """
     if preventOverflow == -1:
       preventOverflow = self.defaultPreventOverflow
 
-    for i in range(len(self.pool)+1):
-      if i % alignment != 0:
-        continue
-      if self.isRangeAvailable(i, size, preventOverflow, wantedStatus):
-        return i
+    left, right = 0, 0
+    while right < len(self.pool):
+      if self.pool[right].status == wantedStatus:
+        blockSize = right - left + 1
+        if blockSize >= size:
+          return left
+        right += 1
+      else:
+        left += alignment
+        right = left
 
     if preventOverflow:
       return None
     else:
       loc = self.startOfLastAvailableBlock()
       return roundUpToNearestMultiple(loc, alignment)
+
 
   def checkOutAt(self, start, size, tag, preventOverflow, wantedStatus = Status.Available):
     if preventOverflow:
--- a/Tensile/AsmRegisterPool.py
+++ b/Tensile/AsmRegisterPool.py
@@ -219,11 +219,7 @@ class RegisterPool:
         left += alignment
         right = left
 
-    if preventOverflow:
-      return None
-    else:
-      loc = self.startOfLastAvailableBlock()
-      return roundUpToNearestMultiple(loc, alignment)
+    return None if preventOverflow else left
 
 
   def checkOutAt(self, start, size, tag, preventOverflow, wantedStatus = Status.Available):
--- a/Tensile/Tests/unit/test_AsmRegisterPool.py
+++ b/Tensile/Tests/unit/test_AsmRegisterPool.py
@@ -142,11 +142,12 @@ def test_findFreeRange_noOverflowPastPoolLength():
     assert out_new == out_old
     assert out_new == 0
 
+
 def test_findFreeRange_noOverflowPastPoolLength2():
     poolSize = 12
     vgprPool = RegisterPool(poolSize, "v", defaultPreventOverflow=False, printRP=0)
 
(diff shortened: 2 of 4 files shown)


Tensile/AsmRegisterPool.py
```diff
@@ -181,22 +182,50 @@ def isRangeAvailable(self, start, size, preventOverflow=-1, wantedStatus=Status.
...
     if preventOverflow == -1:
       preventOverflow = self.defaultPreventOverflow
 
-    for i in range(len(self.pool)+1):
-      if i % alignment != 0:
-        continue
-      if self.isRangeAvailable(i, size, preventOverflow, wantedStatus):
-        return i
+    left, right = 0, 0
+    while right < len(self.pool):
```
nakajee, line 212: What is the main cause of bad performance of the original code?
i % alignment?
calling isRangeAvailable?

So far, your code looks like not handling preventOverflow.
If i%alignment is the primary cause, we can just call isRangeAvailable here.

bstefanuk: Calling `isRangeAvailable` is the cause of bad performance. In its current form, it uses nested for loops to achieve the same outcome as the solution presented here, which uses a single pass through the register pool (`self.pool`).

I do believe we are appropriately handling preventOverflow because the loop will exit of the right index exceeds the pool size.

nakajee: Difference is this part

    if preventOverflow and end > len(self.pool):
      return False

    end = min(end, len(self.pool))

if preventOverflow == false, isRangeAvailable does not check start + size.
It returns true as long as from start to end = min(end, len(self.pool)) is available.

We can add extra if condition to follow this logic.
(however, it does not make much sense we do not check size....)

bstefanuk: isRangeAvailable will continue to check start+size if preventOverflow==False, so long as start+size isn't larger than the pool. In the case that `end <= len(self.pool)`, we loop over every entry from start to end, always starting from the beginning of an alignment.

In the code update, we're effectively collecting the alignment step in the same iteration where we search for the range. If at anytime the block size matches the requested size, we return the left index. 

Either way, I will write some tests to ensure the behaviour is the same as before.

nakajee: I am OK with your change, but what I am saying is isRangeAvailable seems incorrect in that case.

nakajee:
By the way, is the original implementation really O(n2)?
I think both before and after are O(n) (O(n*size))

Tensile/AsmRegisterPool.py
```diff
@@ -181,22 +182,50 @@ def isRangeAvailable(self, start, size, preventOverflow=-1, wantedStatus=Status.
...
+        right += 1
+      else:
+        left += alignment
+        right = left
 
     if preventOverflow:
       return None
     else:
       loc = self.startOfLastAvailableBlock()
       return roundUpToNearestMultiple(loc, alignment)
```
ellosel, line 226:
```suggestion
  return left - alignment if self.pool[left].status == wantedStatus else len(self.pool) - 1
```

nakajee: When we need left - alignment?
I do not come up with any scenario.

ellosel: > When we need left - alignment? I do not come up with any scenario.

I turns out we can use the existing branch as is so technically we don't need it but if we rewrite this else logic we would need it in the case where we don't prevent overflow and we are in that last branch because we increment left too far in some cases when exiting the while loop.

nakajee: There are 2 cases in above while loop.

(1) if self.pool[right].status == wantedStatus: is true and right reached pool size.
 in this case, left is not incremented. We do not need left -alignment
(2) else case (means status != wantedStatus in one of element between left-alignment to left -1
 we should not need left-alignment since one of the element is not wantedStatus

ellosel: For case (2) in the example above we would have incremented past the end of the list so we would need to step back to the beginning of the last two elements and in that case we would find that the first element is a 1 so we would want to return that index.

nakajee: lets say, 
size = 11
alignment = 4
self.pool[9].status != wantedStatus .

In this case, 
it hits else condition at 9 and left=right=12.
There is no possibility 12 -alignment = 8 is OK because it already hit != wantedStatus somewhere before 12

nakajee: > pool = [ 0 0 0 0 0 0 0 0 1 1]
> 
> For case (2) in the example above (with the corrected pool) we would have incremented past the end of the list so we would need to step back to the beginning of the last two elements. We would still need to traverse the remaining elements to be equivalent to the previous code.

If states is available, left is not incremented. It should not be left -alignment

nakajee: and just checking self.pool[left-alignment].status is not enough

ellosel: Yes you are correct that it won't work (and may not be necessary) in all cases. The purpose of the comment was to illustrate that if we need to replace the code in the else block for correctness or performance that we should consider cases where we need to correct left. Braden and I are both convinced that the existing code is fine as-is so you can ignore the suggestion entirely.

nakajee: OK.

bstefanuk:
@nakajee Yes, taking another look regarding the time complexity, I agree that the original implementation is O(n*size), however I do still think the new implementation is O(n) since for each iteration we are only conducting constant time operations, as opposed to potentially checking `size` elements.

Either way it speeds up the build, so it's that's a plus.

nakajee (approved):
Thanks for your comment.
I do not have further comment.
Looks good.

PR #2017, opened 2024-08-27
Refactor kernel source and header creation

bstefanuk:
**Summary:**

`buildKernelSourceAndHeaderFiles` in TensileCreateLibrary is a large function that has multiple responsibilities. This PR aims to refactor logic into single-responsibility functions, improve the clarity of the code, and add unit tests.

**Outcomes:**

- `buildKernelSourceAndHeaderFiles` renamed to `generateKernelSourceAndHeaderFiles`.
- Logic related to the control of which files should be generated has been moved into its own function: `collectFilesToWrite`.
- The renamed `generateKernelSourceAndHeaderFiles` is now parameterized only by the kernel lookup (hash) table returned by `collectFilesToWrite`.
- The combined usage of the two new functions is *identical* in behaviour to the previous function -- confirmed via unittests that were written before the refactor.

**Testing and Environment:**

- `collectFilesToWrite` is only tested end-to-end within larger tests, which matches the existing behaviour of `buildKernelSourceAndHeaderFiles`.
- Standard unit testing suite, static analysis, and TensileManifest.txt consistency before and after change.
- Environment: ROCm 6.3, Ubuntu 22.04


Tensile/TensileCreateLibrary.py
```diff
@@ -74,12 +74,15 @@
 TENSILE_MANIFEST_FILENAME = "TensileManifest.txt"
 TENSILE_LIBRARY_DIR = "library"
 
+ProcKernResult = List[Tuple[int, str, str, str, Optional[str]]]
```
ellosel, line 77: Should we collocate this with the function?

Tensile/TensileCreateLibrary.py
```diff
@@ -484,30 +487,32 @@ def prepAsm(
     os.chmod(assemblerFileName, 0o777)
 
 
-################################################################################
-def buildKernelSourceAndHeaderFiles(results, outputPath):
-    """
-    Logs errors and writes appropriate info to kernelSourceFile and kernelHeaderFile.
+def buildKernelSourceAndHeaderFiles(
+    results: ProcKernResult, outputPath, lazyLoading: bool, mergeFiles: bool, numMergedFiles: int
```
ellosel, line 491: outputPath doesn't have a type?

Tensile/TensileCreateLibrary.py
```diff
@@ -484,30 +487,32 @@ def prepAsm(
...
+) -> Tuple[List[str], Dict[str, int]]:
+    """Logs errors and writes appropriate info to kernel source and header files.
```
ellosel, line 493: Logs errors for what? Are the source files and header files all cpp files?

Tensile/TensileCreateLibrary.py
```diff
@@ -484,30 +487,32 @@ def prepAsm(
...
 
     Arguments:
-      results:              list of (err, src, header, kernelName, filename)
-      outputPath:           path to source directory
-      kernelsWithBuildErrs: Dictionary to be updated with kernels that have errors
-      kernelSourceFile:     File to write source data to
-      kernelHeaderFile:     File to write header data to
+        results: List of (err, src, header, kernelName, filename)
```
ellosel, line 496: The name `results` may confuse. It suggests to me that this is the result of the function rather than solely input.

Tensile/TensileCreateLibrary.py
```diff
@@ -518,53 +523,45 @@ def buildKernelSourceAndHeaderFiles(results, outputPath):
         kernelsToWrite.append((err, src, header, kernelName))
```
ellosel, line 523: Seems like we are recreating `results` aside from dropping filename.

Tensile/TensileCreateLibrary.py
```diff
@@ -518,53 +523,45 @@ def buildKernelSourceAndHeaderFiles(results, outputPath):
...
-        else:
-            filesToWrite[os.path.join(os.path.normcase(outputPath), kernelName)].append(
-                (err, src, header, kernelName)
-            )
+            key = pathJoin(filename)
+        elif mergeFiles:
+            kernelSuffix = str(validKernelCount % numMergedFiles) if numMergedFiles > 1 else ""
+            key = pathJoin(f"Kernels{kernelSuffix}")
+
+        filesToWrite[key].append((err, src, header, kernelName))
```
ellosel, line 533: We might consider making this block of code a function (as well as the block above and below). We could call this `makeFilesToWrite` the block above `makeKernelsToWrite`. Maybe we define them inline like the functions below thought that makes them difficult to test.

Tensile/TensileCreateLibrary.py
```diff
@@ -518,53 +523,45 @@ def buildKernelSourceAndHeaderFiles(results, outputPath):
...
-                kernelHeaderFile.write(header)
+        with open(f"{filename}.h", "w", encoding="utf-8") as hdrFile, open(
+            f"{filename}.cpp", "w", encoding="utf-8"
+        ) as srcFile:
+
+            writeHeader(hdrFile)
+            writeSource(srcFile, filename)
+            for _, src, header, _ in kernelList:
+                srcFile.write(src)
+                hdrFile.write(header)
```
ellosel, line 564: Function?

Tensile/TensileCreateLibrary.py
```diff
@@ -754,6 +751,7 @@ def writeKernelHelpers(
 ################################################################################
 # Write Solutions and Kernels for BenchmarkClient or LibraryClient
 ################################################################################
+@profile
```
ellosel, line 754: Should we put this back on TCL?

Tensile/TensileCreateLibrary.py
```diff
@@ -1662,7 +1666,7 @@ def writeMasterFile(
 ################################################################################
 # Tensile Create Library
 ################################################################################
-@profile
+# @profile
```
ellosel, line 1669: Restore?

Tensile/Utils.py
```diff
@@ -48,7 +48,7 @@ class ProgressBar:
         createTime: The timestamp when the progress bar was created.
         message: The message displayed alongside the progress bar.
     """
-    def __init__(self, maxValue: int, desc: str, width=40):
+    def __init__(self, maxValue: int, desc: str, width=25):
```
ellosel, line 51: Was this supposed to be included in this PR?

Tensile/TensileCreateLibrary.py
bstefanuk: And we're appending to two objects, one of which we only use to check that it's not null. I'll remove it.

Tensile/TensileCreateLibrary.py
bstefanuk: "Collects" is probably a better term here. Note the Return section in the doc-string.

Tensile/Utils.py
bstefanuk: No, it was not.

Tensile/TensileCreateLibrary.py
bstefanuk: The PR description has been updated with details about the new function.

Tensile/TensileCreateLibrary.py
bstefanuk: Done, however, it's called `collectFilesToWrite`

Tensile/TensileCreateLibrary.py
bstefanuk: I did not change this, but instead added sematic type-hints to the functions to help guide the developer.

Tensile/TensileCreateLibrary.py
```diff
@@ -484,91 +486,101 @@ def prepAsm(
...
-    kernelsWithBuildErrs: Dict[str, int] = {}
     filesToWrite = collections.defaultdict(list)
+    kernelsWithBuildErrs = collections.defaultdict(int)
     validKernelCount = 0
-    for err, src, header, kernelName, filename in results:
 
-        # Keep track of kernels with errors
+    for err, src, header, kernelName, filename in results:
         if err:
             kernelsWithBuildErrs[kernelName] = err
```
ellosel, line 523: Seems a little awkward to leave this logic in this function. Would be nice to separate into another function such as:

```
def kernelsWithBuildErrors(results: List[ProcessedKernelResult]):
  return {kernelName: err for err, *, *,  kernelName, * in results if err}
```

At this point, do we even need a function?

Tensile/TensileCreateLibrary.py
```diff
@@ -484,91 +486,98 @@ def prepAsm(
...
     validKernelCount = 0
-    for err, src, header, kernelName, filename in results:
-
-        # Keep track of kernels with errors
-        if err:
-            kernelsWithBuildErrs[kernelName] = err
 
-        # Don't create a file for empty kernels
+    for err, src, header, kernelName, filename in results:
         if len(src.strip()) == 0:
```
ellosel, line 521: From the pep 8 programming reccomendations:

> For sequences, (strings, lists, tuples), use the fact that empty sequences are false:

```
# Correct:
if not seq:
if seq:

# Wrong:
if len(seq):
if not len(seq):
```

Tensile/Tests/unit/test_TensileCreateLibrary.py
```diff
@@ -729,11 +729,11 @@ def test_processKernelSource(setupSolutionsAndKernels):
     assert results == expected, "Assembly files shouldn't have any header or source content"
 
 
-def test_buildKernelSourceAndHeaderFiles():
+def test_generateKernelSourceAndHeaderFiles_checkBuildErrorsAsmKernels():
```
bstefanuk, line 732: Remove this test after most recent refactor.

ellosel (approved):

PR #2018, opened 2024-08-31
Add logic to keep build_tmp when running Tensile

ellosel:
**Summary:**

Changes were added to TensileCreateLibrary to remove (or not) the `build_tmp` directory. However, the same changes were overlooked in Tensile. This PR adds the logic necessary to remove (default) or keep `build_tmp` via the `--keep-build-tmp` option.

**Outcomes:**

Temporary source files and build artifacts are available for inspection after running Tensile command.

**Notable changes:**

Adds `--keep-build-tmp` option to Tensile application.

**Testing and Environment:**

Local testing ran using a config file provided by @nakajee. After reviewing tensile ouput directory, temporary build files are removed (or not) depending on whether option is set.

### DIFF
--- a/Tensile/Common.py
+++ b/Tensile/Common.py
@@ -2390,6 +2390,9 @@ def assignGlobalParameters( config ):
     else:
       tPrint(3, " %24s: %8s (unspecified)" % (key, defaultValue))
 
+  if "KeepBuildTmp" in config:
+    globalParameters["KeepBuildTmp"] = config["KeepBuildTmp"]
+
   globalParameters["ROCmPath"] = "/opt/rocm"
   if "ROCM_PATH" in os.environ:
     globalParameters["ROCmPath"] = os.environ.get("ROCM_PATH")
--- a/Tensile/Common.py
+++ b/Tensile/Common.py
@@ -2391,7 +2391,8 @@ def assignGlobalParameters( config ):
       tPrint(3, " %24s: %8s (unspecified)" % (key, defaultValue))
 
   if "KeepBuildTmp" in config:
-    globalParameters["KeepBuildTmp"] = config["KeepBuildTmp"]
+    globalParameters["KeepBuildTmp"] = config["KeepBuildTmp"] 
+
 
   globalParameters["ROCmPath"] = "/opt/rocm"
   if "ROCM_PATH" in os.environ:
--- a/Tensile/Common.py
+++ b/Tensile/Common.py
@@ -250,6 +250,7 @@ globalParameters["SupportedISA"] = [(8,0,3),
                                     (11,5,1),
                                     (12,0,0), (12,0,1)] # assembly kernels writer supports these architectures
 
+globalParameters["KeepBuildTmp"] = True                           # remove build artifacts during the build process and build_tmp after build completes
 globalParameters["CleanupBuildFiles"] = False                     # cleanup build files (e.g. kernel assembly) once no longer needed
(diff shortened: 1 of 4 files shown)


ellosel:
@nakajee should the yaml or command line option take precedence?

nakajee:
> @nakajee should the yaml or command line option take precedence?

Not sure. Any other parameters for reference?

Tensile/Common.py
```diff
@@ -2390,6 +2390,10 @@ def assignGlobalParameters( config ):
     else:
       tPrint(3, " %24s: %8s (unspecified)" % (key, defaultValue))
 
+  if "KeepBuildTmp" in config:
+    globalParameters["KeepBuildTmp"] = config["KeepBuildTmp"]
```
nakajee, line 2394: Don't we need to initialize globalParameters["KeepBuildTmp"]?
I thought we needed to add some init code at the beginning of Common.py (as other globalParameters).

ellosel: For this feature it is both a command line option and a config file option. What this means is that setting a default in common will break the ability to fall back on a command line option if it isn't set in the config file. If we want to only specify this in a config file we could set a default as you suggested. An example from your config file would be PrintSolutionRejectionReason. This has no command line option so it needs a default in common.

ellosel:
> > @nakajee should the yaml or command line option take precedence?
> 
> Not sure. Any other parameters for reference?

There aren't. For this to work, we have to give precedence to the yaml file. If it isn't specified in the file then we can fall back to the command line option (which will default to removing temporaries).

nakajee:
> > > @nakajee should the yaml or command line option take precedence?
> > 
> > 
> > Not sure. Any other parameters for reference?
> 
> There aren't. For this to work, we have to give precedence to the yaml file. If it isn't specified in the file then we can fall back to the command line option (which will default to removing temporaries).

OK.
The only thing is it is a little difficult to recognize this global parameter without having it in init part.
Maybe better to put some comment for this parameter?
Or, any better idea?

ellosel:
> > > > @nakajee should the yaml or command line option take precedence?
> > > 
> > > 
> > > Not sure. Any other parameters for reference?
> > 
> > 
> > There aren't. For this to work, we have to give precedence to the yaml file. If it isn't specified in the file then we can fall back to the command line option (which will default to removing temporaries).
> 
> OK. The only thing is it is a little difficult to recognize this global parameter without having it in init part. Maybe better to put some comment for this parameter? Or, any better idea?

Another option is to drop the command line option and only support setting it via the config file. Then we could move it all into common. That said, there are a few others like this in the Tensile main.

Tensile/Common.py
```diff
@@ -250,6 +250,7 @@ class DeveloperWarning(Warning):
                                     (11,5,1),
                                     (12,0,0), (12,0,1)] # assembly kernels writer supports these architectures
 
+globalParameters["KeepBuildTmp"] = True                           # remove build artifacts during the build process and build_tmp after build completes
```
nakajee, line 253: Maybe better to update the comment.
What you described is the behavior of False case.
By the way, looks like we have similar global parameter option just below this (CleanupBuildFiles).
Any difference?

ellosel: Sure. As for the other option. I would like to remove it and the corresponding code. It has a similar effect but seems unsafe.

nakajee:
The change looks good.
Please update year in each header comment (as I said in another PR).

nakajee (approved):
Thanks for your update.
Looks good.
We have already confirmed CI test passed.

PR #2019, opened 2024-09-04
Cache asm capabilities for reuse

ellosel:
**Summary:**

This PR adds an option to populate a yaml file containing the ASM capabilities and if it exists to re-use it. 

**Outcomes:**

For short, iterative work using tensile this can significantly reduce the runtime. Further, when running the precheckin tests (i.e. `tox -e ci`) with this option enabled the runtime is reduce from 68 minutes to 56 minutes.

**Notable changes:**

Adds `--asm-cache <path to file>` to enable caching asm capabilities.

**Testing and Environment:**

Tested locally in ub22 image. Running CI pipelines.


nakajee:
Please do not forget to update year in each file.

ellosel:
> Please do not forget to update year in each file.

Will do thanks.

nakajee:
I am not quite sure about the intention of this change.
Does this mean calling GetAsmCaps is slow?

nakajee:
Looks like CI tests are failing.
Please fix it.

ellosel:
> I am not quite sure about the intention of this change. Does this mean calling GetAsmCaps is slow?

Yes it is relatively slow. It depends on the context as always.

ellosel:
> Looks like CI tests are failing. Please fix it.

Working on it now. It's a bit complicated to get this working for the unit tests.

ellosel:
@nakajee Tests are passing.

Tensile/LibraryIO.py
```diff
@@ -410,3 +413,49 @@ def createLibraryLogicList(schedulePrefix, architectureName, deviceNames, logicT
...
+def writeAsmCapsCache(cacheFile: str, data: dict):
+    """
+    Writes an asm cache to disk.
+
+    Arguments:
+      cacheFile: User spedified file to write the cache.
+      data:      dict containg the computed cache.
+    """
+    newcache = {", ".join(map(str, k)): v for k, v in data.items()}
+    writeYAML(cacheFile, newcache)
```
cgmb, line 461: Missing a newline.

cgmb, line 461: Would JSON be appropriate here? Last I checked, the json.load was significantly faster than yaml.load. I suppose the loading time is probably not significant if this is a small file.

ellosel: I 100% agree but we would have to take on another dependency and loading time is negligible for these files. I have looked at using json for the logic files and it does help but when we first started measuring a few months ago the cost was negligible compared to other operations. That may no longer be the case given the current state of the library. We are now using CSafeLoader which also improved yaml load times.

ellosel: Good catch thanks!

Tensile/Tensile.py
```diff
@@ -131,7 +131,8 @@ def splitExtraParameters(par):
     argParser.add_argument("--client-build-path", default=None)
     argParser.add_argument("--client-lock", default=None)
     argParser.add_argument("--prebuilt-client", default=None)
-
+    argParser.add_argument("--asm-cache", dest="AsmCacheFile", action="store", type=str, \
+        help="Path to ASM cache YAML file. If it does not exist, generate the cache. If it does exist, use the cache file")
```
bstefanuk, line 135: Does making this user configurable add unnecessary complexity? Should the ASM cache simply always be written to a specific file in the build directory? If not, what benefit/use case does this offer?

Tensile/LibraryIO.py
```diff
@@ -410,3 +412,38 @@ def createLibraryLogicList(schedulePrefix, architectureName, deviceNames, logicT
 
     data.append(logicTuple[7])
     return data
+
+
+def initAsmCapsCache(cacheFile: str) -> Optional[dict]:
+    """
+    If requested and exists, reads an asm cache from disk.
```
bstefanuk, line 419: I've never loved that we shorten "assembler capabilities" to AsmCaps, as (imo) it obscures the target. However, I do understand that it can be overly verbose to type it all out. Perhaps updating this doc-string to say,

> If requested and exists, reads the cached assembler capabilities (asm caps) from disk.

Will help provide clarity to Tensile newcomers?

Tensile/Common.py
```diff
@@ -2495,7 +2500,8 @@ def assignGlobalParameters( config ):
   if "IgnoreAsmCapCache" in config:
     globalParameters["IgnoreAsmCapCache"] = config["IgnoreAsmCapCache"]
     
-  globalParameters["AsmCaps"] = {}
+  globalParameters["CacheAsmCaps"] = True if capabilitiesCache is not None else False
```
bstefanuk, line 2503: From the way I understand this change, this will cache an empty dict if the cache file doesn't exist on disk since `{}` will be returned, which is not none. Is this the intended behaviour?

bstefanuk (commented):
Few comments regarding behaviour and styling, but overall looks great.

Tensile/Tensile.py
ellosel: I went about it that way initially. For me it was a question of where do we want this file? Do we put it in build_tmp? What if we delete build_tmp? It occurred to me that by the option this way the user can decide where they want the file, its name etc.

Tensile/Common.py
ellosel: If a user provides an asm cache file, but that file doesn't exists that means I want a cache but I don't yet have one (i.e. a clean run of Tensile). So when you reach this point in the code `capabilitiesCache` would be `{}` and `bool(capabilitiesCache0 == False` so we would return false even though we really want to cache the asm caps. That is why we set it to None for the case where we do not want a cache and check for `None` i.e. an it is an `Optional[Dict]`.  If the optional is engaged that means we want a cache if not we don't. Maybe there is a more pythonic way of checking if an optional is engaged.

bstefanuk (approved):

Tensile/LibraryIO.py
ellosel: Fixed

nakajee (approved):
Looks good

PR #2022, opened 2024-09-13
use venv for Tensile create on Linux

TorreZuk:
* if found use virtual env on Linux for Tensile create command match what has been done on Windows
* Utilizes python updates added during initial install, e.g. rocBLAS build

### DIFF
--- a/Tensile/cmake/TensileConfig.cmake
+++ b/Tensile/cmake/TensileConfig.cmake
@@ -216,7 +216,7 @@ function(TensileCreateLibraryFiles
   endif()
 
   set(CommandLine ${Script} ${Options} ${Tensile_LOGIC_PATH} ${Tensile_OUTPUT_PATH} HIP)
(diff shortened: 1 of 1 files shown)


TorreZuk:
@ellosel I first looked at rocBLAS side but as we just call directly the cmake function TensileCreateLibraryFiles this is simpler and it still is the client who sets up the vars/hooks to enable the virtual env use.   It also matches what was required and in use windows, so for now think it should suffice unless there are known user conflicts.   I assume any cmake defined virtualenv (implicit function argument) would be valid for use here, otherwise it could be a new argument to TensileCreateLibraryFiles.

TorreZuk:
If this doesn't cause problems would like to merge as need to cherry pick to 6.3 branch to fix installation on new OS.

ellosel (approved):
Assuming this was tested in an environment where joblib isn't available on the system and only through the venv, I am good with this change.

PR #2040, opened 2024-11-01
Use hipcc.bat as a compiler on windows, instead of the perl script

chinmaydd:
**Summary:** Invoke "hipcc.bat" script instead of "perl hipcc". Invoking "perl hipcc" is not ideal, because there is no guarantee that hipcc is actually a perl script. Submitting in lieu of https://github.com/ROCm/Tensile/pull/2037 .

**Outcomes:** 

*What is the result of this change? What components of the project does it affect?*

At this point this is NFC - hipcc.bat script calls perl script.

**Notable changes:** N/A

**Testing and Environment:**

*What environment are you targeting (OS, ROCm version, Python versions, etc.)?* Windows

### DIFF
--- a/Tensile/Common.py
+++ b/Tensile/Common.py
@@ -2483,7 +2483,7 @@ def assignGlobalParameters( config, capabilitiesCache: Optional[dict] = None ):
   # The following try except block computes the hipcc version
   try:
     if os.name == "nt":
-      compileArgs = ['perl'] + [which('hipcc')] + ['--version']
+      compileArgs = [which('hipcc.bat')] + ['--version']
       output = subprocess.run(compileArgs, check=True, stdout=subprocess.PIPE).stdout.decode()
     else:
       compiler = "hipcc"


chinmaydd:
ping. Could someone help us here? Thanks !

ellosel:
Adding @TorreZuk to see if there are any unforeseeable issues with this change.

TorreZuk (approved):
I gave precedence to the .bat in my early commit 10ee6c6107ad21c2013f986736856b5ce8287947.  
Do we know why this was changed later (before this PR) ?

So bypassing perl and isSupportedCompiler seems fine to me as the .bat has been around for a while.  Is there no windows CI yet?  

I built with this change locally and see limited verbosity to confirm much:
This is now generated and untrue.
> UserWarning: hipcc.bat is unsupported for os nt
Local build with this change appears to pass some gemm tests.

TorreZuk:
Okay git blame confusion, I think the original .bat was buggy so the hipcc without extension was the wrapper for perl I used, thus the use of 'perl'  + which( ).   So when you later switched to .bat taking precedence the perl should have been removed.    There is also a .exe now but without windows CI I will not suggest anything further.

ellosel:
> Okay git blame confusion, I think the original .bat was buggy so the hipcc without extension was the wrapper for perl I used, thus the use of 'perl' + which( ). So when you later switched to .bat taking precedence the perl should have been removed. There is also a .exe now but without windows CI I will not suggest anything further.

We've been waiting on windows CI for a while... I don't recall what the reasoning was for the ordering but this changed when migrating to amdclang++.

TorreZuk:
Well fine with this merging to unblock, but think it probably worth a new tech. debt ticket to move off hipcc entirely.  Was  the reason this was left as amdclang++ couldn't provide the hip version info?

chinmaydd:
@TorreZuk thanks for the review. Could someone with access merge this? Thanks again !

chinmaydd:
@ellosel thanks! If you dont mind me asking, what is the `tensile_tag.txt` update cadence in rocBLAS ?

PR #2042, opened 2024-11-05
Fix stream-k dynamic grid model

neoblizz:
There's a typo where `grid_start` is being set to `cuCount`, which should be `grid_end`. I have also removed the defaults to avoid this from happening, as `cuCount` differs from architecture to architecture and even within different compute modes.


AlexBrownAMD (approved):

neoblizz:
> @neoblizz please merge this PR now if you want it in the next release. Thanks.

I don't have merge access to the repo, could you please merge this? Thanks!

PR #2051, opened 2024-11-18
Add master solution library documentation

bstefanuk:
New documentation that clarifies the role of the master solution library (now called solution selection catalog) and details about how the build outputs differ for different build modes.


docs/src/conceptual/solution-selection-catalogs.rst
```diff
@@ -0,0 +1,181 @@
...
+.. _solution-catalogs:
+
+***************************
+Solution selection catalogs
+***************************
+
+After kernels are compiled and linked into code objects libraries (.co files), we still have the problem of how these kernels are executed at runtime. The naive approach
+would be to search through all of the code object libraries until an appropriate kernel is found. A more sophisticated approach is to use a heirarchical structure
+that allows client code to search for kernels using predicates such as hardware, problem size, and transpose, among others. 
+This is the role of the **solution selection catalog** [1]_---a serialized file that uses a heirarchical schema to organize kernel metadata for efficient lookup at runtime.
```
ellosel, line 14:
```suggestion
 Tensile provides a mechanism by which only a subset of the code object files produced during a build are loaded at runtime. This is necessary to avoid the overhead associated with loading code object files including initialization time and the memory footprint of the loaded code object files. However, this introduces the problem of knowing which code object file to load. Solution selection is the process 
by which the **TensileHost** library determines what kernel is preferred and, in turn, 
what code object file contains the selected kernel. This process uses
a hierarchical structure
to efficiently search for kernels based on hardware, problem size, and transpose, among others. 
This is the role of the **solution selection catalog** [1]_---a serialized file that uses a hierarchical schema to organize kernel metadata for efficient lookup at runtime.
```

docs/src/conceptual/solution-selection-catalogs.rst
```diff
@@ -0,0 +1,181 @@
...
+
+.. note::
+    Throughout this document we will refer to catalog files with the .yaml extension. In practice
```
ellosel, line 17: should we add a note that the default format is .dat?

docs/src/conceptual/solution-selection-catalogs.rst
```diff
@@ -0,0 +1,181 @@
...
+
+.. figure:: ../../assets/msl.svg
+    :alt: Master Solution Library hierarchy
+    :align: center
+
+    Solution selection catalog heirarchy for gfx900 and gfx90a
+
+**Level 1: Hardware**
+
+At runtime, only kernels compatible with the host machine's installed hardware can execute. As such, the top level of the hierarchy involves hardware comparisions using GFX architecture.
```
ellosel, line 33:
```suggestion
At runtime, only kernels compatible with the device can execute. As such, the top level of the hierarchy involves hardware comparisons using GFX architecture.
```

docs/src/conceptual/solution-selection-catalogs.rst
```diff
@@ -0,0 +1,181 @@
...
+This layer is a mapping from a GEMM transpose setting, defined using 
+Einstein tensor notation (e.g. *Contraction_l_Alik_Bjlk_Cijk_Dijk*), to a list of problem properties (Level 3).
+
+**Level 3: Problem**
+
+This layer matches against specific problem properties such as input and output types, and features like high precision accumulation and stochastic rounding.
+
+**Level 4: Exact solution**
+
+Finally, exact solutions contain fine-grained details about each solution that can be used during solution selection to locate the best kernel and to assert that the requested problem predicates are satisfied. At this level, there will typically exist a small pool of acceptable kernels for any requested GEMM. Each kernel will have an index and a performance ranking. During solution selection, this highest ranked kernel from this pool will be selected.
```
ellosel, line 46:
```suggestion
Finally, exact solutions contain fine-grained details about each solution that can be used during solution selection to locate the best kernel and to assert that the requested problem predicates are satisfied. Each kernel will have an index and a performance ranking. During solution selection, the highest ranked kernel from this pool will be selected.
```
> During solution selection, the highest ranked kernel from this pool will be selected.

Is this only if there is a math in the pool?

> Each kernel will have an index and a performance ranking

This sounds like it is only true for equality matches.
>

docs/src/conceptual/solution-selection-catalogs.rst
```diff
@@ -0,0 +1,181 @@
...
+
+Build modes
+===========
+
+Tensile comes equipped with multiple build modes, which affect the way solution selection catalogs are generated.
+
+Mode 1: Merge files
+-------------------
+
+When ``--merge-files`` is enabled, one solution catalog is generated for each architecture, named
```
ellosel, line 57: I wonder if we should provide a deprecation warning for merge-files

docs/src/conceptual/solution-selection-catalogs.rst
```diff
@@ -0,0 +1,181 @@
...
+
+.. centered:: TensileLibrary_<gfx>.yaml
+
+The generated catalog contains all information about supported GEMM types, as well as references to 
+solution metadata, which is used to locate the optimal kernel for a requested GEMM. This pattern
```
ellosel, line 62:
```suggestion
A catalog contains information about supported GEMM types and 
solution metadata that is used to locate the optimal kernel for a requested GEMM. This pattern
```

docs/src/conceptual/solution-selection-catalogs.rst
```diff
@@ -0,0 +1,181 @@
...
+is still occasionally used, but has the drawback that all code object libraries need to be loaded eagerly,
+thereby increasing both the initialization time and memory footprint of the calling application.
```
ellosel, line 64:
```suggestion
has the drawback that all code object libraries are loaded eagerly,
thereby increasing both the initialization time and memory footprint of the calling application.
```

docs/src/conceptual/solution-selection-catalogs.rst
```diff
@@ -0,0 +1,181 @@
...
+
+Mode 2: Lazy library loading
+----------------------------
+
+If ``--lazy-library-loading`` is enabled, then a "parent" catalog is generated for each architecture, named
+
+.. centered:: TensileLibrary_lazy_<gfx>.yaml
+
+This file, contains a
+reference to each of it's "child" catalogs, but doesn't have much details about the exact solutions. These settings are instead
```
ellosel, line 95:
```suggestion
reference to each of it's "child" catalogs, but doesn't have details about the exact solutions. These settings are instead 
```

docs/src/conceptual/solution-selection-catalogs.rst
```diff
@@ -0,0 +1,181 @@
...
+Mode 2: Lazy library loading
+----------------------------
+
+If ``--lazy-library-loading`` is enabled, then a "parent" catalog is generated for each architecture, named
+
+.. centered:: TensileLibrary_lazy_<gfx>.yaml
+
+This file, contains a
+reference to each of it's "child" catalogs, but doesn't have much details about the exact solutions. These settings are instead 
+held in the "child" catalogs, whic use the naming convention
```
ellosel, line 96:
```suggestion
held in the "child" catalogs, which use the naming convention 
```

docs/src/conceptual/solution-selection-catalogs.rst
```diff
@@ -0,0 +1,181 @@
...
+
+.. centered:: TensileLibrary_Type_<precision>_<problem type>_<gfx>.yaml
+
+Here, *precision* is the data type, *problem type* is the GEMM type, including transpose and accumulate settings, and *gfx* is the hardware GFX archiecture.
+
+For example, *TensileLibrary_Type_HH_Contraction_l_Alik_Bjlk_Cijk_Dijk_<gfx>.yaml* identifies a code object library for half precision
+contractions on two transpose matrices, otherwise known as HGEMM TT.
+In this way, the child catalogs are responsible for holding the actual solution metadata, while the parent catalog is responsible for organizing the child catalogs
+by hardware, problem type, transpose, precision, and other predicates.
+This has the benefit of reducing the memory footprint of the calling application, as code object libraries are compiled separately and loaded only when required.
```
ellosel, line 106:
```suggestion
In this way, the child catalogs are responsible contain the solution metadata, while the parent catalog is responsible for organizing the child catalogs
by hardware, problem type, transpose, precision, and other predicates.
This has the benefit of reducing the memory footprint of the calling application, as code object libraries are compiled separately and loaded only when required.
```

ellosel (approved):

PR #2068, opened 2024-12-17
Do not use aliases as nested namespace specifiers

ellosel:
**Summary:**

With clang 19, existing code is triggering warnings. The warning points out that one cannot use an alias template as a namespace specifier. See this reproducer: https://godbolt.org/z/xn6YhPKxv 

**Outcomes:**

Warnings are no longer emitted.

**Notable changes:**

Used the name of the type directly rather than the alias.

**Testing and Environment:**

This only occurs in environments with clang 19. Tested by running the following command in Tensile:

```
tox run -e hostlibtest
```

and the following command from rocBLAS:

```
./install.sh -a gfx900 -t /mnt/host/projects/Tensile
```


TorreZuk (approved):
LGTM, maybe same for hipblaslt

PR #2070, opened 2024-12-17
Mark VOP3v_dot4_i32_i8 as True for 11.0.0

ellosel:
Recent versions of amdclang do not agree with capabilities cache. This change corrects one such disagreement.


ellosel:
Closing in favor of #2076

PR #2073, opened 2024-12-20
Improve default toolchain configuration

bstefanuk:
The link command is invoked in a non-standard way, using a compile_args.txt file. Updating the directory where this file is created fixes builds on Windows. This PR also includes improvements to out-of-the-box toolchain configuration such that environment variables are no longer required for specifying assembler and offload bundler paths, particularly on Windows.

**Testing**
Clean rocBLAS installation with target architecture gfx1201 and gfx1151 (separate installation runs)

Windows
```bash
python rmake.py -a "gfx1201" --lazy-library-loading --no-merge-architectures -t "C:\Users\bstefanu\dev\Tensile" --no_hipblaslt
```
```bash
python rmake.py -a "gfx1151" -t "C:\Users\bstefanu\dev\Tensile" -i
```
Linux
```bash
./install.sh -dc -a "gfx1151;gfx1201" -i -t "/mnt/host/Tensile" --no_hipblaslt
```

**Environment:**
Windows 10, ROCm 6.2.41512, Python 3.13.1, GPU gfx1100, CMake 3.31.3

### DIFF
--- a/Tensile/BuildCommands/AssemblyCommands.py
+++ b/Tensile/BuildCommands/AssemblyCommands.py
@@ -28,12 +28,13 @@ def _linkIntoCodeObject(
     """
     args = []
     if os.name == "nt":
-        # On Windows, the objectFiles list command line (including spaces)
-        # exceeds the limit of 8191 characters, so using response file
-        with tempfile.NamedTemporaryFile(mode="wt", delete=False) as f:
-            f.write(" ".join(o.replace("\\", "/") for o in objFiles))
-            f.flush()
-            args = kernelWriterAssembly.getLinkCodeObjectArgs([f"@{f.name}"], str(coPathDest))
+        # On Windows, it is possible for the list of `objFiles` to exceed the command line limit
+        # LLVM allows the provision of compiler arguments via a "response file" (`rf` below)
+        # Reference: https://llvm.org/docs/CommandLine.html#response-files
+        with tempfile.NamedTemporaryFile(mode="wt", delete=False) as rf:
+            rf.write(" ".join(o.replace("\\", "/") for o in objFiles))
+            rf.flush()
+            args = kernelWriterAssembly.getLinkCodeObjectArgs([f"@{rf.name}"], str(coPathDest))
     else:
         args = kernelWriterAssembly.getLinkCodeObjectArgs(objFiles, str(coPathDest))
 
--- a/Tensile/BuildCommands/AssemblyCommands.py
+++ b/Tensile/BuildCommands/AssemblyCommands.py
@@ -29,13 +29,12 @@ def _linkIntoCodeObject(
     if os.name == "nt":
         # On Windows, the objectFiles list command line (including spaces)
         # exceeds the limit of 8191 characters, so using response file
-
-        responseFile = os.path.join("/tmp", "clangArgs.txt")
+        responseFile = os.path.join(Path.cwd(), "clang_args.txt")
         with open(responseFile, "wt") as file:
             file.write(" ".join(objFiles))
             file.flush()
 
-        args = kernelWriterAssembly.getLinkCodeObjectArgs(["@clangArgs.txt"], str(coPathDest))
+        args = kernelWriterAssembly.getLinkCodeObjectArgs(["@clang_args.txt"], str(coPathDest))
     else:
         args = kernelWriterAssembly.getLinkCodeObjectArgs(objFiles, str(coPathDest))
 
--- a/Tensile/BuildCommands/AssemblyCommands.py
+++ b/Tensile/BuildCommands/AssemblyCommands.py
@@ -32,7 +32,8 @@ def _linkIntoCodeObject(
         # LLVM allows the provision of compiler arguments via a "response file" (`rf` below)
         # Reference: https://llvm.org/docs/CommandLine.html#response-files
         with tempfile.NamedTemporaryFile(mode="wt", delete=False) as rf:
-            rf.write(" ".join(o.replace("\\", "/") for o in objFiles))
+            # Use repr to a get raw string with non-escaped path separators
+            rf.write(" ".join(map(repr, objFiles)))
             rf.flush()
             args = kernelWriterAssembly.getLinkCodeObjectArgs([f"@{rf.name}"], str(coPathDest))
     else:
--- a/Tensile/BuildCommands/AssemblyCommands.py
+++ b/Tensile/BuildCommands/AssemblyCommands.py
@@ -2,6 +2,7 @@ import collections
 import os
 import shutil
 import subprocess
+import tempfile
 from pathlib import Path
 from typing import List, Union
 
@@ -29,12 +30,10 @@ def _linkIntoCodeObject(
     if os.name == "nt":
         # On Windows, the objectFiles list command line (including spaces)
         # exceeds the limit of 8191 characters, so using response file
-        responseFile = os.path.join(Path.cwd(), "clang_args.txt")
-        with open(responseFile, "wt") as file:
-            file.write(" ".join(objFiles))
-            file.flush()
-
-        args = kernelWriterAssembly.getLinkCodeObjectArgs(["@clang_args.txt"], str(coPathDest))
+        with tempfile.NamedTemporaryFile(mode="wt", delete=False) as f:
+            f.write(" ".join(o.replace("\\", "/") for o in objFiles))
+            f.flush()
+            args = kernelWriterAssembly.getLinkCodeObjectArgs([f"@{f.name}"], str(coPathDest))
     else:
         args = kernelWriterAssembly.getLinkCodeObjectArgs(objFiles, str(coPathDest))
 
--- a/Tensile/Common.py
+++ b/Tensile/Common.py
@@ -25,6 +25,7 @@
 from . import __version__
 from . import Parallel
 from .Utilities.ConditionalImports import print, TENSILE_TERM_COLORS
+from .Utilities.Toolchain import getVersion
 from collections import OrderedDict
 
 from copy import deepcopy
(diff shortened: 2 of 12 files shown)


Tensile/Common.py
```diff
@@ -348,23 +349,36 @@ def getArchitectureName(gfxName: str) -> Optional[str]:
...
+#   Return:
+#       If supported True; otherwise, False.
+#   """
+#   isSupported = (compiler == "hipcc")
+#   if os.name == "nt":
+#     isSupported = (isSupported or compiler == "clang++" or compiler == "hipcc.bat")
+#   else:
+#     isSupported = isSupported or compiler == "amdclang++"
+#   if not isSupported: printWarning(f"{compiler} is unsupported for os {os.name}")
+#   return isSupported
```
bstefanuk, line 367: Remove commented code

Tensile/BuildCommands/AssemblyCommands.py
```diff
@@ -29,13 +29,12 @@ def _linkIntoCodeObject(
     if os.name == "nt":
         # On Windows, the objectFiles list command line (including spaces)
         # exceeds the limit of 8191 characters, so using response file
-
-        responseFile = os.path.join("/tmp", "clangArgs.txt")
+        responseFile = os.path.join(Path.cwd(), "clang_args.txt")
```
TorreZuk, line 32: The current working directory may not be appropriate for temp files.

Tensile/Common.py
```diff
@@ -2232,36 +2246,47 @@ def gfxName(arch):
...
-        if arch is not None:
-          if arch in globalParameters["SupportedISA"]:
-            tPrint(1, "# Detected local GPU with ISA: " + gfxName(arch))
-            globalParameters["CurrentISA"] = arch
-    if (process.returncode):
-      printWarning("%s exited with code %u" % (globalParameters["ROCmAgentEnumeratorPath"], process.returncode))
-    return process.returncode
-  return 0
+  if globalParameters["CurrentISA"] != (0,0,0) or not globalParameters["ROCmAgentEnumeratorPath"]:
+    return 0
```
TorreZuk, line 2276: How does this work on windows without rocm agent enumerator?   Do you have windows CI running now?

TorreZuk (commented):
The cmake code I was also interested in as we call it from rocblas

Tensile/BuildCommands/AssemblyCommands.py
bstefanuk: This was the line of code that originally prompted a number of defect reports for Windows build. `/tmp` won't work either, I haven't found an issue with the CWD. Do you have another recommendation? At least the CWD is guaranteed to exist.

Tensile/Common.py
bstefanuk: We do not have CI running on Windows. However, since this ticket is aimed at bolstering Windows out-of-the-box configuration, I'm testing all changes on Windows regularly.

Working within the constraints of the current impl, ROCmAgentEnumerator is either set to False (if no enumerate is used) or to the fully qualified path of the device enumerator, `rocm_agent_enumerator` for Linux and `hipInfo.exe` for Windows.

Tensile/BuildCommands/AssemblyCommands.py
TorreZuk: Well these are temp files so if you are willing to " import tempfile " then get the temp directory.  Or any other helpers to get a full temp path file name you could use,  doesn't have to be clang_args.txt of course.   Or just say create /tmp in cwd if you know it is in build tree.

ellosel: I like what @TorreZuk suggested - quick google search:

https://stackoverflow.com/questions/847850/cross-platform-way-of-getting-temp-directory-in-python

Tensile/BuildCommands/AssemblyCommands.py
```diff
@@ -29,13 +29,12 @@ def _linkIntoCodeObject(
...
         with open(responseFile, "wt") as file:
             file.write(" ".join(objFiles))
             file.flush()
 
-        args = kernelWriterAssembly.getLinkCodeObjectArgs(["@clangArgs.txt"], str(coPathDest))
+        args = kernelWriterAssembly.getLinkCodeObjectArgs(["@clang_args.txt"], str(coPathDest))
```
ellosel, line 37: This seems obscure enough to warrant adding a comment with a link to clang docs.

Tensile/Common.py
```diff
@@ -2232,36 +2230,47 @@ def gfxName(arch):
...
-        arch = gfxArch(line.strip())
-        if arch is not None:
-          if arch in globalParameters["SupportedISA"]:
-            tPrint(1, "# Detected local GPU with ISA: " + gfxName(arch))
-            globalParameters["CurrentISA"] = arch
-    if (process.returncode):
-      printWarning("%s exited with code %u" % (globalParameters["ROCmAgentEnumeratorPath"], process.returncode))
-    return process.returncode
-  return 0
+  if globalParameters["CurrentISA"] != (0,0,0) or not globalParameters["ROCmAgentEnumeratorPath"]:
```
ellosel, line 2259: Seems odd to return 0 if rocm_agent_enumerator is none. What are the semantics here?

Tensile/TensileCreateLibrary.py
```diff
@@ -1177,7 +1159,9 @@ def verifyManifest(manifest: Path) -> bool:
     """
     with open(manifest, mode="r") as generatedFiles:
         for f in generatedFiles.readlines():
+            tPrint(1, f"Checking file: {f}")
```
ellosel, line 1162: Seems like this could get pretty verbose. Should we just add the warning below?

Tensile/TensileCreateLibrary.py
```diff
@@ -1407,7 +1402,10 @@ def TensileCreateLibrary():
             printExit("Failed to verify all files in manifest")
 
     tPrint(1, "# CodeObjectVersion: %s" % args["CodeObjectVersion"])
-    tPrint(1, "# CxxCompiler:       %s" % cxxCompiler)
+    tPrint(1, "# CxxCompiler:       %s" % args["CxxCompiler"])
+    tPrint(1, "# CCompiler:         %s" % args["CCompiler"])
+    tPrint(1, "# Assembler:         %s" % args["Assembler"])
+    tPrint(1, "# OffloadBundler:    %s" % args["OffloadBundler"])
```
ellosel, line 1408: Once we implement a toolchain class, we should consider implementing __str__ and __repr__:

```python
class Test:
    def __init__(self, a, b):
        self.a = a
        self.b = b

    def __repr__(self):
        return f"<Test a:{self.a} b:{self.b}>"

    def __str__(self):
        return f"From str method of Test: a is {self.a}, b is {self.b}"
```

Tensile/TensileCreateLibrary.py
```diff
@@ -1437,6 +1435,8 @@ def TensileCreateLibrary():
     masterLibraries = generateLogicData(
         logicFiles, args["Version"], args["PrintLevel"], args["SeparateArchitectures"]
     )
+    tPrint(1, f"# Found {len(masterLibraries)} Master Files")
+    tPrint(1, f"# {', '.join(masterLibraries.keys())}")
```
ellosel, line 1439: Was this for debugging purposes?

Tensile/Common.py
bstefanuk: Semantics are simply unchanged from the previous implementation

Tensile/TensileCreateLibrary.py
bstefanuk: Yes, I agree, will remove

Tensile/Common.py
bstefanuk: To clarify, this behaviour is unchanged, all that is different now is that we return 0 immediately with a conditional, instead of nesting all of the important logic inside an if block.

Tensile/TensileCreateLibrary.py
bstefanuk: Yes, removed.

Tensile/Common.py
```diff
@@ -348,23 +349,20 @@ def getArchitectureName(gfxName: str) -> Optional[str]:
...
+  isSupported = hipExe == "hipcc" or hipExe == "hipconfig"
   if os.name == "nt": 
-    isSupported = (isSupported or compiler == "clang++")
-  else:
-    isSupported = (isSupported or compiler == "amdclang++")
-  
-  if not isSupported: printWarning(f"{compiler} is unsupported for os {os.name}")
-  
+    isSupported = (isSupported or hipExe == "hipcc.bat")
+  if not isSupported: printWarning(f"{hipExe} is unsupported for os {os.name}")
```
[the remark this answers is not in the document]
bstefanuk: Deleted.

Tensile/Common.py
ellosel: To me that implies we succeed if we can't find the tool, is that true?

bstefanuk: It's a weird way of changing the CurrentISA on Windows
```
  # read current gfx version
  returncode = detectGlobalCurrentISA()
  if globalParameters["CurrentISA"] == (0,0,0):
    printWarning("Did not detect SupportedISA: %s; cannot benchmark assembly kernels." % globalParameters["SupportedISA"])
  if returncode:
    if os.name == "nt":
      globalParameters["CurrentISA"] = (9,0,6)
      printWarning("Failed to detect ISA so forcing (gfx906) on windows")
```
See L2469 of Common.py

bstefanuk:
@TorreZuk There are no cmake code changes to make for this PR regarding how rocBLAS calls TensileCreateLibrary, I will post a PR for rocBLAS in the coming days with the intended functionality.

Tensile/BuildCommands/AssemblyCommands.py
```diff
@@ -27,15 +28,13 @@ def _linkIntoCodeObject(
...
-        with open(responseFile, "wt") as file:
-            file.write(" ".join(objFiles))
-            file.flush()
-
-        args = kernelWriterAssembly.getLinkCodeObjectArgs(["@clangArgs.txt"], str(coPathDest))
+        # On Windows, it is possible for the list of `objFiles` to exceed the command line limit
+        # LLVM allows the provision of compiler arguments via a "response file" (`rf` below)
+        # Reference: https://llvm.org/docs/CommandLine.html#response-files
+        with tempfile.NamedTemporaryFile(mode="wt", delete=False) as rf:
+            rf.write(" ".join(o.replace("\\", "/") for o in objFiles))
```
TorreZuk, line 35: Does the clang on windows expect linux path seperator?  Must have worked before...

Tensile/Common.py
TorreZuk: I still don't follow if no ROCmAgentEnumeratorPath  I expected the returncode non zero to force the ISA 906.

TorreZuk (approved):
Fine by me if it works.   I think it would be improvement if you have a cmake addition to set the compiler global variables if set already set in cmake as per rocBLAS usage.   Thus if we did setup a custom clang++ you would inherit it, I only problem is offload bunder etc. but could look in same path.

Tensile/Common.py
bstefanuk: I have updated the return code to be non-zero so it will force the Windows device to be gfx906

TorreZuk: The negated conjuction is not the easiest to read, but if CurrentISA was set non zero you really shouldn't force it IMO.    The first part of the conjunction really should be separate test.  Sure maybe it never gets preset and pre-existing design was just as weak.

Tensile/BuildCommands/AssemblyCommands.py
bstefanuk: Without this the following error is thrown: "error: no such file or directory: 'C:UsersbstefanudevTensilebuild_tmp_BUILD1151assemblyCijk_Ailk_Bjlk_BBS_BH_GB_MT128x128x32_MI16x16x1621PTd_JjhS-eLfm_2MUQ4RtvL0gaX4jkR58d9Aok1u0=.o"

I've updated to use raw-strings instead (using `repr`) such that the escape characters are ignored in the paths in the temp file.

ellosel (approved):

Tensile/BuildCommands/AssemblyCommands.py
bstefanuk: Ahh, the reason this is needed now is because the algorithm changed to use absolute paths instead of a CWD on the call to subprocess. Previously, the code didn't have path separators for entries in the response file.

TorreZuk: Okay the argument list will be a lot larger than before too

PR #2076, opened 2025-01-06
Compiler version dependent capabilities

ellosel:
**Summary:**

On occasion the capabilities cache is out of sync with the test environments used across the different teams. As a result, we can't update tensile to resolve a defect without breaking downstream math ci projects. This change attempts to cope with these environment issues without losing the ability to error out when there is a capability mismatch.

**Outcomes:**

The AsmCache module is updated to include a function that computes the cache according to the compiler version. Common.py was updated to make use of the new function.

**Notable changes:**

The added function is called `getCapabilitiesCache`

**Testing and Environment:**

Tested using routine CI and local testing.

**What testing did you do to ensure this change will integrate successfully?**

Using images with clang 18 and clang 19.


Tensile/Common.py
```diff
@@ -2353,9 +2353,6 @@ def populateCapabilities(
     Note:
```
bstefanuk, line 2353: Pleas update doc-string

Tensile/Common.py
```diff
@@ -2480,12 +2477,18 @@ def assignGlobalParameters( config, capabilitiesCache: Optional[dict] = None ):
   globalParameters["CacheAsmCaps"] = True if capabilitiesCache is not None else False
   globalParameters["AsmCaps"] = capabilitiesCache if globalParameters["CacheAsmCaps"] else {}
   globalParameters["ArchCaps"] = {}
-  populateCapabilities(globalParameters, CACHED_ASM_CAPS)
+
+  compilerVer = CompilerVersion(
+    *[int(c) for c in globalParameters["HipClangVersion"].split(".")[:2]]
+  )
```
bstefanuk, line 2483: Perhaps not in scope, but we compute the compiler version in several different ways, sometimes with a global parameter, sometimes not. It would be good if we could consolidate the methods to determine and access the compiler version.

Tensile/AsmCaps.py
```diff
@@ -22,884 +22,909 @@
...
-              'v_fma_mix_f32': True,
-              'v_fmac_f16': False,
-              'v_fmac_f32': True,
-              'v_mac_f16': False,
-              'v_mac_f32': False,
-              'v_mad_mix_f32': False,
-              'v_mov_b64': False,
-              'v_pk_fma_f16': True,
-              'v_pk_fmac_f16': False}}
+def getCapabilitiesCache(compiler_version: tuple) -> dict:
```
bstefanuk, line 25:
```suggestion
def getCapabilitiesCache(compilerVersion: tuple) -> dict:
```
Staying consistent with camelCase is probably best.

Tensile/Common.py
ellosel: I think this part of the code needs to be separated out and probably collected into the toolchain module. Perhaps another time.

Tensile/AsmCaps.py
```diff
@@ -22,884 +22,909 @@
...
+    Args:
+        compiler_maj_version: The major version of the compiler.
+
+    Returns:
+        A dictionary representing the capabilities cache.
+    """
+
+    print(f"Creating capabilities cache for ROCm {compiler_version}")
+
+    CACHED_ASM_CAPS = \
```
bstefanuk, line 37: I wonder if we should re-name this for the capabilities cache for the origin compiler that it represents, e.g., if it's amdclang++ 18.0.0 then we could call it CACHED_ASM_CAPS_AMDCLANG_18 or something similar.

Tensile/AsmCaps.py
```diff
@@ -22,884 +22,909 @@
...
+                  'v_mad_mix_f32': False,
+                  'v_mov_b64': False,
+                  'v_pk_fma_f16': True,
+                  'v_pk_fmac_f16': False}}
+
+    def nested_update_(left, right):
+        for key, val in right.items():
+            left[key].update(val)
+
+    if compiler_version.major * 10 + compiler_version.minor > 63:
```
bstefanuk, line 923: I see what you're doing here, but a small inline documentation example could be nice for future reference since this isn't as intuitive as the expected syntax of `if ver.major >= 6 and ver.minor > 3`

Tensile/Common.py
ellosel: Done

Tensile/AsmCaps.py
ellosel: I was trying to avoid a cycle by pulling in Common. Will update to camel.

bstefanuk (commented):
Looks great overall. Something I'm noticing is the naming semantics for compilerVersion, which represents the version for hipcc/hipconfig. But, this is misleading because these aren't the default supported compilers anymore. Perhaps it should be renamed to "hipVersion" and we grab the actual amdclang++ version for the "compilerVersion" variable.

Tensile/AsmCaps.py
ellosel:
```
if ver.major >= 6 and ver.minor > 3
```
 would not give correct behavior for 7.0.0 and I wanted to avoid the complication of having nested if statements.

AlexBrownAMD (approved):
We should have a similar update for tensilelite

PR #2079, opened 2025-01-08
Fixes various issues in the tuneing scripts

IMbackK:
672afc68985b149f1f19c90758c796cee1424537 fixes https://github.com/ROCm/Tensile/issues/2077 issue 1

8829b024de7db18a0773ac6e2542fb9ba308435b fixes https://github.com/ROCm/Tensile/issues/2077 Issue 4

39451b8faa4f1d055118e58817e7d76b80f646dc updates the tuning script to align with rocblas change 7972a13cb93611d12c47d72e3bf15acd8ca4b1ee

ff26d8ab116e0148c1db92ffac5dac2e104b047b partially fixes https://github.com/ROCm/Tensile/issues/2077 Issue 2, but many bugs with the rocblas log to tensile benchmark config conversion remains, not limited to:
1. Only the first data type configuration set in the rocblas log file is taken as the configuration for all the benchmark problems, however the log may contain different configurations 
    1. The different configurations should be forked instead of being only in the common benchmark config paramters, but they are not
2. Configurations where the input and output dtypes are not the same are borken

59226983dff72d875edbb3000a564d96f18567cf updates the parameters of TensileCreateLibrary to work with the current version and fixes the TODO in provision_verification that previously hardcoded python 3.6, which is ofc not useful in 2025

Overall the tuning scripts are still in an __extremely sorry__ state, there are many pitfalls in these scripts such as:
1. various Chips are hard coded in various places, but it only goes up to Arcturus, leaving Aldebaran and up in the cold, never mind gfx10+
2. Very little error handling while with many faulty assumptions about system environment
3. various flags like --redo do not work due to faulty assumptions about the behavior of a posix shell
4. many many more

### DIFF
--- a/tuning/automation/ExtractSizes.py
+++ b/tuning/automation/ExtractSizes.py
@@ -731,6 +731,8 @@ def ConvertToRocBlasBenchCall(line):
     sameParams = set(['b_type','c_type','d_type','compute_type','lda','ldb','ldc','ldd','batch','batch_count','algo','solution_index','flags','stride_a','stride_b','stride_c','stride_d','alpha','beta'])
 
     for item in range(2,len(line)):
+        if line[item] == 'flags' and line[item+1] == 'none':
+            line[item+1] = '0'
         if line[item] in sameParams:
             benchLine += ('--'+line[item]+' '+line[item+1]+' ')
         if line[item] == 'transA':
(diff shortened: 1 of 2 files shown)


IMbackK:
Please add a run of master_tuning_script.sh to your ci to prevent the tuning scripts from falling into this degree of disrepair again.

Please also consider releasing a full set of benchmark problem files for all the logic files in rocBlas with a script to retune every solution contained in rocblas. Ie end to end documentation and scripts on how to regenerate all logic files in rocblas from https://github.com/ROCm/Tensile/tree/develop/Tensile/Configs

babakpst:
@IMbackK Thanks for your comments and the PR. We no longer use these scripts in our tuning workflow, and, as a result, we do not update them. I will clean the repo soon. I appreciate your time and consideration.

IMbackK:
Well, will there be a replacement for application specific tuning?

PR #2080, opened 2025-01-08
Find non-default HIP SDK on Windows

bstefanuk:
**Summary:**

When the HIP SDK isn't installed to the default location, then the current build strategy cannot find toolchain components without fully qualified paths, even though the environment variable HIP_PATH contains the relevant information. This PR updates the strategy that Tensile uses to find and verify toolchain components. Minor updates to how this process is conducted on Linux is included as needed. Also includes a document on troubleshooting the relevant error and adds formatting to Toolchain.py.


Tensile/TensileCreateLib/ParseArguments.py
```diff
@@ -79,7 +79,8 @@ def parseArguments(input: Optional[List[str]] = None) -> Dict[str, Any]:
         default=ToolchainDefaults.CXX_COMPILER,
         type=str,
         help="C++ compiler used when generating binaries."
-        " On linux, amdclang++ (default) or hipcc. On Windows clang++ (default) or hipcc.",
+        " On linux, amdclang++ (default) or hipcc. On Windows clang++.exe (default) or hipcc. "
+        "If clang++.exe is specified, it must include the extension `.exe`.",
```
ellosel, line 83: Excellent

docs/src/support/troubleshooting.rst
```diff
@@ -0,0 +1,38 @@
...
+
+In this situation, Tensile cannot locate one or more binaries required fir proper program execution. This includes compilers, assemblers, linkers, and bundlers. 
+
+.. note::
+
+   On Linux, the default installation location is */opt/rocm*. 
+   
+.. note::
+
+   On Windows, the default installation location is *C:\\Program Files\\AMD\\ROCm\\X.Y*, where *X.Y* identify the major and minor version of the current ROCm installation.
```
ellosel, line 30:
```suggestion
   On Windows, the default installation location is *C:\\Program Files\\AMD\\ROCm\\X.Y*, where *X.Y* identifies the major and minor version of the current ROCm installation.
```

docs/src/support/troubleshooting.rst
```diff
@@ -0,0 +1,38 @@
...
+   When the HIP SDK is installed on Windows, the variable ``HIP_PATH`` is set to the installation location of the HIP SDK.
```
ellosel, line 31: Did/should we correct HIP_DIR in common?

bstefanuk: I did not, yes we should. I will add all of these changes to the commit.

TorreZuk:
Is the end use case here to use with windows CI or within rocBLAS based build?

bstefanuk:
Yes, partly, the builds for rocBLAS are intended to remain the same, that is, using default install locations. This update is targeted to improve toolchain management for the increasing number of one-off builds requested by Windows users.

ellosel (approved):

PR #2089, opened 2025-01-29
Change clang-offload-bundler to 4-tuple

### LINKED ISSUE #2086: Use of `clang-offload-bundler` needs to be updated
We are moving to enforce 4-tuple triple in the bundle id. Specifically, the bundle id in `-targets=` has to be in the form of:

```
<kind>-<4-tuple-triple>[-<target id>[:<target features>]]
```

where `<4-tuple-triple>` is `arch-vendor-os-env`. For host, it will be something like `host-x86_64-unknown-linux-gnu `; for AMDGPUs, it will be `amdgcn-amd-amdhsa-unknown` or even `amdgcn-amd-amdhsa-`.

Please update the use accordingly to avoid any potential break.

darren-amd:
resolves #2086 

**Summary:**
Change usage of clang-offload-bundler from 3-tuple to 4-tuple with the new <env> parameter. This was the only usage I could find in the repo.

### DIFF
--- a/Tensile/BuildCommands/SharedCommands.py
+++ b/Tensile/BuildCommands/SharedCommands.py
@@ -25,7 +25,7 @@ def compressCodeObject(
         "--compress",
         "--type=o",
         "--bundle-align=4096",
-        f"--targets=host-x86_64-unknown-linux,hipv4-amdgcn-amd-amdhsa--{gfx}",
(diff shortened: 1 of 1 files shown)


bstefanuk:
Thank you for the PR @darren-amd 

We're working on resolving some CI issues, after which I'll retrigger this and see if it passes all tests

bstefanuk (approved):

PR #2090, opened 2025-01-29
Only warn when caps mismatch

ellosel:
There are different environments used in testing and each of the environments are reporting different capabilities. This has introduced a dependency cycle that we cannot get out of without temporarily making this change.


TorreZuk (approved):
LGTM, if you don't squash and merge I'll start a run using this commit for rocblas develop now.

bstefanuk (approved):

PR #2091, opened 2025-01-30
Only warn when caps mismatch (#2090)

ellosel:
Capabilities cache temporary workaround.


bstefanuk:
Why do you have a different issue number in the title than is for the ticket, "#2090"?

PR #2092, opened 2025-01-30
Update supported toolchain components

bstefanuk:
**Summary:**

Expand the supported toolchain components, listing them out explicitly, so that there are no hidden assumptions made within the build system.


Tensile/Utilities/Toolchain.py
```diff
@@ -110,7 +114,11 @@ def supportedCxxCompiler(compiler: str) -> bool:
     Return:
         If supported True; otherwise, False.
     """
-    return _supportedComponent(compiler, [ToolchainDefaults.CXX_COMPILER, "hipcc"])
+    if os.name == "nt":
+        return _supportedComponent(
+            compiler, ["clang++.exe", "clang++", "hipcc", "hipcc.bat", "amdclang++"]
```
TorreZuk, line 119: If you include clang++.exe then amdclang++.exe may as well be included.  Same goes for C compiler, if you allow extension do it for both.

TorreZuk (approved):
Recommend minor changes.  LGTM if a windows CI build is passing with this.   There is also a hipcc.exe on windows

Tensile/Utilities/Toolchain.py
bstefanuk: This is where I pulled these details from
image

bstefanuk: Which OS is providing amdclang++ on Windows?

TorreZuk: I have a hipcc.exe and I didn't make it.  If you trust your source you could add the strangely named hipcc.bin.exe.  The amdclang is to match linux, so no need to break things on an expected change.

bstefanuk: For future reference, amdclang will be added in future ROCm version to better differentiate between AMD's clang from upstream clang.

PR #2095, opened 2025-02-03
Use PATHEXT extensions on Windows

bstefanuk:
**Summary:**

Search PATHEXT extensions on Windows. Trialled on two different Windows machines showed PATHEXT defaults to the following: `.COM;.EXE;.BAT;.CMD;.VBS;.VBE;.JS;.JSE;.WSF;.WSH;.MSC;.CPL`.


ellosel (approved):
This looks reasonable to me. We take the requested file and add to it a known list of extensions and/or those extracted from PATHEXT.

PR #2110, opened 2025-03-06
Update find_package for msgpack to work with 5.x and 6.x and Capture configure-time PATH when invoking tensile at build time

### LINKED ISSUE #278: [Tensile] Update find_package for msgpack to work with 5.x and 6.x
## Summary

To work with msgpack 5.x and 6.x a patch is needed. The parch is adapted from: https://github.com/msgpack/msgpack-c/wiki/Q%26A#how-to-support-both-msgpack-c-c-version-5x-and-6x-

Note that 6.x also defines compile definitions that must be respected.

### Link to patch

https://github.com/ROCm/TheRock/blob/main/patches/amd-mainline/Tensile/0001-Update-find_package-for-msgpack-to-work-with-5.x-and.patch

### Permalink to patch

https://github.com/ROCm/TheRock/blob/6e1bd9b82a581487712f95ef86c0ba27c77faec9/patches/amd-mainline/Tensile/0001-Update-find_package-for-msgpack-to-work-with-5.x-and.patch

### LINKED ISSUE #279: [Tensile] Capture configure-time PATH when invoking tensile at  build time
## Summary

Tensile relies on the tools from the path, so capture the configure time path. This is a minimal patch as making it explicit would be a pretty big change.

### Link to patch

https://github.com/ROCm/TheRock/blob/main/patches/amd-mainline/Tensile/0002-Capture-configure-time-PATH-when-invoking-tensile-at.patch

### Permalink to patch

https://github.com/ROCm/TheRock/blob/6e1bd9b82a581487712f95ef86c0ba27c77faec9/patches/amd-mainline/Tensile/0002-Capture-configure-time-PATH-when-invoking-tensile-at.patch

zichguan-amd:
See TheRock patches https://github.com/ROCm/TheRock/issues/278 and https://github.com/ROCm/TheRock/issues/279 for more details.

Closes https://github.com/ROCm/TheRock/issues/278, Closes https://github.com/ROCm/TheRock/issues/279

### DIFF
--- a/Tensile/Source/lib/CMakeLists.txt
+++ b/Tensile/Source/lib/CMakeLists.txt
@@ -90,24 +90,20 @@ if(TENSILE_USE_LLVM OR TENSILE_USE_MSGPACK)
 endif()
 
 if(TENSILE_USE_MSGPACK)
-    find_package(msgpack REQUIRED NAMES msgpack msgpack-cxx msgpack-c)
+    # See: https://github.com/msgpack/msgpack-c/wiki/Q%26A#how-to-support-both-msgpack-c-c-version-5x-and-6x-
+    # Prefer 6.x (msgpack-cxx) as that is what we bundle in the build.
+    find_package(msgpack-cxx CONFIG)
+    if(msgpack-cxx_FOUND)
+        # Version 6.x
+        message(STATUS "Found msgpack-cxx (>=6.x)")
+        target_link_libraries(TensileHost PUBLIC msgpack-cxx)
+    else()
+        # Fallback to <= 5.x
+        find_package(msgpackc-cxx CONFIG REQUIRED NAMES msgpackc-cxx msgpack)
+        message(STATUS "Found msgpack (<=5.x)")
+        target_link_libraries(TensileHost PUBLIC msgpackc)
+    endif()
     target_compile_definitions(TensileHost PUBLIC -DTENSILE_MSGPACK=1)
-
-    if(TARGET msgpackc-cxx)
-        get_target_property(msgpack_inc msgpackc-cxx INTERFACE_INCLUDE_DIRECTORIES)
-    elseif(TARGET msgpackc)
-        get_target_property(msgpack_inc msgpackc INTERFACE_INCLUDE_DIRECTORIES)
-    elseif(TARGET msgpack-cxx)
-        get_target_property(msgpack_inc msgpack-cxx INTERFACE_INCLUDE_DIRECTORIES)
-    endif()
-
-    if(DEFINED msgpack_inc)
-        # include C++ headers manually
-        # External header includes included as system files
-        target_include_directories(TensileHost 
-            SYSTEM PRIVATE $<BUILD_INTERFACE:${msgpack_inc}>
-        )
-    endif()
 endif()
 
 if(TENSILE_USE_LLVM)
--- a/Tensile/cmake/TensileConfig.cmake
+++ b/Tensile/cmake/TensileConfig.cmake
@@ -219,6 +219,12 @@ function(TensileCreateLibraryFiles
   if (WIN32 OR (VIRTUALENV_BIN_DIR AND VIRTUALENV_PYTHON_EXENAME))
     set(CommandLine ${VIRTUALENV_BIN_DIR}/${VIRTUALENV_PYTHON_EXENAME} ${CommandLine})
   endif()
+  # Tensile relies on the tools from the path, so capture the configure time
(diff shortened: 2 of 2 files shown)


Tensile/Source/lib/CMakeLists.txt
```diff
@@ -90,24 +90,20 @@ if(TENSILE_USE_LLVM OR TENSILE_USE_MSGPACK)
...
+    find_package(msgpack-cxx CONFIG)
+    if(msgpack-cxx_FOUND)
+        # Version 6.x
+        message(STATUS "Found msgpack-cxx (>=6.x)")
+        target_link_libraries(TensileHost PUBLIC msgpack-cxx)
+    else()
+        # Fallback to <= 5.x
+        find_package(msgpackc-cxx CONFIG REQUIRED NAMES msgpackc-cxx msgpack)
+        message(STATUS "Found msgpack (<=5.x)")
+        target_link_libraries(TensileHost PUBLIC msgpackc)
```
davidd-amd, line 104: If we bundle v6 in the build do we need/want a fallback. I need to look into what versions are available across all of the supported platforms.

Tensile/cmake/TensileConfig.cmake
```diff
@@ -219,6 +219,12 @@ function(TensileCreateLibraryFiles
   if (WIN32 OR (VIRTUALENV_BIN_DIR AND VIRTUALENV_PYTHON_EXENAME))
     set(CommandLine ${VIRTUALENV_BIN_DIR}/${VIRTUALENV_PYTHON_EXENAME} ${CommandLine})
   endif()
+  # Tensile relies on the tools from the path, so capture the configure time
+  # path. It would be better if this were explicit, but that would be a pretty
+  # big change.
+  set(CommandLine
+    "${CMAKE_COMMAND}" -E env "'PATH=$ENV{PATH}'" --
+    ${CommandLine})
   message(STATUS "Tensile_CREATE_COMMAND: ${CommandLine}")
```
davidd-amd, line 228: I trust this is necessary, but it is difficult to understand why. Can you provide some context for posterity?

Tensile/Source/lib/CMakeLists.txt
stellaraccident: This isn't about "bundling" the dep with the build so much as it is about supporting the versions of msgpack-cxx that are in the wild. The msgpack project, for better or worse, changed the way you depend on their library between v5 and v6, forcing all of their consumers to react to that change.

TheRock does in fact bundle msgpack (6.x) but that could easily just come from a distro which installs that version (I think this is where the issue was first discovered: on a distro that was shipping v6).

stellaraccident: afaict, nothing that we used of msgpack-cxx changed materially between v5 and v6 aside from how you take the dep -- and other consumers of it seem to concur (see link).

Tensile/cmake/TensileConfig.cmake
stellaraccident: Sure. CMake builds are split into configure time and build time, with everything running at configure time being part of the former. Build time (i.e. things that run in make/ninja/etc) needs to not depend on environment variables provided in the configure time CMake invocation.

As it is presently, it is only possible to build a Tensile based project where you configure and build in the same shell environment, with the same variables set. This is not how multi-project builds (and various other kinds of automation) are done and is considered an anti-pattern.

In most projects, the configure side should fully resolve all needed tools to absolute paths and use them explicitly in the build phase vs relying on implicit build-time path based resolution. However, Tensile uses path based resolution extensively. I had a patch which attempted to plumb through absolute paths for everything, but it was quite intricate and something that should be considered for the future as its own thing (and the needed surgery is even worse in the hipblaslt version). As a partial step, since the project physically is not correct in its current state, this patch does the next best thing and at least preserves the configure-time PATH for use at build time when setting up custom commands. Without this, the build time PATH could be anything and the project fails to build unless if the build invocation is in the same shell environment.

stellaraccident: I would recommend capturing this feedback in an issue for posterity and linking it in the source code. Ideally, in the future, we remove all PATH-based assumptions from Tensile and use explicit flags for tool locations, discovering them correctly at configure time.

Tensile/Source/lib/CMakeLists.txt
stellaraccident: (we can take out the part of the comment about "bundle in the build". That is specific to how TheRock operates)

davidd-amd (approved):

PR #2116, opened 2025-03-20
Loosen check for CLI compiler choices

### LINKED ISSUE #2112: [Issue]: accepted options for --cxx-compiler are too narrow
### Problem Description

Since https://github.com/ROCm/rocBLAS/commit/e88499263a68ed2d1699010d7d62d9fc164a870a rocblas will pass its `CMAKE_CXX_COMPILER` to tensile via --cxx-compiler unfortionatly tensile is way to strict with this, causing issues. building rocblas `-DCMAKE_CXX_COMPILER=/opt/rocm/bin/amdclang++` is perfectly valid, however tensile will not accept this bailing with:

```
TensileCreateLibrary: error: argument --cxx-compiler: invalid choice: '/opt/rocm/bin/amdclang++' (choose from amdclang++, hipcc)
```

This is due to https://github.com/ROCm/Tensile/blob/a164ed70df55ceb076f4a1f2fa2ea11f0f1384cc/Tensile/Tensile.py#L131 please remove the choices array here. If desired you can instead check that the path to the given compiler ends with hipcc or amdclang++ but really imo this should only be a warning or something the user can override anyways since there is no reason tensile objects could not be compiled with upstream llvm too. There might also be distro specific reasons for the compiler binary to be renamed.

### Operating System

arch

### CPU

Any Build issue

### GPU

Any Build issue

### ROCm Version

Tensile at 80dd256bbdfff46538210ce2f862743d2b22b211 rocblas at 2ba9a12d62d3de877c75cbd8dff73c50d54b6a98

bstefanuk:
Resolves #2112 

**Summary:**

Removes overly restrictive choices for `--cxx-compiler` when invoking Tensile.

**Outcomes:**

The user can pass fully qualified paths to Tensile. There is still a validation check that occurs during program initialization, but this validation is significantly less restrictive compared to setting choices directly in the argument parser.

### DIFF
--- a/Tensile/Tensile.py
+++ b/Tensile/Tensile.py
@@ -132,7 +132,7 @@ def addCommonArguments(argParser):
         type=str,
         help="C++ compiler used when generating binaries."
         "On Linux, amdclang++ (default) or hipcc. On Windows clang++ (default) or hipcc. "
-        "On Windows, file extensions will be searched according to the PATHEXT environment variable. "
+        "On Windows, include the file extension, or extensions will be searched according to the PATHEXT environment variable. "
         "Pass a fully-qualified path to override environment inspection when searching for the compiler."
     )
     argParser.add_argument(
--- a/Tensile/Tensile.py
+++ b/Tensile/Tensile.py
@@ -131,9 +131,9 @@ def addCommonArguments(argParser):
         default=ToolchainDefaults.CXX_COMPILER,
         type=str,
         help="C++ compiler used when generating binaries."
-        "On linux, amdclang++ (default) or hipcc. On Windows clang++.exe (default) or hipcc. "
-        "Pass a fully-qualified path to override environment inspection when searching for the compiler. "
-        "If clang++.exe is specified, it must include the extension `.exe`.",
+        "On Linux, amdclang++ (default) or hipcc. On Windows clang++ (default) or hipcc. "
+        "On Windows, file extensions will be searched according to the PATHEXT environment variable. "
+        "Pass a fully-qualified path to override environment inspection when searching for the compiler."
     )
     argParser.add_argument(
         "--c-compiler",
--- a/Tensile/Tensile.py
+++ b/Tensile/Tensile.py
@@ -128,11 +128,12 @@ def addCommonArguments(argParser):
     argParser.add_argument(
         "--cxx-compiler",
         dest="CxxCompiler",
-        choices=[ToolchainDefaults.CXX_COMPILER, "hipcc"],
         default=ToolchainDefaults.CXX_COMPILER,
         type=str,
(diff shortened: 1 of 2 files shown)


yoichiyoshida (approved):

davidd-amd (approved):

Tensile/Tensile.py
```diff
@@ -128,11 +128,12 @@ def splitExtraParameters(par):
...
         "--cxx-compiler",
         dest="CxxCompiler",
-        choices=[ToolchainDefaults.CXX_COMPILER, "hipcc"],
         default=ToolchainDefaults.CXX_COMPILER,
         type=str,
         help="C++ compiler used when generating binaries."
-        " On linux, amdclang++ (default) or hipcc. On Windows clang++ (default) or hipcc.",
+        "On linux, amdclang++ (default) or hipcc. On Windows clang++.exe (default) or hipcc. "
+        "Pass a fully-qualified path to override environment inspection when searching for the compiler. "
+        "If clang++.exe is specified, it must include the extension `.exe`.",
```
TorreZuk: Might be clearer to say "On windows include the file extension..."

bstefanuk: Ah, good catch, actually this restriction has been lifted. I will update the usage string to reflect. Thank you.

bstefanuk: @TorreZuk Please review recent commits for the requested change and resolve if you are satisfied.

Tensile/Tensile.py
```diff
@@ -128,11 +128,12 @@ def splitExtraParameters(par):
     argParser.add_argument(
         "--cxx-compiler",
         dest="CxxCompiler",
-        choices=[ToolchainDefaults.CXX_COMPILER, "hipcc"],
         default=ToolchainDefaults.CXX_COMPILER,
         type=str,
         help="C++ compiler used when generating binaries."
-        " On linux, amdclang++ (default) or hipcc. On Windows clang++ (default) or hipcc.",
+        "On Linux, amdclang++ (default) or hipcc. On Windows clang++ (default) or hipcc. "
+        "On Windows, file extensions will be searched according to the PATHEXT environment variable. "
```
TorreZuk: Isn't it true that you can pass a file extension on windows?
If so maybe "On Windows, include the file extension, or extensions will be searched..."

TorreZuk (approved):
Looks good, suggestion made

Tensile/Tensile.py
bstefanuk: You're right, this would technically be more correct. Updated.

TorreZuk:
@bstefanuk is this required as 6.5 cherry pick?   Not sure when the choices was added....

PR #2133, opened 2025-05-19
revert WIN32 breaking path change (#2131)

TorreZuk:
(cherry picked from commit 22e594b9b9e3e710b7ed7f78ca7494429de98aed)

* fixes rocBLAS windows build unblocking all downstream  (build fix verified in https://github.com/ROCm/rocBLAS/pull/1613)

### DIFF
--- a/Tensile/cmake/TensileConfig.cmake
+++ b/Tensile/cmake/TensileConfig.cmake
@@ -219,12 +219,17 @@ function(TensileCreateLibraryFiles
   if (WIN32 OR (VIRTUALENV_BIN_DIR AND VIRTUALENV_PYTHON_EXENAME))
     set(CommandLine ${VIRTUALENV_BIN_DIR}/${VIRTUALENV_PYTHON_EXENAME} ${CommandLine})
   endif()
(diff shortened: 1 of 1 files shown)


TorreZuk:
A small part of https://github.com/ROCm/Tensile/pull/2110 was reverted as breaks the windows build.  If this path caching is required fix the syntax and re-introduce to develop after confirming windows build succeeds.

stellaraccident (commented):
Thanks. There is a more robust change in hipBLASLt next-cmake which makes proper use of VERBATIM in custom commands vs quotes like this. I can't see the legacy windows build error (this does work on windows but for some reason does not work on that windows build).

I will also shortly be having us turn down the existing closed source windows build: it is checking many things that are not appropriate for the OSS projects, and in the very near future, we will not be gating on it. 

If you don't mind, I'll be back at my keyboard in a few hours and will look more closely at the logs.

AlexBrownAMD (approved):

TorreZuk:
@stellaraccident if you fix using VERBATIM in develop it can be cherry picked into this same target branch

PR #2135, opened 2025-05-26
Limit number of Tensile workers to 61 on Windows due to system limit.

lucbruni-amd:
**Summary:**

Windows supports at most 61 workers because the scheduler uses WaitForMultipleObjects directly, which has the limit (the limit is actually 64, but some handles are needed for accounting).

Ensure `cpu_count` is at most 61 to account for this.

### DIFF
--- a/Tensile/Parallel.py
+++ b/Tensile/Parallel.py
@@ -36,7 +36,10 @@ def CPUThreadCount(enable=True):
         return 1
     else:
         if os.name == "nt":
-            cpu_count = os.cpu_count()
+            # Windows supports at most 61 workers because the scheduler uses
+            # WaitForMultipleObjects directly, which has the limit (the limit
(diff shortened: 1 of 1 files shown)


bstefanuk (approved):

lucbruni-amd:
Hi @bstefanuk, are these CI checks expected to fail? If so, and if the changes LGTM, feel free to merge as I have no authorization. Thanks!

bstefanuk:
> Hi @bstefanuk, are these CI checks expected to fail? If so, and if the changes LGTM, feel free to merge as I have no authorization. Thanks!

They shouldn't be failing. I'm looking into these failures, and I'll merge if they can be resolved with a re-run.

ammallya:
Mirgrated to ROCm/rocm-libraries

PR #2138, opened 2025-06-03
Enable code coverage for Tensile Host C++ tests

ibrahimw1:
**Summary:**

Add initial support for generating code coverage reports for the Tensile C++ Host using existing tests. This enables developers to locally generate and inspect coverage reports, and lays the ground work for future CI integration.
- Adds `TENSILE_ENABLE_COVERAGE` CMake option to toggle code coverage instrumentation.
- Integrates a `make coverage` target to build, run tests, and generate HTML reports in:
```
build/
    coverage-report/
```
- Updates `tasks.py` to support building and running with coverage via `invoke`.
- Updates `tox.ini` for enabling coverage.


HostLibraryTests/CMakeLists.txt
```diff
@@ -157,6 +158,7 @@ if(TENSILE_USE_HIP)
     add_subdirectory(hip)
 endif()
 
+# Target file
```
davidd-amd: delete

HostLibraryTests/CMakeLists.txt
```diff
@@ -37,6 +37,7 @@ option(TENSILE_USE_MSGPACK   "Use msgpack for parsing config files." ON)
 option(TENSILE_USE_OPENMP    "Enable OpenMP multithreading." ON)
 option(TENSILE_STATIC_ONLY   "Disable exposing Tensile symbols in a shared library." ON)
 option(TENSILE_DISABLE_CTEST "Disable CTest integration." ON)
+option(TENSILE_ENABLE_COVERAGE "Enable code coverage for TensileHost library." OFF)
```
reboss, line 40: Would you mind naming this `BUILD_CODE_COVERAGE` to be consistent with other projects?

HostLibraryTests/CMakeLists.txt
```diff
@@ -196,3 +198,51 @@ endif()
...
 endif()
+
+if(TENSILE_ENABLE_COVERAGE)
+    message(STATUS "Code coverage for TensileHost is ENABLED.")
+
+    target_compile_options(TensileHost PRIVATE -g -O0 -fprofile-instr-generate -fcoverage-mapping)
+    target_link_options(TensileHost PRIVATE -fprofile-instr-generate)
+    message(STATUS "Applied coverage flags to TensileHost.")
+
+    target_compile_options(TensileTests PRIVATE -g -O0 -fprofile-instr-generate -fcoverage-mapping)
```
reboss: I don't think you want to instrument your test cases.  This will have misleading coverage results.

HostLibraryTests/CMakeLists.txt
```diff
@@ -196,3 +198,51 @@ endif()
 if(TENSILE_USE_OPENMP)
     target_link_libraries(TensileTests PRIVATE custom_openmp_cxx)
 endif()
+
+if(TENSILE_ENABLE_COVERAGE)
+    message(STATUS "Code coverage for TensileHost is ENABLED.")
+
+    target_compile_options(TensileHost PRIVATE -g -O0 -fprofile-instr-generate -fcoverage-mapping)
+    target_link_options(TensileHost PRIVATE -fprofile-instr-generate)
```
reboss: Generally link options are `PUBLIC` so that anything that links to the target will also get instrumented.

HostLibraryTests/CMakeLists.txt
```diff
@@ -196,3 +198,51 @@ endif()
...
+        COMMAND ${CMAKE_COMMAND} -E make_directory "${coverage_dir}/profraw"
+        COMMAND ${CMAKE_COMMAND} -E env "LLVM_PROFILE_FILE=${coverage_dir}/profraw/tensile-host-coverage_%p.profraw"
+                                        GTEST_LISTENER=NO_PASS_LINE_IN_LOG 
+                                        $<TARGET_FILE:TensileTests>
+      )
+
+    find_program(
+        LLVM_PROFDATA
+        llvm-profdata
+        HINTS /opt/rocm/llvm/bin
```
davidd-amd: use relative path to `llvm/bin`. Then this will resolve to `<cmake search path>/llvm/bin`. So if `<cmake search path>` is `/opt/rocm` you will have `/opt/rocm/llvm/bin`.

Experiment with [CMAKE_FIND_DEBUG_MODE](https://www.bing.com/search?pglt=163&q=cmake+find+debug+mode&cvid=617a80eb421e44bea8e9a27782025298&gs_lcrp=EgRlZGdlKgYIABBFGDkyBggAEEUYOTIGCAEQABhAMgYIAhAAGEAyBggDEAAYQDIGCAQQABhAMgYIBRAAGEAyBggGEAAYQDIGCAcQABhAMgYICBAAGEAyCAgJEOkHGPxV0gEINDQ2MGowajGoAgCwAgA&FORM=ANNAB1&PC=U531)=ON

HostLibraryTests/CMakeLists.txt
davidd-amd: @reboss I agree on being consistent. That said, we also need to ensure that variables like this are prefixed with the project name in a super build context. Otherwise, we wind up with a coarse grain than we would like. Maybe we could honor both a project specific variable and BUILD_CODE_COVERAGE?

HostLibraryTests/CMakeLists.txt
davidd-amd: TensileTests links against TensileHost. If I understand correctly, that means if we remove the coverage flags from the test target per your comment above and change this to PUBLIC we will wind up transitively adding the instrumentation flag to the test target.

Another reason we may not want to make this PUBLIC is again for super builds this would leak coverage flags to consumers even if we have coverage disabled on the consuming targets. Granted that scenario probably won't play mush or at all in practice.

tox.ini
```diff
@@ -46,6 +47,12 @@ deps =
 commands =
     invoke hostlibtest --clean --configure --build
 
+[testenv:hostlibtest-cov-run]
+deps =
+    invoke
+commands =
+    invoke hostlibtest --clean --configure --build --run --coverage
+
```
bstefanuk: I think we should prefer to use the existing `testev:hostlibtest` for this instead of adding another simply to pass the run and coverage flags. You can review the tox documentation here: https://tox.wiki/en/latest/index.html).wiki/en/latest/index.html, but calling something like `tox run -e hostlibtests -- --run --coverage` should work if you add `{posargs}` to the end of the command in `testenv:hostlibtest`.

Note this is a developer feature and doesn't affect the pipelines if we're running `make coverage`

tasks.py
```diff
@@ -17,9 +18,14 @@ def hostlibtest(c, clean=False, configure=False, build=False, run=False):
...
+            pty=True
         )
     if build:
         c.run(f"cmake --build `pwd`/{dir} -j4", pty=True)
     if run:
-        c.run("./{dir}/TensileTest")
+        if coverage:
+            c.run(f"cmake --build `pwd`/{dir} --target coverage --parallel", pty=True)
+        else:
+            c.run("./{dir}/TensileTests")
```
bstefanuk: Add newlines at EOF

HostLibraryTests/CMakeLists.txt
```diff
@@ -196,3 +197,51 @@ endif()
...
+        coverage
+        DEPENDS code_cov_tests
+        COMMAND ${LLVM_PROFDATA} merge -sparse "${coverage_dir}/profraw/tensile-host-coverage_*.profraw" -o "${coverage_dir}/tensile-host.profdata"
+        COMMAND ${LLVM_COV} report -object $<TARGET_FILE:TensileHost> -instr-profile="${coverage_dir}/tensile-host.profdata"
+        COMMAND ${LLVM_COV} show -object $<TARGET_FILE:TensileHost> -instr-profile="${coverage_dir}/tensile-host.profdata" -format=html -output-dir="${coverage_dir}"
+        
+        COMMENT "Generating TensileHost coverage report... HTML report in ${coverage_dir}/index.html"
+    )
+
+endif()
```
bstefanuk: Add newline at EOF

HostLibraryTests/CMakeLists.txt
reboss: Right, so `TensileTests` needs to be instrumented when linking to instrumented code, but not compiled with instrumentation.  Essentially, `TensileTests` needs to have the llvm runtime available to be able to record when an instrumented call is made but we don't want to instrument the functions that are defined in `TensileTests` themselves.

Without the link options on the tests, you'll just be adding a bunch of instrumentation to your code that can't be read by anything and ultimately won't generate a report.

HostLibraryTests/CMakeLists.txt
```diff
@@ -196,3 +197,51 @@ endif()
 if(TENSILE_USE_OPENMP)
     target_link_libraries(TensileTests PRIVATE custom_openmp_cxx)
 endif()
+
+if(TENSILE_ENABLE_COVERAGE)
+    message(STATUS "Code coverage for TensileHost is ENABLED.")
+
+    target_compile_options(TensileHost PRIVATE -g -O0 -fprofile-instr-generate -fcoverage-mapping)
```
bstefanuk: I believe this will create an unoptimized build, even for Release builds, if coverage is enabled.

reboss (approved):

.jenkins/common.groovy
```diff
@@ -26,14 +26,14 @@
...
-def runCompileCommand(platform, project, jobName, boolean debug=false)
+def runCompileCommand(platform, project, jobName, boolean debug=false, boolean codecov=false)
 {
     project.paths.construct_build_prefix()
 
     String compiler = '/opt/rocm/bin/amdclang++'
     // Do release build of HostLibraryTests on CI until it is upgraded to rocm 5.3 to
     // avoid bug causing long build times of certain files.
-    String buildType = 'Release' // debug ? 'Debug' : 'RelWithDebInfo'
+    String buildType = 'Debug' // debug ? 'Debug' : 'RelWithDebInfo'
```
davidd-amd: Seems like we should be conditionally setting this to debug
```suggestion
    String buildType = (debug || codecov) ? 'Debug' : 'Release'
```

davidd-amd (approved):

bstefanuk:
@ibrahimw1 Great work! Last thing: preferably, we have the code coverage posted somewhere from the CI we so can view it without having the run it locally. I don't currently see it in the Jenkins build artifacts. Is there another place it's posted?

davidd-amd (approved):

bstefanuk (approved):

PR #2144, opened 2025-06-06
Re-apply Windows compatibility fixes to command construction.

marbre:
The original patch
https://github.com/ROCm/TheRock/blob/6e1bd9b82a581487712f95ef86c0ba27c77faec9/patches/amd-mainline/Tensile/0002-Capture-configure-time-PATH-when-invoking-tensile-at.patch was only partly merged into upstream with commit 204b9a4. As a result it broke Windows and got partly reverted with commit 22e594b. This re-applies to patch to fix building Tensile with TheRock, both Linux and Windows.

### DIFF
--- a/Tensile/cmake/TensileConfig.cmake
+++ b/Tensile/cmake/TensileConfig.cmake
@@ -215,21 +215,24 @@ function(TensileCreateLibraryFiles
     set(Options ${Options} "--architecture=${archString}")
   endif()
 
+  # We do not need to do device enumeration at library build time.
+  set(Options ${Options} "--no-enumerate")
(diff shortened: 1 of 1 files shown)


bstefanuk:
@marbre does next-cmake eliminate the need for this commit?

marbre:
> @marbre does next-cmake eliminate the need for this commit?

I hope so, but so far we're not building rocBLAS or Tensile with next-cmake.

bstefanuk (approved):

ammallya:
Mirgrated to ROCm/rocm-libraries

PR #2145, opened 2025-06-07
Update treatment of msgpack in cmake support

davidd-amd:
**Summary:**

Tensile has a build time dependency on msgpack. In particular, Tensile depends on the msgpack header files to build. When consuming msgpack in the current build system we link a version specific target:

-` msgpackc` <= v5
- `msgpack-cxx` > v5

`msgpack-cxx` is an interface target and provides paths to msgpack header files. Whereas, `msgpackc` is shared library target. As a consequence, this introduces an unnecessary dependency on libmsgpack.so. Ideally, we don't link libraries that we do not require. More importantly, in some of our test environments, libmsgpack.so is propagated to librocblas as a runtime dependency and fails because libmsgpack.so cannot be found:

```
error while loading shared libraries: libmsgpackc.so.2: cannot open shared object file: No such file or directory
```

In this patch, we set the msgpack target depending on if:

1. v6 or greater is detected
2. v5 or older is detected

We then query the msgpack target for the `INTERFACE_INCLUDE_DIRECTORIES` and add these directories as a build requirement for the tensile host target. The goal of using the `INTERFACE_INCLUDE_DIRECTORIES` directly is to future proof against future changes in msgpack that might reintroduce a runtime dependency. In addition, it communicates to the reader that Tensile only depends on the msgpack header files at build time as msgpack does not provide a "msgpack-headers" target similar to fmt or spdlog.
 
**Notable changes:**

Tensile host target only depends on msgpack headers at build time.

**Testing and Environment:**

Mathlibs ubuntu 22 test image and CQE test image where runtime error was discovered.


TorreZuk (approved):
I think this now matches the design of the code pre May 6th ( before theROCK PR landed ) so can't imagine runtime dependency from this change. Even if it does find different targets we should only expect build errors and not runtime errors.   The only difference besides target preference order is:
        # include C++ headers manually
        # External header includes included as system files
        target_include_directories(TensileHost 
            SYSTEM PRIVATE $<BUILD_INTERFACE:${msgpack_inc}>
        )
        Which again I think is fine as only build errors should differ.  
        
        If all passes and you merge you can ping me for rocBLAS tag PR, or you can do it and I will watch for it.  You can also make the tensile "hotfix" rel 7.0 cherry pick PRs as we can also start those in rocblas after you merge to Tensile release-staging/rocm-rel-7.0.

PR #2146, opened 2025-06-07
Update treatment of msgpack in cmake support

davidd-amd:
**Summary:**

Cherry picks msgpack patch from #2145


TorreZuk (approved):
Fine by me same review as PR to develop

PR #2154, opened 2025-07-22
Add gfx1150 and complete gfx1151 support

amd-mtrifuno:
- Add gfx1150 support
- Add missing changes for gfx1151


davidd-amd (approved):
Approving - tests no longer run on this project but need to merge into 6.4 branch here. Staging testing was done on this patch according to PM. Also, we have similar patches in rocm-libraries develop and 7.0 release branches that were tested.

babakpst (approved):

